PR #11, opened 2016-10-11
Add ObjC properties to Python classes

Longhanks:
This PR will allow classes defined in Python to add Objective C properties using the following syntax:

```Python
class A(NSObject):
    prop = objc_property()
```

This is especially useful for implementing protocols that require properties.
Properties are created using getter and setter methods (that means, they are strong, nonatomic). They can be queried from within the ObjC-Runtime as well as in Python using the familiar dot syntax.

With this pull request, it's now possible to assign a main storyboard to the Xcode project and have the PythonAppDelegate behave just like the compiled AppDelegate.h/AppDelegate.m (like [this](https://gist.github.com/Longhanks/c3c36b97cc9f1dbcef9410177cee4341)

So Python is now a full drop-in replacement for Objective C!

### DIFF
--- a/rubicon/objc/__init__.py
+++ b/rubicon/objc/__init__.py
@@ -4,7 +4,7 @@ from .objc import (
     objc, send_message, send_super,
     get_selector,
     ObjCClass, ObjCInstance, NSObject,
-    objc_ivar, objc_rawmethod, objc_method, objc_classmethod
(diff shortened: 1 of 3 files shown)


Longhanks:
Xcode 6.2 beta has failed... (?)

freakboy3742:
I'm going to guess the XCode failure is due to an eccentricity of older versions of Objective C. I'm going to accept this on the basis that it works on all the more recent XCode versions.

Issue #14, opened 2016-11-24
Type handling - No restype encoding for...


exvito:
Hello BeeWare,

Thanks for sharing rubicon-objc. I very much like the light approach at playing with Mac OS APIs with it, instead of PyObjc -- while perfectly confortable with Python, C and generic low-level systems/networking code, I confess myself as a complete Objective-C + Mac OS newbie.

(a bit of context: I've been exploring bluetooth programming on various platforms. I started with Linux, which was surprisingly nice and while waiting for some PRs to be reviewed and hopefully merged, I decided to move on to the Mac; this issue arises from my first very simple attempt).

## The facts

Exploring the Apple docs on bluetooth, in particular https://developer.apple.com/reference/iobluetooth/iobluetoothdevice?language=objc, I came up with the following script to enumerate the names and bluetooth addresses of paired devices:

```
from ctypes import cdll, util
from rubicon.objc import ObjCClass

if __name__ == '__main__':

    cdll.LoadLibrary(util.find_library('IOBluetooth'))
    IOBluetoothDevice = ObjCClass('IOBluetoothDevice')

    paired_devices = IOBluetoothDevice.pairedDevices()
    num_devices = paired_devices.count()
    for i in range(num_devices):
        device = paired_devices.objectAtIndex_(i)
        name = device.name
        address = device.getAddress()
        print('name=%r address=%r' % (name, address))
```

When run rubicon-objc complains, printing out:
```
No restype encoding for b'getAddress' (b'r^{BluetoothDeviceAddress=[6C]}')
```

With this, the names of the devices are printed out nicely, but the addresses show up as `None`.

Browsing the code, I quickly figured out the culprit in the `ObjCMethod` class. I went digging and learned a bit about Objective C type encodings (at https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html) and quickly matched the `b'r^{BluetoothDeviceAddress=[6C]}'` to the following declaration in `/System/Library/Frameworks/IOBluetooth.framework/Versions/A/Headers/Bluetooth.h`:

```
typedef struct  BluetoothDeviceAddress          BluetoothDeviceAddress;
struct  BluetoothDeviceAddress
{
        uint8_t         data[ 6 ];
};
```

(this is under 10.9.5, but per docs should be present in all the more recent OS releases)

## The way forward

[...]

PR #16, opened 2016-11-30
Add support for ObjCMethod to return pointers to C structs

exvito:
This PR addresses the discussion in #14 and is in progress.
I'll update with subsequent commits.

### DIFF
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -805,6 +805,11 @@ class ObjCBoundMethod(object):
         """Initialize with a method and ObjCInstance or ObjCClass object."""
         self.method = method
         self.objc_id = objc_id
+        # set restype if previously set at class level via _RestypeHelper
+        if isinstance(self.objc_id, ObjCInstance):
+            restype = self.objc_id.objc_class.instance_restype.get(method.name)
+            if restype:
+                method.restype = restype
 
     def __repr__(self):
         return '<ObjCBoundMethod %s (%s)>' % (self.method.name, self.objc_id)
@@ -1147,6 +1152,23 @@ def objc_rawmethod(f):
 ######################################################################
 
 
+class _RestypeHelper(object):
+    """
+    Supports class level setting of instance method restypes.
+    """
+    def __init__(self, objc_class, method_name):
+        self.__dict__['objc_class'] = objc_class
+        self.__dict__['method_name'] = method_name.encode('utf8')
+
+    def __setattr__(self, name, value):
+        # Only support setting 'restype'. Raise the exception that
+        # would have been raised when we were instantiated, otherwise.
+        if name != 'restype':
+            raise AttributeError('ObjCClass %s has no attribute %s' % (
+                self.objc_class, self.objc_method))
+        self.objc_class.instance_restype[self.method_name] = value
+
+
 class ObjCClass(type):
     """Python wrapper for an Objective-C class."""
 
@@ -1235,6 +1257,7 @@ class ObjCClass(type):
                     'metaclass': metaclass,
                     'name': objc_class_name,
                     'instance_methods': {},     # mapping of name -> instance method
+                    'instance_restype': {},    # mapping of name -> ctypes.restype
                     'class_methods': {},        # mapping of name -> class method
                     'instance_properties': {},  # mapping of name -> (accessor method, mutator method)
                     'class_properties': {},     # mapping of name -> (accessor method, mutator method)
@@ -1278,8 +1301,9 @@ class ObjCClass(type):
             try:
                 return self.__dict__[name]
             except KeyError:
-                # Otherwise, raise an exception.
-                raise AttributeError('ObjCClass %s has no attribute %s' % (self.name, name))
+                # Defer to _RestypeHelper in order to support
+                # class level setting of instance method restypes.
+                return _RestypeHelper(self, name)
 
     def __setattr__(self, name, value):
         # Convert enums to their underlying values.
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -739,6 +739,9 @@ class ObjCMethod(object):
         elif encoding[0:2] == b'r^' and encoding[2:] in self.typecodes:
             # const pointer, also don't care
             return POINTER(self.typecodes[encoding[2:]])
+        elif encoding[0:2] == b'^{' or encoding[0:3] == b'r^{':
+            # (const) pointer to a C structure, type handling up to caller
+            return c_void_p
         else:
             raise Exception('unknown encoding for %s: %s' % (self.name, encoding))
 


exvito:
@freakboy3742,

Here is my first take at both sides of the issue:
* Returning ```ctypes.c_void_p``` for methods returning pointers to C structs.
* Adding support for setting instance method restypes at the class level, so that it can be used in a "set it and forget it" approach.

Notes:
* I understand I may have not followed your guidance to the letter. The approach I took is maybe a bit clumsy and feels a bit spread, not sure. But I couldn't figure a way of avoiding the ```_RestypeHelper``` class (name may also need improvement!) Your feedback will be appreciated. :)
* From a caller's standpoint, this is working pretty much as you suggested:

```
from ctypes import cdll, util, Structure, c_ubyte, POINTER

class BluetoothDeviceAddress(Structure):
    _fields_ = [
        ('data', c_ubyte * 6),
    ]

if __name__ == '__main__':

    cdll.LoadLibrary(util.find_library('IOBluetooth'))
    IOBluetoothDevice = ObjCClass('IOBluetoothDevice')
    IOBluetoothDevice.getAddress.restype = POINTER(BluetoothDeviceAddress)

    paired_devices = IOBluetoothDevice.pairedDevices()
    num_devices = paired_devices.count()
    for i in range(num_devices):
        device = paired_devices.objectAtIndex_(i)
        name = device.name
        btda_ptr = device.getAddress()
        addr_str = ':'.join(map(lambda b: '%x' % b, btda_ptr[0].data))
        print('name=%r addr_str=%r' % (name, addr_str))
```

Final note:
* I can't seem to run the tests successfully, regardless of my changes.
* After ``make``` and setting ```DYLD_LIBRARY_PATH```, I run ```python setup.py test``` and get:

```
...
test_class_properties (tests.test_rubicon.RubiconTest)
A Python class can have ObjC properties with synthezied getters and setters. ... Traceback (most recent call last):
  File "_ctypes/callbacks.c", line 259, in 'converting callback result'
TypeError: cannot be converted to pointer
Exception ignored in: <function objc_method.<locals>._objc_method at 0x102deb400>
Segmentation fault: 11
```

...any hints? (running Mac OS 10.9.5 with Python 3.5.2 built from sources)

Thanks

freakboy3742:
Assuming the tests run fine on a clean build (if they don't there's a much bigger problem at play) - I'd start by narrowing down exactly which case is throwing the exception. I'm guessing something about getters and setters is messing with the return value handling; and when things go a little bit wrong with ctypes, you're only a stone's throw away from a segfault.

exvito:
Regarding tests:
* Tests were not running fine on a clean build - the segfault I shared before was obtained on a clean build.
* Replaced my hand built Python 3.5.2 installation with one from python.org.
* Tests now run:
  * On the master branch I'm getting one error in ```test_class_properties```.
  * On this proposed branch other failures need to be addressed.

Before proceeding, I'd like to understand the failure under the master branch. Do you have any hint why ```test_class_properties``` is failing on my system? Stack trace below:
```
======================================================================
ERROR: test_class_properties (tests.test_rubicon.RubiconTest)
A Python class can have ObjC properties with synthezied getters and setters.
----------------------------------------------------------------------
  File "/Users/tiago.montes/Work/Python/rubicon/exvito/rubicon-objc/tests/test_rubicon.py", line 410, in test_class_properties
    self.assertEqual(box.getSchemeIfPresent(), 'https')
  File "/Users/tiago.montes/Work/Python/rubicon/exvito/rubicon-objc/rubicon/objc/objc.py", line 811, in __call__
    return self.method(self.objc_id, *args)
  File "/Users/tiago.montes/Work/Python/rubicon/exvito/rubicon-objc/rubicon/objc/objc.py", line 783, in __call__
    result = to_value(ObjCInstance(result))
  File "/Users/tiago.montes/Work/Python/rubicon/exvito/rubicon-objc/rubicon/objc/objc.py", line 1325, in __new__
    objc_instance.__dict__['objc_class'] = ObjCClass(class_ptr)
  File "/Users/tiago.montes/Work/Python/rubicon/exvito/rubicon-objc/rubicon/objc/objc.py", line 1186, in __new__
    raise RuntimeError("Couldn't create ObjC class for pointer '%s'." % class_name_or_ptr)
RuntimeError: Couldn't create ObjC class for pointer 'c_void_p(None)'.

----------------------------------------------------------------------
```

I see that one test is skipped if OS < 10.10 with a note about "Property handling" in ```test_decimal_method```, which is my case, running 10.9.5. Without much knowledge on Objective C, could ```test_class_properties ``` be failing for a similar reason, being related to properties?

Thanks for your guidance.

freakboy3742:
It's entirely possible that class properties are affected as well. There was a fairly major API change introduced in OSX 10.10 about how properties are accessed. 

As a side note - It would also be possible to support old and new ObjC APIs at the same time - it just wasn't a personal priority.

exvito:
@freakboy3742,

Thanks for the contextualization, that certainly explains the failure I'm getting when testing the master branch under 10.9. I tried to undestand the underlying cause by reading both the code and the test and also looking for documentation on the "fairly major API change introduced in OSX 10.10 about how properties are accessed" you mentioned. Results:

* Under 10.9, in the ```test_class_properties``` test, the ```url.scheme``` property results in an instance like ```<ObjCBoundMethod b'scheme' (<ObjCInstance 0x101ac6c18: NSURL at 4348596624>)>```. If called, the underlying value is obtained.
* After lots of debugging and tracing I couldn't figure out the culprit exactly.
* This did not help: I tried hard finding for any references or documentation regarding the "API changes introduced in 10.10 about how properties are accessed" but couln't find any... I've probably looked in the wrong places and/or used inadequate keywords, being unfamiliar with the technology specifics. Do you have any such reference/pointer?

Refocusing on this PR:
* Regardless of the above mentioned test failing on my system, I went ahead and examined the  other test failures resulting from the change - they seem to be inline with a feeling that's been growing on me that the "set it and forget it" approach at defining a given method's return types may be far from ideal. I feel that any implementation that supports an approach like ```<class>.<instance-method>.restype = <ctype>``` may need to be reconsidered.
* Currently the following tests are failing:
  * ```test_non_existent_static_method```
  * ```test_non_existent_static_field```
  * ```test_non_static_access_static```
* Given that any implementation following this approach will have (correct me if I'm wrong) to create an intermediary object to handle ```.restype``` assignment (```_RestypeHelper``` in the current implementation):
  * It can be made to pass the first failing test by having the intermediary object raise ```AttributeError``` when called.
  * It may be impossible to raise ```AttributeError```s on the other two cases. Unless some kind of instance method introspection can be put in place before even creating such an object (again, this is where I'm a bit lost given my unfamiliarity with Objective C and libobjc).

Wrapping it up:
* I feel the change in 24de249458ec8db1bb8ba014d5bb9ec8c556e5df that adds support for Objective C methods returning C structs is simple, clear and to the point.
* This capability can have some tests added by extending both the current test harness and the unit tests.
* The "set and forget" approach for setting instance method return type is something I'm not sure of; what do you feel about an alternative, less "magical", approach where some specific function/method would handle that, avoiding interference with attribute access? What about separating that into a different PR / issue so that the "basic capability" gets in first?

PS: This all started because I needed to call ```getAddress()``` on an instance obtained from a class method -- and process its results which are pointers to C structs. Ironically, I later found out that the same instances have an ```addressString``` property that hold the information I need, ready to be used. This discovery makes this change less important to me (though I guess there are many lower-level classes that will have methods returning C structs). Nevertheless, I'll be happy to move it forward with you guidance, if you are willing to do so.

dgelessus:
Superseded by #43.

Issue #15, opened 2016-11-30
Pythonic interface for NSArray / NSMutableArray


exvito:
In the context of #14, a light discussion on a Pythonic interface for ```NSArray``` came up. The idea is to provide a sequence-like interface (collections.abc.Sequence?) that would simplify Python side ```NSArray``` manipulation such that, for example, instead of iterating like:

```
things = obcj.methodReturningNSArray()
for i in range(things.count()):
    thing = things.objectAtIndex_(i)
    do_something_with(thing)
```

...one could:

```
things = objc.methodReturningNSArray()
for thing in things:
    do_something_with(thing)
```

Other desirable/possible interfaces:
* Indexing as in ```things[index]```
* Length as in ```len(things)```
* Write-ability ```things[index] = something```
* Maybe a factory-like class method that would build an NSArray from a Python sequence.

Final notes:
* Implementation strategy could be either via wrapper or by customizing the Python NSArray/NSMutableArray class that rubicon-creates.
* Something similar would be desirable for NSDictionary / NSMutableDictionary.

Thanks.

freakboy3742:
Very much +1 on this. It may not even be that difficult, either; I haven't tried it, but I *think*:
```
NSArray = ObjCClass('NSArray')
NSArray.__getitem__ = nsarray_getitem(self, index):
```
for a suitably defined `nsarray_getitem`, with similar definitions for `__len__` etc, and an overload of `__getattr__` to allow direct lookup of these dunder methods, should be all that is needed. 

Of course, it might be more complicated than that :-)

The factory method for passing in lists as `NSArray` shouldn't be needed, either; it should be possible to pass in a Python list and transparently convert it into an NSArray. However, even then, a conversion function will be required; it just won't be needed as exposed API.

dimaqq:
A minor complication from fixed/immutable arrays and dictionaries, as their type is like `NSArray0` and `NSDictionary0`. I'm not sure how hight the trailing numbers may go...

dgelessus:
Those names belong to the various implementation classes of `NSArray` and such. They aren't meant to be used directly (their names start with two underscores). The proper way to check if something is a `NSArray` is to use the `isKindOfClass` method.

PR #17, opened 2016-12-04
iterable interface for NSArray (and subclasses).

Longhanks:
As requested in #15 

Implements a pythonic interface for NSArray and its subclasses.

Examples from the unit test:

```python
NSMutableArray = ObjCClass('NSMutableArray')
ns_array = NSMutableArray.array()
test_list = ['item1', 'item2', 'item3', 'item4', 'item5']
for item in test_list:
    ns_array.addObject_(item)

# Todo: implement len
self.assertEqual(ns_array.count, 5)

# NSArray itself can be compared to other NSArrays
# NSArray can be copied via copy-slice [:]
self.assertEqual(ns_array[:], ns_array)

# Needed because an NSArray does NOT equal a Python list. Their elements can, however, be equal.
def compare(some_NSArray, python_list):
    self.assertEqual([item for item in some_NSArray], python_list)

# Object retrieval via index
self.assertEqual(ns_array[3], test_list[3])

# positive indices
compare(ns_array[1:], test_list[1:])
compare(ns_array[2:3], test_list[2:3])
compare(ns_array[4:2], test_list[4:2])
compare(ns_array[:4], test_list[:4])

# negative indices
compare(ns_array[-1:], test_list[-1:])
compare(ns_array[-2:-3], test_list[-2:-3])
compare(ns_array[-4:-2], test_list[-4:-2])
compare(ns_array[:-4], test_list[:-4])

# positive indices with step 2
compare(ns_array[1::2], test_list[1::2])
compare(ns_array[2:3:2], test_list[2:3:2])
compare(ns_array[4:2:2], test_list[4:2:2])
compare(ns_array[:4:2], test_list[:4:2])

# negative indices with step 2
compare(ns_array[-1::2], test_list[-1::2])
compare(ns_array[-2:-3:2], test_list[-2:-3:2])
compare(ns_array[-4:-2:2], test_list[-4:-2:2])
compare(ns_array[:-4:2], test_list[:-4:2])

# positive indices with step -3
compare(ns_array[1::-3], test_list[1::-3])
compare(ns_array[2:3:-3], test_list[2:3:-3])
compare(ns_array[4:2:-3], test_list[4:2:-3])
compare(ns_array[:4:-3], test_list[:4:-3])

# negative indices with step -3
compare(ns_array[-1::-3], test_list[-1::-3])
compare(ns_array[-2:-3:-3], test_list[-2:-3:-3])
compare(ns_array[-4:-2:-3], test_list[-4:-2:-3])
compare(ns_array[:-4:-3], test_list[:-4:-3])

# test string representation
self.assertEqual(repr(ns_array), 'NSArray ' + repr(test_list))
```

The one thing I can't get to work is `__len__`, implementing it confuses `core_foundation.py to_value` and the Objective C classes can't be fetched from the runtime anymore... any hint would be great!

### DIFF
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -20,6 +20,7 @@ elif sizeof(c_void_p) == 8:
 ######################################################################
 
 objc = cdll.LoadLibrary(util.find_library('objc'))
+Foundation = cdll.LoadLibrary(util.find_library('Foundation'))
 
 ######################################################################
 
@@ -1287,7 +1288,7 @@ class ObjCClass(type):
         if method:
             ObjCBoundMethod(method, self.__dict__['ptr'])(value)
         else:
-            self.__dict__[name] = value
+            super(ObjCClass, self).__setattr__(name, value)
 
 
 ######################################################################
@@ -1324,6 +1325,10 @@ class ObjCInstance(object):
         class_ptr = c_void_p(objc.object_getClass(object_ptr))
         objc_instance.__dict__['objc_class'] = ObjCClass(class_ptr)
 
+        # If it's an array, we need the internal counter in the instance to implement iterable.
+        if objc.class_getName(class_ptr).decode('utf-8').startswith('__NSArray'):
+            objc_instance.__dict__['_counter'] = 0
+
         # Store new object in the dictionary of cached objects, keyed
         # by the (integer) memory address pointed to by the object_ptr.
         cls._cached_objects[object_ptr.value] = objc_instance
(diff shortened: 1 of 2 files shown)


rubicon/objc/objc.py
```diff
@@ -20,6 +20,7 @@
 ######################################################################
 
 objc = cdll.LoadLibrary(util.find_library('objc'))
+Foundation = cdll.LoadLibrary(util.find_library('Foundation'))
```
freakboy3742: Leaving out Foundation was a strategic choice; I was trying to avoid the core of Rubicon spiralling into a a huge project. If it's necessary, then we will include it, but I'd like to explore if we can do this with CFArray (from CoreFoundation)

rubicon/objc/objc.py
```diff
@@ -1287,7 +1288,7 @@ def __setattr__(self, name, value):
         if method:
             ObjCBoundMethod(method, self.__dict__['ptr'])(value)
         else:
-            self.__dict__[name] = value
+            super(ObjCClass, self).__setattr__(name, value)
```
freakboy3742: This seems like a red flag to me; what's the purpose behind this change?

rubicon/objc/objc.py
```diff
@@ -1324,6 +1325,10 @@ def __new__(cls, object_ptr):
         class_ptr = c_void_p(objc.object_getClass(object_ptr))
         objc_instance.__dict__['objc_class'] = ObjCClass(class_ptr)
 
+        # If it's an array, we need the internal counter in the instance to implement iterable.
```
freakboy3742: No you don't - because that makes the iterable object stateful. A ListIterator is a separate object from a list, so you can perform 2 parallel iterations over the same list.

tests/test_rubicon.py
```diff
@@ -428,3 +428,66 @@ def test_function_NSEdgeInsetsMake(self):
         self.assertEqual(insets.bottom, other_insets.bottom)
         self.assertEqual(insets.right, other_insets.right)
 
+    def test_NSArray_iterable(self):
+        "Pythonic interface for NSArray (and its subclasses) as iterable."
```
freakboy3742: This looks like it's just a test for `NSMutableArray`, not `NSArray`.

tests/test_rubicon.py
```diff
@@ -428,3 +428,66 @@ def test_function_NSEdgeInsetsMake(self):
...
+
+        NSMutableArray = ObjCClass('NSMutableArray')
+
+        ns_array = NSMutableArray.array()
+        test_list = ['item1', 'item2', 'item3', 'item4', 'item5']
+        for item in test_list:
+            ns_array.addObject_(item)
+
+        # Todo: implement len
```
freakboy3742: Intrigued to know what problem you encountered here.

tests/test_rubicon.py
```diff
@@ -428,3 +428,66 @@ def test_function_NSEdgeInsetsMake(self):
...
+        compare(ns_array[:4:-3], test_list[:4:-3])
+
+        # negative indices with step -3
+        compare(ns_array[-1::-3], test_list[-1::-3])
+        compare(ns_array[-2:-3:-3], test_list[-2:-3:-3])
+        compare(ns_array[-4:-2:-3], test_list[-4:-2:-3])
+        compare(ns_array[:-4:-3], test_list[:-4:-3])
+
+        # test string representation
+        self.assertEqual(repr(ns_array), 'NSArray ' + repr(test_list))
```
freakboy3742: This is a fine test of the capabilities of the Python interface when the object has been created Python-side; but it doesn't check if an object that has been received through an API will acquire those methods. To test this, you'll need to create a method on the test harness that returns `NSArray`/`NSMutableArray`.

rubicon/objc/objc.py
```diff
@@ -1324,6 +1325,10 @@ def __new__(cls, object_ptr):
...
+        if objc.class_getName(class_ptr).decode('utf-8').startswith('__NSArray'):
```
dimaqq: What about `NSMutableArray`? Does it not belong to same category? And how about `NSDictionary`

rubicon/objc/objc.py
```diff
@@ -1342,14 +1347,69 @@ def __new__(cls, object_ptr):
...
+                    key = self.count - (-1 * key)
+                if key > (self.count - 1) or key < 0:
+                    raise IndexError('list index out of range')
+                else:
+                    return self.objectAtIndex_(key)
+            elif isinstance(key, slice):
+                array = ObjCClass('NSMutableArray').array()
+                helper = [item for item in self]
+                for item in helper[key]:
+                    array.addObject_(item)
```
dimaqq: This design decision needs discussion:

If `isinstance(a, NSArray)`, would you rather `a[:]` return a Python list or clone `NSArray`?

freakboy3742:
This has been superseded by #42.

PR #18, opened 2016-12-10
Fixes for send_message on non-x86_64 architectures

dgelessus:
My original goal was to fix structure returns on 32-bit iOS (ARM32), but then I went down a bit of a rabbit hole. I apologize for the large PR - there were some tests/fixes for one arch that uncovered more issues on another arch, and I didn't want to PR half-broken code.

Short summary of the changes:

* Architecture detection for ARM32 and ARM64 (`__arm__` and `__arm64__` bools to match `__i386__` and `__x86_64__`)
* Fixes for `objc_msgSend`/`objc_msgSend_stret` selection on i386, ARM32 and ARM64
* Unit tests for struct-returning methods
* Unit tests for `send_message` (which is not used often, and had some bugs on non-x86_64 architectures because of that)
* Universal build (i386 and x86_64) for the test harness library
* Test environments to test the library in i386 mode on 64-bit OS X
* Some minor fixes

For more details on how and why all of this happened, see the commit log, I explained the reasons for most of the fixes there.

The unit tests all run fine for me on Python 3.4 and 3.5 on i386 and x86_64, and I did some manual testing on ARM32 and ARM64 on iOS as well. But if I broke something else by accident, please tell me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,10 @@ local
 *.o
 *.d
 *.dylib
+
+# OS X
+.DS_Store
+
+# PyCharm
+.idea/
+*.iml
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,17 @@
-
 OBJ_FILES=tests/objc/Thing.o tests/objc/Example.o tests/objc/BaseExample.o
+
+# By default, build a universal i386/x86_64 binary.
+# Modify here (or on the command line) to build for other architecture(s).
+EXTRA_FLAGS=-arch i386 -arch x86_64
 
 all: tests/objc/librubiconharness.dylib
 
 tests/objc/librubiconharness.dylib: $(OBJ_FILES)
-	clang -dynamiclib $(OBJ_FILES) -fobjc-arc -fobjc-link-runtime -o tests/objc/librubiconharness.dylib
+	clang -dynamiclib -fobjc-link-runtime $(EXTRA_FLAGS) $(OBJ_FILES) -o tests/objc/librubiconharness.dylib
 
 clean:
 	rm -rf tests/objc/*.o tests/objc/*.d tests/objc/librubiconharness.dylib
 
-%.o : %.m
-	clang -x objective-c -fobjc-arc -I./tests/objc -c $< -o $@
+%.o: %.m
+	clang -x objective-c -fobjc-arc -I./tests/objc -c $(EXTRA_FLAGS) $< -o $@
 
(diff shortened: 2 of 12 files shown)


freakboy3742:
Wow - this is amazing! Thanks for the contribution!

The patch looks fantastic; the only issue is the CI failure. Looks like a small modification to the build command is needed for XCode 6.4... or make your case why we should drop XCode 6.4 from the build matrix. You won't get much pushback from me *personally* about that; but we've had a couple of questions recently about support on older versions of macOS, so I'm a little hesitant to drop 6.4.

dgelessus:
OK, it seems that Xcode 6 doesn't support ARC on the old Objective-C runtime, which is used by the i386 build. The code for the test harness isn't very long, so the easiest solution might be to not use ARC, and add manual `retain`/`release` calls in the few places where it's needed. I'll see if I can do that, perhaps Xcode 6 is happy then.

It's kind of annoying that the i386 build requires the old ABI/runtime - it's also why I had to manually add ivars and `@synthesize` for all properties. I tried forcing the new ABI with `-fobjc-nonfragile-abi`, but then the linker complained about missing symbols from Foundation. I'm guessing the i386 build of Foundation doesn't support the new ABI.

Regarding the Travis build - it looks like Travis runs `python3 setup.py test` directly rather than using `tox`. This means that all of the non-default configurations (Python 3.4, PyPy, and everything on i386) aren't tested. It might be good to make Travis use `tox` too.

dgelessus:
OK, unless I'm missing something, dropping ARC for the test harness really isn't a big deal. The synthesized properties do the retaining and releasing automatically, and most of the other code passes objects through from/to properties and doesn't need any manual retaining/releasing either.

dgelessus:
Ah, the `python3` installed by `brew` is only for x86_64. Any idea if it's possible to make Homebrew install a universal binary for Python? With MacPorts you can do that by adding `+universal` to the port name.

dgelessus:
Nevermind, tox doesn't want to run properly on Travis ([example](https://travis-ci.org/pybee/rubicon-objc/jobs/183213758#L443)), it always causes a strange exception in setuptools. I've reverted back to normal `python3 setup.py test` for now. (I'll squash all my Travis testing commits away, those don't really need to be merged into the main repo.)

dgelessus:
@freakboy3742 Sorry to be a pain, but is there anything else that needs to be done before this can be merged?

The thing where I tried to run tox in the Travis builds isn't necessary for the rest of the PR - if I figure out a fix for the setuptools issue, I can make a separate PR for that later. I removed my (broken) changes to the Travis config, so the Travis builds aren't affected by the PR anymore.

freakboy3742:
Apologies for the delay in the review for this patch. The last month or so has been pretty hectic for me.

dgelessus:
No worries @freakboy3742. Hope you enjoyed the holidays. `:)`

PR #22, opened 2017-01-31
Better str and repr for ObjCInstance

dgelessus:
Adds improved `__str__` and `__repr__` methods to `ObjCInstance`. For `CFString`/`NSString` `ObjCInstance`, `__str__` converts it to a Python `str`, for other `ObjCInstance`s it returns the `debugDescription`. (For `NSString` the `debugDescription` is the string itself, but `CFString` has a different `debugDescription`, which is why the special case is necessary.) The `__repr__` now also includes the `__str__`.

Example:

```python
>>> rubicon.objc.ObjCClass("NSArray").arrayWithObjects_("foo", "bar", "bee")
<ObjCInstance 0x104807898: __NSArrayI at 0x1010b34b0: <__NSArrayI 0x1010b34b0>(
foo,
bar,
bee
)
>
>>> rubicon.objc.core_foundation.at("abcdef")
<ObjCInstance 0x1045b2518: __NSCFString at 0x1010c0cd0: abcdef>
>>> str(_)
'abcdef'
```

### DIFF
--- a/rubicon/objc/__init__.py
+++ b/rubicon/objc/__init__.py
@@ -10,7 +10,6 @@ from .objc import (
 from .core_foundation import at, to_str, to_number, to_value, to_set, to_list
 
 from .types import (
-    text,
     NSInteger, NSUInteger,
     CGFloat,
     NSPointEncoding, NSSizeEncoding, NSRectEncoding, NSRangeEncoding, UIEdgeInsetsEncoding, NSEdgeInsetsEncoding,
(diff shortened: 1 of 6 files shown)


dgelessus:
Oh, it seems I based this on my branch for #20 instead of master, so there are some extra commits in here that belong to the other PR. I can probably fix that if necessary.

freakboy3742:
No worries - the changes seemed pretty straight forward. Thanks for another awesome PR!

PR #23, opened 2017-02-03
Allow disabling argument and result conversion when calling ObjCMethods

dgelessus:
Suppressing result conversion is useful when you want to call methods on an objects that would be converted to Python objects (such as `NSString`), or when a method returns an `id` that isn't actually valid. Suppressing argument conversion is not as useful, but there are probably use cases for that too, so I added that as well.

I also made a small change to argument conversion, so it only converts arguments to (Core) Foundation types if the respective argtype is `objc_id`. Otherwise a call like `NSString.stringWithUTF8String_(b"foobar")` fails, because the bytestring is converted to `NSString` by `to_value`.

(Also added a test for my previous PR.)

### DIFF
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -818,14 +818,28 @@ class ObjCMethod(object):
             self.func.argtypes = self.argtypes
         return self.func
 
-    def __call__(self, receiver, *args):
+    def __call__(self, receiver, *args, convert_args=True, convert_result=True):
         """Call the method with the given id and arguments.  You do not need
         to pass in the selector as an argument since it will be automatically
         provided."""
         f = self.get_callable()
-        # Automatically convert Python strings into ObjC strings
-        from .core_foundation import from_value, to_value
-        converted_args = [from_value(arg) for arg in args]
+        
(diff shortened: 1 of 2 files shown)


freakboy3742:
Repeating an earlier offer - not sure if you saw it or not: You've been doing some great work on Rubicon, so if you'd like to join the committer team, you'd be most welcome.

dgelessus:
Oh, it looks like I forgot to reply to that part last time, sorry. I appreciate the offer, I'm just always a bit afraid that I'll break something by accident. That's why I wanted to stick with PRs for a short while, so you'd have a look at the changes and tell me if I did something wrong `:)` Though I am a bit more familiar with the code by now, so if you think it's okay, I would be fine with committing directly to the repo.

freakboy3742:
Awesome! It still won't be "commit direct to repository"; since there will be two of us with commit access, I can turn on the "review required before commit" switch on the repository. So you'll still be getting a set of eyes on your changes - I can just call on the same in response. 

If you can shoot me an email at russell@keith-magee.com, we can sort out a couple of other details.

PR #24, opened 2017-02-08
Make ObjCClass inherit from ObjCInstance

dgelessus:
(I had to swap the order of `ObjCClass` and `ObjCInstance` for this, which makes the default diff somewhat useless. [This diff](https://github.com/dgelessus/rubicon-objc/compare/53b0c4bbc9b110208043a8f714fd115356b55d8b...objcclass_extends_objcinstance) is a bit more useful.)

Objective-C classes are also objects, and class methods are actually instance methods of the corresponding metaclass ([helpful explanation](http://www.sealiesoftware.com/blog/archive/2009/04/14/objc_explain_Classes_and_metaclasses.html), [helpful diagram](http://www.sealiesoftware.com/blog/class%20diagram.pdf)). By making `ObjCClass` inherit from `ObjCInstance`, it is possible to reuse `ObjCInstance`'s code for method and property access in `ObjCClass`. As a side effect, it also means that any methods added to `ObjCInstance` in the future will also work on `ObjCClass`.

To implement this I had to add a small hack to `ObjCInstance.__new__`, which passes on the `name, bases, attrs` arguments to `type` when it gets them from `ObjCClass` via `super`. It's not a major thing though, and it's only needed in a single place - subclasses of `ObjCInstance` are not affected.

Other changes:

* Made `ObjCInstance.objc_class` a property instead of a "real" attribute. I don't remember why exactly this was necessary (I had this branch lying around half-complete for a while), but most likely creating the `objc_class` attribute on instance creation would have resulted in infinite recursion.
* Added an `ObjCMetaClass` class, whose main purpose is to have a different name from `ObjCClass` when looking at the `repr` of a metaclass. Objective-C metaclasses aren't as powerful as Python's, so there's not a lot of other things that can be done with them.
* Made `ObjCInstance` return `ObjCClass`es when receiving a class pointer, and `ObjCClass` return `ObjCMetaClass`es when receiving a metaclass pointer.
* Merged the `ObjCClass` instance cache into the one for `ObjCInstance`s.
* Added unit tests for the behavior of classes and metaclasses (lookup by name or pointer, caching, etc.).

### DIFF
--- a/rubicon/objc/__init__.py
+++ b/rubicon/objc/__init__.py
@@ -3,7 +3,7 @@ __version__ = '0.2.3'
 from .objc import (
     objc, send_message, send_super,
     get_selector,
-    ObjCClass, ObjCInstance, NSObject,
+    ObjCInstance, ObjCClass, ObjCMetaClass, NSObject,
     objc_ivar, objc_property, objc_rawmethod, objc_method, objc_classmethod
 )
 
--- a/rubicon/objc/core_foundation.py
+++ b/rubicon/objc/core_foundation.py
@@ -164,7 +164,7 @@ def to_number(cfnumber):
 
     # NSDecimalNumber reports as a double. So does an NSNumber of type double.
     # In the case of NSDecimalNumber, convert to a Python decimal.
-    if numeric_type == kCFNumberDoubleType and cfnumber.__dict__['objc_class'].name == 'NSDecimalNumber':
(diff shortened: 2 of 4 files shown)


dgelessus:
Bump? @freakboy3742

freakboy3742:
My apologies - I've been travelling to PyCaribbean, so I've got behind on reviews of bigger patches. I'll take a look at this ASAP.

dgelessus:
Okay, no worries. Thanks for the update `:)`

Issue #26, opened 2017-02-17
Proposal for nicer init()


dimaqq:
Through trial and error I got following to pass:

```
tmp = ObjCClass("UINavigationController").alloc().initWithRootViewController_(root)
```

Wouldn't it be nice to get that data from call signature:

```
tmp = ObjCClass("UINavigationController").alloc().init(rootViewController=root)
```

And then, perhaps, move to:

```
UINavigationController = ObjCClass("UINavigationController")
tmp = UINavigationController(rootViewController=root)
```

dgelessus:
Pythonista's `objc_util` module supports calling Objective-C methods using keyword arguments like this, and it definitely helps with readability. Sounds like a good idea to me.

This would also be a good use case for [PEP 468](https://www.python.org/dev/peps/pep-0468/)'s order-preserving keyword arguments, which are available as of Python 3.6. Ordered keyword arguments are not necessary for this feature though, they are just an optimization - on older Python versions, the correct method name would need to be guessed, by trying every possible permutation of the keyword arguments, until a valid method name is found.

freakboy3742:
Completely agreed that this sounds like a good idea - it's just a matter of implementation. Since argument order isn't guaranteed (until Python 3.6, anyway), that means the lookup process will be (a) O(2^n) complexity for the number of arguments, and (b) potentially ambiguous, because the method `method(arga=1, argb=2)` will match `methodWithArga:Argb:` and `methodWithArgb:Arga:`. 

I'm not sure if the latter will be a problem in practice - and it becomes a non-issue under Python3.6 - but it's worth keeping in mind.

There's also a minor issue of how `initWithArgA:ArgB` maps to a method call; it would be nice if the method was `init(argA=..., argB=...)`; but we might need to compromise with `initWith(argA=..., argB=...)` - unless of course, the "with" becomes a magic keyword in method names.

dgelessus:
To keep it simple at first, maybe it would be good to require `initWithArgA(..., argB=...)`. Then the only thing that needs to be done is inserting the `:`s after each selector part (after `initWithArgA` and `argB`). More advanced "name guessing" could be added later. It would also halve the "guessing time" on Pythons with unordered kwargs.

dgelessus:
[...]

PR #31, opened 2017-03-12
Nicer method call syntax using keyword arguments

dgelessus:
Allows calling methods using keyword arguments, like `string.getCharacters(buffer, range=NSRange(0, 5))`. This implements most of #26.

As a side effect, `ObjCClass` now gets the entire array of method pointers on creation, instead of looking them up on demand. (However, `ObjCMethod` objects are *not* created for every method - doing so causes lots of "unknown restype" errors for methods that aren't used.) This appears to have no noticeable effect on performance (the unit tests take roughly 200ms for me before and after the change).

Other changes that were needed to implement this:

* The custom `c_void_p` subclasses (`SEL`, `objc_id`, `Class`, etc.) are now exported in `rubicon.objc`'s `__init__.py`.
* `ObjCClass`es now have a `superclass` attribute.
* `ObjCClass.__setattr__` no longer tries to look for an Objective-C setter if the attribute exists in `__dict__`. This allows setting attributes on an `ObjCClass` without manually accessing the `__dict__`.
* Better handling of null selectors, and simple unit tests for `SEL`. (This is not strictly necessary, but it helped while tracking down a segfault.)
* Minor simplification and clarification of how `register` works when creating new Objective-C subclasses. (This was also a result of me tracking down the segfault.)

### DIFF
--- a/rubicon/objc/__init__.py
+++ b/rubicon/objc/__init__.py
@@ -3,6 +3,7 @@ __version__ = '0.2.3'
 from .objc import (
     objc, send_message, send_super,
     get_selector,
+    SEL, objc_id, Class, IMP, Method, Ivar, objc_property_t,
     ObjCInstance, ObjCClass, ObjCMetaClass, NSObject,
     objc_ivar, objc_property, objc_rawmethod, objc_method, objc_classmethod
 )
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -1251,6 +1251,14 @@ class ObjCInstance(object):
 class ObjCClass(ObjCInstance, type):
     """Python wrapper for an Objective-C class."""
 
+    @property
+    def superclass(self):
+        super_ptr = objc.class_getSuperclass(self)
+        if super_ptr.value is None:
+            return None
+        else:
+            return ObjCClass(super_ptr)
+
     def __new__(cls, *args):
         """Create a new ObjCClass instance or return a previously created
         instance for the given Objective-C class.  The argument may be either
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -1016,7 +1016,7 @@ def objc_method(f):
 
     def register(cls, attr):
         name = attr.replace("_", ":")
-        cls.__dict__['imp_table'][name] = add_method(cls, name, _objc_method, encoding)
+        cls.__dict__['imp_keep_alive_table'][name] = add_method(cls, name, _objc_method, encoding)
 
     _objc_method.register = register
 
@@ -1041,7 +1041,7 @@ def objc_classmethod(f):
 
     def register(cls, attr):
         name = attr.replace("_", ":")
-        cls.__dict__['imp_table'][name] = add_method(cls.objc_class, name, _objc_classmethod, encoding)
+        cls.__dict__['imp_keep_alive_table'][name] = add_method(cls.objc_class, name, _objc_classmethod, encoding)
 
     _objc_classmethod.register = register
 
@@ -1114,8 +1114,8 @@ class objc_property(object):
 
         setter_name = 'set' + attr[0].upper() + attr[1:] + ':'
 
-        cls.__dict__['imp_table'][attr] = add_method(cls.__dict__['ptr'], attr, _objc_getter, getter_encoding)
-        cls.__dict__['imp_table'][setter_name] = add_method(cls.__dict__['ptr'], setter_name, _objc_setter, setter_encoding)
+        cls.__dict__['imp_keep_alive_table'][attr] = add_method(cls.__dict__['ptr'], attr, _objc_getter, getter_encoding)
+        cls.__dict__['imp_keep_alive_table'][setter_name] = add_method(cls.__dict__['ptr'], setter_name, _objc_setter, setter_encoding)
 
 
 def objc_rawmethod(f):
@@ -1123,7 +1123,7 @@ def objc_rawmethod(f):
 
     def register(cls, attr):
         name = attr.replace("_", ":")
-        cls.__dict__['imp_table'][name] = add_method(cls, name, f, encoding)
+        cls.__dict__['imp_keep_alive_table'][name] = add_method(cls, name, f, encoding)
     f.register = register
     return f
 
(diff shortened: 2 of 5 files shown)


rubicon/objc/objc.py
```diff
@@ -867,6 +873,40 @@ def __call__(self, receiver, *args, convert_args=True, convert_result=True):
...
+    def __call__(self, receiver, first_arg=_sentinel, **kwargs):
+        if first_arg is ObjCPartialMethod._sentinel:
+            if kwargs:
+                raise TypeError("Missing first (positional) argument")
+            
+            args = []
+            rest = frozenset()
+        else:
+            args = [first_arg]
+            # Add "" to rest to indicate that the method takes arguments
```
dimaqq: I see below it's being decoded for selectors.

Is this convention more general?

rubicon/objc/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .objc import (
     objc, send_message, send_super,
     get_selector,
+    SEL, objc_id, Class, IMP, Method, Ivar, objc_property_t,
```
dimaqq: `IMP` unused?

dgelessus: Yes, `IMP` is not used much. It's mainly used in some of the function declarations (such as `class_addMethod` or `method_getImplementation`) as a placeholder type for any function pointer. Because of this, `IMP` needs to be publicly available, to allow others to call the functions that have an `IMP` in their signature.

rubicon/objc/objc.py
dgelessus: I'm not sure what you mean, I don't see any decoding here... If you're talking about the "add empty string to rest" comment, that is specific to how I implemented the "partial method" dict, and not a general convention.

rubicon/objc/__init__.py
dimaqq: Oh, I guess I misunderstood... It was confusing enough that I looked elsewhere in the changeset and somehow decided that `# Selectors end in a colon` is where same convention was ultimately used.

To be honest I'm still not sure whether I got it right or wrong.

dgelessus:
Right, performance is definitely an issue when many classes are loaded. As a "stress test" I tried loading every class that starts with `NS` (that is, everything from Foundation and AppKit):

```python
import ctypes.util
import timeit

from rubicon import objc

c = ctypes.CDLL(ctypes.util.find_library("c"))
Foundation = ctypes.CDLL(ctypes.util.find_library("Foundation"))
AppKit = ctypes.CDLL(ctypes.util.find_library("AppKit"))

c.free.restype = None
c.free.argtypes = [ctypes.c_void_p]

objc.objc.objc_copyClassList.restype = ctypes.POINTER(objc.Class)
objc.objc.objc_copyClassList.argtypes = [ctypes.POINTER(ctypes.c_uint)]

out_count = ctypes.c_uint(0)
classes_ptr = objc.objc.objc_copyClassList(ctypes.byref(out_count))

names = []
try:
    for i in range(out_count.value):
        name = objc.objc.class_getName(classes_ptr[i])
        if name.startswith(b"NS"):
            names.append(name.decode("utf-8"))
finally:
    c.free(classes_ptr)

def our_main():
    for name in names:
        print("Loading", name)
        objc.ObjCClass(name)

if __name__ == "__main__":
    time = timeit.timeit("__main__.our_main()", setup="import __main__", number=1)
    print(f"{time:.3f} sec")
```

Before the changes, this ran in about 90 ms, now it runs in about 900 ms. That means it's now ten times slower, which is not great.

The original reason why I cached the method list on class creation was not related to the argument order. The issue I ran into was that `ObjCInstance.__getattr__` needs to know when an attribute does or doesn't exist. With normal method calls this is easy - the attribute name is the full method name, so you can look up the method with the Objective-C runtime and see if it exists.

With "partial methods" this is not possible, because at the time of `__getattr__` you only know the start of the method name. There's no way to ask Objective-C "are there any methods starting with this name", so you have to assume that there is one, and return a partial method for every attribute that you can't resolve otherwise. The issue with that is that you then don't have `AttributeError`s anymore. If you write `url.scheem` instead of `url.scheme` by accident, you get a partial method, because at that point it's not known if there are any methods that start with `scheem`. This breaks a test, doesn't play well with duck typing (`try`-`except AttributeError` or `hasattr`), and is not very intuitive IMHO. That's why I went with the upfront caching approach.

As for the performance, I already tried to do as little work as possible when caching the method names. The method list is only cached once when the class is first loaded, and not when it's loaded from the object cache. The methods are only stored as C `Method` pointers, not as full `ObjCMethod` objects, and the `ObjCPartialMethod` is only created once the method is actually called. If memory is an issue, I could also remove the method pointers from the cache. Then it would only contain the method names, which can then be used to look up the desired method when it's called.

freakboy3742:
Thanks for the explanation. I figured the limitation would be the absence of a "get all methods starting with ..." method - I've had similar limitations bridging into Java. 

And yes - while a 10x slowdown isn't ideal, it's also a lot better than I would have expected. Out of interest - how many classes are we talking about? If we're talking about all of Foundation and AppKit, 'm guessing it's quite a few.... 

I'm wondering if we could have the best of both worlds: Would it be possible to fully instantiate the method list, but only when the user attempts to access a method? i.e., defer populating the method list until you actually use it. That way, your "load everything" test should continue to run in close to 90ms; which means there won't be as much startup lag; you only pay the loading overhead for the classes you actually use, so there won't be a major performance penalty for just importing a class that isn't actually used; and any lag that *does* exist will be distributed over a longer period of time, as you only pay the startup cost when you start using a feature (for example, NSURL wouldn't be instantiated at startup, but when you first hit a JSON API).

dgelessus:
It should be possible to postpone getting the method list until a method is called. I'm not sure how much of a performance difference that would make though. Because any non-Python attribute could be a method, the method list would be loaded as soon as any non-Python attribute is accessed, even if turns out that the attribute is not a method. I'll give it a try though and see how it works out.

dgelessus:
This took a bit longer than expected, sorry about that. The change itself was quick to do, but I had some uncommitted work in a different branch that I needed to finish up first.

I've modified my test script a little, it now also includes the implementation-private `NS` classes that have underscores at the start of the name. On my system (OS X 10.11.6), there are 2579 `NS` classes in Foundation and AppKit, with 59558 methods in total. Loading all of them with the old version (method caching in `ObjCClass.__new__`) takes about 1.5 seconds.

With the new version (method caching in `ObjCInstance.__getattr__`), the test script only takes about 0.15 seconds. (Of course this isn't a fair comparison, because it only loads the classes and never uses any methods or properties.) The test suite runs in basically the same amount of time as before. I can't think of any case where the new version would be *slower* than the old one, and it definitely helps with classes that are loaded but not really used. I'd say that counts as a good idea, thanks for the suggestion. `:)`

PR #33, opened 2017-03-19
Add reference to NSObject.

Dayof:


Dayof:
I'll send another PR with the right recommendations.

PR #37, opened 2017-04-01
0.2.5 release

freakboy3742:


rubicon/objc/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = '0.2.3'
+__version__ = '0.2.5'
```
dgelessus: What happened to 0.2.4 though? I see it's on https://pypi.python.org/pypi/rubicon-objc, but in the repo it existed for two minutes before being replaced with 0.2.5.

dgelessus:
Also, is there a changelog for rubicon-objc somewhere? If not, I could see if I can make one.

rubicon/objc/__init__.py
freakboy3742: I messed up. I pushed the PyPI version before checking I had all the latest Github updates.

freakboy3742:
There isn't a changelog - but if you are motivated to write one, that would be amazing!

PR #41, opened 2017-05-12
Fixes #40 - Add handling for ambiguous property/method access

### LINKED ISSUE #40: 0.2.6 issues
Running Sierra, xCode 8.3

Running the [Briefcase Tutorial 2](https://github.com/pybee/briefcase/blob/50d0597d5f10789645b6822d51bf5cddefaf36ad/docs/tutorials/tutorial-1.rst), using toga 0.2.8, and rubicon-objc 0.2.6

```
$ python -m appname
  File "/Users/kmclaughlin/.pyenv/versions/3.5.1/lib/python3.5/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/kmclaughlin/.pyenv/versions/3.5.1/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/kmclaughlin/git/pybee/tutorial1/appname/appname/__main__.py", line 4, in <module>
    main().main_loop()
  File "/Users/kmclaughlin/git/pybee/tutorial1/venv/lib/python3.5/site-packages/toga_cocoa/app.py", line 137, in main_loop
    self._startup()
  File "/Users/kmclaughlin/git/pybee/tutorial1/venv/lib/python3.5/site-packages/toga_cocoa/app.py", line 89, in _startup
    self.resource_path = os.path.dirname(os.path.dirname(NSBundle.mainBundle().bundlePath))
TypeError: 'ObjCInstance' object is not callable
```

Substituting out rubicon for the version that was up yesterday, 0.2.4 (0.2.5 may have been skipped?)

```
$ pip install rubicon-objc==0.2.4
$ python -m appname
```

Gives me a valid f2c

freakboy3742:


tests/objc/Example.h
```diff
@@ -20,14 +20,21 @@ struct large {
...
 
 @interface Example : BaseExample {
+
     int _intField;
     Thing *_thing;
     id<Callback> _callback;
+    int _ambiguous;
 }
 
+#if (XCODE_VERSION >= 80000)
```
dgelessus: I think it might be better to use Clang's `__has_extension` macro here to check whether class properties are available. `__has_extension(objc_class_property)` returns `1` if you can use them and `0` if not.

Makefile
```diff
@@ -1,8 +1,13 @@
+XCODE_MAJOR=`xcodebuild -version | head -n 1 | cut -f 2 -d' ' | cut -d. -f1`
+XCODE_MINOR=`xcodebuild -version | head -n 1 | cut -f 2 -d' ' | cut -d. -f2`
+XCODE_MICRO=`xcodebuild -version | head -n 1 | cut -f 2 -d' ' | cut -d. -f3`
+XCODE_VERSION=$(shell printf '%d' $(XCODE_MAJOR))$(shell printf '%02d' $(XCODE_MINOR))$(shell printf '%02d' $(XCODE_MICRO))
```
dgelessus: Would it make sense to `export` the `XCODE_VERSION` definition? Then it could be reused inside Python using `os.environ['XCODE_VERSION']` without having to parse Xcode version again in Python. That would also make it impossible for the shell and Python `XCODE_VERSION` to have different values.

tests/test_rubicon.py
```diff
@@ -260,6 +269,46 @@ def test_mutator_like_method(self):
...
+        # In Sierra, mainBundle was turned into a class property.
+        # It should be accessible as a method
+        NSBundle = ObjCClass('NSBundle')
+        try:
+            NSBundle = ObjCClass('NSBundle')
+            NSBundle.mainBundle()
+        except:
+            self.fail("Properties shadowing methods should be OK.")
+
+    @unittest.skipIf(XCODE_VERSION < 80000, "class properties not supported before XCode 8")
```
dgelessus: `__has_extension(objc_class_property)` can't be used from Python of course. Maybe we can store it in a global constant variable in the Objective-C code, and then query it using `ctypes`?

tests/test_rubicon.py
```diff
@@ -260,6 +269,46 @@ def test_mutator_like_method(self):
...
+        obj1 = Example.alloc().init()
+        self.assertEqual(obj1.ambiguous(), 42)
+
+        # In Sierra, mainBundle was turned into a class property.
+        # It should be accessible as a method
+        NSBundle = ObjCClass('NSBundle')
+        try:
+            NSBundle = ObjCClass('NSBundle')
+            NSBundle.mainBundle()
+        except:
```
dgelessus: Noo, not a bare `except`, think of the poor tracebacks that get lost there! `:)` I'm not sure, doesn't a test automatically count as failed if it raises an unexpected exception?

tests/objc/Example.h
freakboy3742: Ah - so *that* is the flag I spent yesterday morning looking for :-)

tests/test_rubicon.py
freakboy3742: It's raised as an Error, not a Failure; it's a subtle difference, but one I like to preserve.

tests/test_rubicon.py
freakboy3742: If there was a good reason to access this during normal runtime, I'd agree - but since it's just an internal testing thing, I think it's easier to just re-evaluate the constant at test time.

freakboy3742:
In a discussion on Gitter with @dgelessus, I proposed three possible solutions for this problem:

a. Make property access always be a method. 
b. Give up on the property version if there is a method version. This removes the ambiguity by making access always via the method if a method exists. 
c. Look into using a lazy object in the ambiguous case, that supports being called by "return self".

David's response was that (a) is the best option here. (b) is in practice always equivalent to (a), because by convention Objective-C property getter methods are named like the property. (c) might be possible by writing a "proxy" class that forwards everything to the underlying object, but can also be called to return that object. That is never a perfect illusion though, it breaks as soon as you look at the object's type or pass it to most built-in functions.

Overnight, I thought of a fourth option: an analog of `getattr()` (say, `getprop()`), that does the lookup, but allows for a fallback mode. This would also work, but is a bit messy.

This PR currently implements a new option - explicit declaration. 

My reasoning is that forcing the use of a method would work, but didn't feel especially Pythonic. And while declaration isn't pretty, it works well enough for this use case. 

Effectively, the problem is caused when Apple decides that a method with no arguments (class or instance) should be "promoted" to property status. In this specific case, the method in question is `NSBundle.mainBundle()`. NSBundle is declared as part of Toga; so there isn't really any public API to consider; to fix our underlying problem, we'll patch Toga to predeclare `mainBundle` as a class property, and then use it as a property regardless of OS version. 

Proactively, we can then look at our API usage, and consider whether any 0-argument methods are possible candidates to be converted into properties in the future, and pre-declare them. This cleans up *our* API, because we can use obvious properties *as* properties (for example, `NSColor.redColor()` can become `NSColor.redColor`); and protects us from the whims of future Apple API changes.

freakboy3742:
Thanks @glasnt 

@dgelessus If you have concerns/reservations about this approach, let me know - I'm more than happy to revisit this discussion. Getting a fix - any fix - in place for PyCon was a concern.

dgelessus:
Sorry I wasn't able to look at this earlier, yesterday was a bit busy for me. This approach looks good - the default behavior is never wrong or unstable, and explicitly declared properties are always guaranteed to work, regardless of whether the runtime knows of them as properties or not. There's little that can go wrong here.

Again, sorry about the holdup, I wasn't aware that you were under time pressure for PyCon. I hope things are working now.

PR #42, opened 2017-05-22
Add shim code so that NS{,Mutable}{Array,Dict} will act like list and dict.

### LINKED ISSUE #15: Pythonic interface for NSArray / NSMutableArray

jeamland:
Fixes #15

### DIFF
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -1180,6 +1180,18 @@ class ObjCInstance(object):
     def objc_class(self):
         return ObjCClass(objc.object_getClass(self))
 
+    @classmethod
+    def _select_mixin(cls, object_ptr):
+        nsarray = objc.objc_getClass(b'NSArray')
+        if send_message(object_ptr, 'isKindOfClass:', nsarray):
+            return ObjCListInstance
+
+        nsdictionary = objc.objc_getClass(b'NSDictionary')
+        if send_message(object_ptr, 'isKindOfClass:', nsdictionary):
+            return ObjCDictInstance
+
+        return cls
+
     def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
         """Create a new ObjCInstance or return a previously created one
         for the given object_ptr which should be an Objective-C id."""
@@ -1208,10 +1220,10 @@ class ObjCInstance(object):
             # Special case for ObjCClass to pass on the class name, bases and namespace to the type constructor.
             self = super().__new__(cls, _name, _bases, _ns)
         else:
+            cls = cls._select_mixin(object_ptr)
             self = super().__new__(cls)
         super(ObjCInstance, type(self)).__setattr__(self, "ptr", object_ptr)
         super(ObjCInstance, type(self)).__setattr__(self, "_as_parameter_", object_ptr)
-
         # Store new object in the dictionary of cached objects, keyed
         # by the (integer) memory address pointed to by the object_ptr.
         cls._cached_objects[object_ptr.value] = self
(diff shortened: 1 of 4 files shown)


rubicon/objc/objc.py
```diff
@@ -1314,6 +1326,65 @@ def __setattr__(self, name, value):
                 super(ObjCInstance, type(self)).__setattr__(self, name, value)
 
 
+class ObjCListInstance(ObjCInstance):
```
dgelessus: ~~It might be useful to inherit from `collections.abc.Sequence` here. It provides default implementations for most standard Python sequence methods, like `find`, `count`, etc.~~

PS: Nevermind, that would cause name conflicts. `-[NSArray count]` returns the number of objects in the array, but Python `list.count` is a method that counts how often an object is present in a list.

rubicon/objc/objc.py
```diff
@@ -1314,6 +1326,65 @@ def __setattr__(self, name, value):
...
+    def __getitem__(self, item):
+        if item >= self.count:
+            raise IndexError('list index out of range')
+        return self.objectAtIndex(item)
+
+    def __setitem__(self, item, value):
+        try:
+            self.insertObject_atIndex_(value, item)
```
dgelessus: I'm not sure, but isn't [`-[NSMutableArray insertObject:atIndex:]`](https://developer.apple.com/reference/foundation/nsmutablearray/1416682-insertobject?language=objc) for *inserting* an extra item into the array (like `list.insert`)? [`-[NSMutableArray replaceObjectAtIndex:withObject:]`](https://developer.apple.com/reference/foundation/nsmutablearray/1414510-replaceobjectatindex?language=objc) looks like the correct method for reassigning an object at a given index.

rubicon/objc/objc.py
```diff
@@ -1314,6 +1326,65 @@ def __setattr__(self, name, value):
...
+        except AttributeError:
+            name = self.objc_class.name
+            raise TypeError("'%s' object does not support item assignment" % name)
+
+    def __len__(self):
+        return self.count
```
dgelessus: I'm a bit confused why this works. If I'm not mistaken, #41 makes all properties behave as methods by default, and to access them using attribute syntax you need to declare them with `declare_property`. That would mean that by default you need to write `self.count()` to get the array's item count, because `self.count` would be the method object. Any ideas what is happening here @freakboy3742?

(The same question applies to `ObjCDictInstance.__len__` below.)

rubicon/objc/objc.py
```diff
@@ -1314,6 +1326,65 @@ def __setattr__(self, name, value):
...
+
+    def __contains__(self, item):
+        return self.containsObject_(item)
+
+
+class ObjCDictInstance(ObjCInstance):
+    def __getitem__(self, item):
+        v = self.objectForKey_(item)
+        if v is None:
+            raise KeyError(repr(item))
```
dgelessus: Minor issue: the default Python behavior is to pass the missing key directly to `KeyError`, and not its `repr`:

```python
>>> import sys
>>> d = {}
>>> d[42]
  File "<pyshell#1>", line 1, in <module>
    d[42]
KeyError: 42
>>> sys.last_value.args[0]
42
```

rubicon/objc/objc.py
jeamland: I think you're right. I'll fix this.

rubicon/objc/objc.py
jeamland: Will fix.

PR #43, opened 2017-05-25
Add support for registering custom type encodings

### LINKED ISSUE #14: Type handling - No restype encoding for...

dgelessus:
Fixes #14.

This moves most of the type-encoding-related code from `objc.py` to `types.py`, and adds a documented API for adding custom conversions between Objective-C type encodings and `ctypes` types.

The API may seem a bit overkill, but the complexity is unfortunately necessary, because it's not always possible to map exactly one Objective-C encoding to exactly one ctype. For example, all of `c_char_p`, `POINTER(c_char)`, `POINTER(c_byte)`, and `POINTER(c_ubyte)` have the encoding `*`. The same issue exists in the other direction - on a 64-bit machine, the encodings `l` and `q` both correspond to `c_long` (because `c_long == c_longlong` on 64-bit).

Because of this, there are two functions for registering an encoding/ctype mapping:

* `register_preferred_encoding(encoding, ctype)` registers conversions from `encoding` to `ctype` and vice versa, and replaces any existing conversions in each direction.
* `register_encoding(encoding, ctype)` does the same, but only adds the conversion for each direction if there isn't already one for that direction.

In many cases, encodings do map one-to-one to ctypes, and it doesn't make a difference which function is used. But sometimes one encoding needs to be explicitly set as preferred. For example, this:

```python
register_encoding(b'*', POINTER(c_byte))
register_encoding(b'*', c_char_p)
```

makes both `POINTER(c_byte)` and `c_char_p` encode to `*`, and `*` decode to `POINTER(c_byte)`. On the other hand, this:

```python
register_encoding(b'*', POINTER(c_byte))
register_preferred_encoding(b'*', c_char_p)
```

makes `*` decode to `c_char_p` instead, but still allows both `POINTER(c_byte)` and `c_char_p` to encode to `*`. (For some more examples of how this works in practice, see `types.py`, where all the primitive types are registered.)

That said, suggestions for improving the API or the documentation are welcome - I would have preferred to have fewer functions, but I couldn't figure out a simpler way to support all the corner cases well.

### DIFF
--- a/rubicon/objc/types.py
+++ b/rubicon/objc/types.py
@@ -59,7 +59,7 @@ def encoding_to_ctype(encoding):
     """Return ctypes type for an encoded Objective-C type."""
     
     # Remove qualifiers, as documented in Table 6-2 here:
-    # https://developer.apple.com/library/prerelease/content/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html
+    # https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html
     encoding = encoding.lstrip(b"NORVnor")
     
     try:
--- a/rubicon/objc/types.py
+++ b/rubicon/objc/types.py
@@ -18,8 +18,6 @@ _any_arm = ('ARM' in platform.version())
 __arm64__ = (_any_arm and __LP64__)
 __arm__ = (_any_arm and not __LP64__)
 
-PyObjectEncoding = b'{PyObject=@}'
-
 
 def encoding_for_ctype(vartype):
     typecodes = {
@@ -44,6 +42,7 @@ if __LP64__:
     NSRangeEncoding = b'{_NSRange=QQ}'
     UIEdgeInsetsEncoding = b'{UIEdgeInsets=dddd}'
     NSEdgeInsetsEncoding = b'{NSEdgeInsets=dddd}'
+    PyObjectEncoding = b'^{_object=q^{_typeobject}}'
 else:
     NSInteger = c_int
     NSUInteger = c_uint
@@ -54,6 +53,7 @@ else:
     NSRangeEncoding = b'{_NSRange=II}'
     UIEdgeInsetsEncoding = b'{UIEdgeInsets=ffff}'
     NSEdgeInsetsEncoding = b'{NSEdgeInsets=ffff}'
+    PyObjectEncoding = b'^{_object=i^{_typeobject}}'
 
 NSIntegerEncoding = encoding_for_ctype(NSInteger)
 NSUIntegerEncoding = encoding_for_ctype(NSUInteger)
--- a/rubicon/objc/types.py
+++ b/rubicon/objc/types.py
@@ -92,6 +92,17 @@ def register_preferred_encoding(encoding, ctype):
     _encoding_to_ctype_map[encoding] = ctype
     _ctype_to_encoding_map[ctype] = encoding
 
+def with_preferred_encoding(encoding):
+    """Decorator for registering a preferred conversion between the given encoding and the decorated type.
+    This is equivalent to calling register_preferred_encoding.
+    """
+    
+    def _with_preferred_encoding_decorator(ctype):
+        register_preferred_encoding(encoding, ctype)
+        return ctype
+    
+    return _with_preferred_encoding_decorator
+
 def register_encoding(encoding, ctype):
     """Register an additional conversion between an Objective-C type encoding and a ctypes type.
     If a conversion already exists in one or both directions, it is not overwritten.
@@ -99,6 +110,17 @@ def register_encoding(encoding, ctype):
     
     _encoding_to_ctype_map.setdefault(encoding, ctype)
     _ctype_to_encoding_map.setdefault(ctype, encoding)
+
+def with_encoding(encoding):
+    """Decorator for registering a conversion between the given encoding and the decorated type.
+    This is equivalent to calling register_encoding.
+    """
+    
+    def _with_encoding_decorator(ctype):
+        register_encoding(encoding, ctype)
+        return ctype
+    
+    return _with_encoding_decorator
 
 def unregister_encoding(encoding):
     """Unregister the conversion between an Objective-C type encoding and its corresponding ctypes type.
@@ -244,38 +266,38 @@ NSZoneEncoding = b'{_NSZone=}'
 register_preferred_encoding(PyObjectEncoding, py_object)
 
 
+@with_preferred_encoding(b'^?')
 class UnknownPointer(c_void_p):
     """Placeholder for the b'^?' "unknown pointer" type. Not to be confused with a b'^v' void pointer.
     Usually a b'^?' is a function pointer, but because the encoding doesn't contain the function signature, you need to manually create a CFUNCTYPE with the proper types, and cast this pointer to it.
     """
-register_preferred_encoding(b'^?', UnknownPointer)
 
 # from /System/Library/Frameworks/Foundation.framework/Headers/NSGeometry.h
+@with_preferred_encoding(NSPointEncoding)
 class NSPoint(Structure):
     _fields_ = [
         ("x", CGFloat),
         ("y", CGFloat)
     ]
 CGPoint = NSPoint
-register_preferred_encoding(NSPointEncoding, NSPoint)
 
 
+@with_preferred_encoding(NSSizeEncoding)
 class NSSize(Structure):
     _fields_ = [
         ("width", CGFloat),
         ("height", CGFloat)
     ]
 CGSize = NSSize
-register_preferred_encoding(NSSizeEncoding, NSSize)
 
 
+@with_preferred_encoding(NSRectEncoding)
 class NSRect(Structure):
     _fields_ = [
         ("origin", NSPoint),
         ("size", NSSize)
     ]
 CGRect = NSRect
-register_preferred_encoding(NSRectEncoding, NSRect)
 
 
 def NSMakeSize(w, h):
@@ -297,12 +319,12 @@ CGPointMake = NSMakePoint
 
 
 # iOS: /System/Library/Frameworks/UIKit.framework/Headers/UIGeometry.h
+@with_preferred_encoding(UIEdgeInsetsEncoding)
 class UIEdgeInsets(Structure):
     _fields_ = [('top', CGFloat),
                 ('left', CGFloat),
                 ('bottom', CGFloat),
                 ('right', CGFloat)]
-register_preferred_encoding(UIEdgeInsetsEncoding, UIEdgeInsets)
 
 def UIEdgeInsetsMake(top, left, bottom, right):
     return UIEdgeInsets(top, left, bottom, right)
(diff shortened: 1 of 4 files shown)


rubicon/objc/types.py
```diff
@@ -18,17 +18,194 @@
...
-        c_ubyte: b'C', c_uint: b'I', c_ushort: b'S', c_ulong: b'L', c_ulonglong: b'Q',
-        c_float: b'f', c_double: b'd', c_bool: b'B', c_char_p: b'*', c_void_p: b'@',
-        py_object: PyObjectEncoding
-    }
-    return typecodes.get(vartype, b'?')
+_type_to_ctype_map = {
+    int: c_int,
+    float: c_float,
+    bool: c_bool,
+    # This is not really correct - Python strings are Unicode - but we need to keep this for compatibility.
```
freakboy3742: Compatibility with what? If the issue is existing versions of Rubicon, I wouldn't consider any of the APIs stable yet, so if we need to fix something to get us out of a corner case, I'd prefer to see that than start a workaround.

rubicon/objc/types.py
```diff
@@ -18,17 +18,194 @@
...
+
+def get_type_to_ctype_map():
+    """Get a copy of all currently registered type-to-ctype conversions as a mapping."""
+    
+    return dict(_type_to_ctype_map)
+
+_encoding_to_ctype_map = {}
+_ctype_to_encoding_map = {}
+
+def encoding_to_ctype(encoding):
```
freakboy3742: As a purely stylistic, bikeshed thing, I prefer the `ctype_for_encoding(encoding)` naming scheme, rather than `encoding_to_ctype(encoding)`. To me, the former is easier to read as "this produces a ctype".

rubicon/objc/types.py
```diff
@@ -54,16 +232,25 @@ def encoding_for_ctype(vartype):
...
 NSZoneEncoding = b'{_NSZone=}'
 
+register_preferred_encoding(PyObjectEncoding, py_object)
+
+
+class UnknownPointer(c_void_p):
+    """Placeholder for the b'^?' "unknown pointer" type. Not to be confused with a b'^v' void pointer.
+    Usually a b'^?' is a function pointer, but because the encoding doesn't contain the function signature, you need to manually create a CFUNCTYPE with the proper types, and cast this pointer to it.
+    """
+register_preferred_encoding(b'^?', UnknownPointer)
```
freakboy3742: There's a part of me that instinctively thinks this should be written as a class decorator - but it might not be worth the effort.

freakboy3742:
Oh - and if some documentation for exactly what the encodings mean was to magically materialise... I can't say i'd complain :-)

rubicon/objc/types.py
dgelessus: Yep, I was thinking of compatibility with previous versions. But if that's not a concern, I'll happily remove that.

If the API isn't stable yet, I would also be in favor of removing the mappings for `int` and `float`, because these Python types don't really match the C types they are currently translated to. Python `int` is arbitrary-precision, while C `int` is (practically always) 32-bit. Python `float` is actually the same precision as C `double`, but it is mapped to C `float`. Both of these can cause unintended loss of precision. And in an Objective-C context, it's more common to use generic typedefs like `NSInteger`, `NSUInteger`, `CGFloat`, etc. rather than a specific primitive C type.

rubicon/objc/types.py
dgelessus: I can see that... also the right side of the name is closer to the argument, so `ctype_for_encoding(b"i")` is "the right way around" in that sense too. I would be fine with switching the naming scheme back again, but unfortunately that breaks the similarity with `get_encoding_to_ctype_map` and `get_ctype_to_encoding_map` (and changing the order there goes against the standard "keys left, values right" order of dicts...)

Argh, naming things. :)

rubicon/objc/types.py
dgelessus: That's a great idea actually, and it makes a lot of sense, because in practice most users would use this for custom `ctypes.Structure` subclasses. I'm not sure what the decorator could be called though. `with_encoding`? `for_encoding`? Reusing the `register_encoding` function is probably not a good idea, it would cause obscure errors if the argument count is wrong.

dgelessus:
@freakboy3742 [This](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html) is Apple's official documentation on the Objective-C type encodings. Certain details are not explained there, so I know some things only from StackOverflow, experimentation, and looking at the encodings of random methods in the standard frameworks.

rubicon/objc/types.py
```diff
@@ -18,17 +18,194 @@
...
+
+def ctype_to_encoding(ctype):
+    """Return the Objective-C type encoding for the given ctypes type."""
+    
+    try:
+        return _ctype_to_encoding_map[ctype]
+    except KeyError:
+        raise ValueError('No type encoding known for ctype {}'.format(ctype))
+
+def register_preferred_encoding(encoding, ctype):
```
dgelessus: While we're here bikeshedding names... I was thinking about renaming this to `register_encoding`, and the current `register_encoding` to `register_alternative_encoding`, or something similar.

The reason being that I later want to make struct types be created automatically for unknown struct encodings. In that case, when a user registers a custom struct type, they probably want it to take priority over the auto-generated one, so it makes more sense for overwriting to be the default behavior.

On the other hand, there is also the possibility of users registering custom types that correspond to integers, such as enums. In that case, the correct behavior would be to *not* overwrite existing encodings, otherwise every integer would be converted to that enum... I think this case is much less likely though, because we already "unwrap" enum constants automatically when passed to an Objective-C method. There's also `enum.IntEnum` and `enum.IntFlag` which can be used for int-compatible enums.

rubicon/objc/types.py
freakboy3742: Yes, Python's int is arbitrary precision, but there's an issue of practicality; int is still a common datatype. Providing *a* mapping for int, even if it's one with known limitations, would seem preferable to not providing a mapping at all.

As for float - modifying it to point to c_double would seem to make sense to me, rather than removing it entirely.

Neither change would affect the `NSInteger` et al use case, but it would preserve basic functionality for a non-zero number of cases.

rubicon/objc/types.py
freakboy3742: :-)

Surely the maps can be renamed too? `_encoding_for_ctype_map = {}` and `get_encoding_for_ctype_map()`?

rubicon/objc/types.py
freakboy3742: Ah - but you could make it a dual-purpose decorator. The simplest implementation would be:
```
def register_encoding(encoding, ctype=None):
    def _register(cls):
        _encoding_to_ctype_map.setdefault(encoding, cls)
        _ctype_to_encoding_map.setdefault(cls, encoding)
        return cls
    if ctype is None:
        return _register
    else:
        _register(ctype)
```
Then in usage:
```
@register_encoding('xxx')
class MyStruct(Structure):
   ...
```
or
```
register_encoding('xxx', MyStruct)
```

rubicon/objc/types.py
freakboy3742: Hrm. The current names make sense to me given the current implementation; `register_preferred_encoding` is a good name for the "overwrite any existing representation" implementation of the code. How would the implementation change in the approach you envisage?

rubicon/objc/types.py
dgelessus: The issue with a dual-purpose function like that is that if you forget the second argument in a normal call (like `register_encoding(b"foo")`) it silently does nothing (because it only returns the `_register` function) and there's no way to give any sort of warning in that case.

rubicon/objc/types.py
dgelessus: The implementation of the register functions would not change. Most likely I'd change `encoding_to_ctype` to automatically create, register and return a new `ctypes.Structure` subclass when it sees an unknown struct encoding. When that has happened and the user later creates their own custom structure for the same encoding, they need to use the "overwrite" behavior, to make sure that their custom type takes precedence over the auto-generated one.

That makes the "overwrite" behavior much more commonly necessary. In many cases it doesn't really matter which one is used - if there is no existing mapping, overwriting and not overwriting is the same. On the other hand the "don't overwrite" behavior is only needed rarely (see previous comment). Because of that I think it makes more sense to make the "overwrite" behavior the default.

PR #49, opened 2017-07-13
Automatic array/struct/union type creation

dgelessus:
This extends `rubicon.objc.types.ctype_for_encoding` to automatically create array, struct and union types if they aren't registered yet. This means that methods with complicated types in their signature can now be used without having to first define and register all of them.

The auto-generated struct/union types often don't have useful field names - if a field doesn't have a name in the encoding, `field_n` is substituted (where `n` is a number counting up from zero). It is possible to compile Objective-C code in a way that also includes field names in structure/union encodings, and some of Apple's frameworks are compiled in that way. I haven't been able to get Clang to do that though. The only thing I've found is the flag `-Xclang -fencode-extended-block-signature`, but that didn't seem to change the encodings at all.

I've also added support for some other unusual encodings, like typed object pointers (`@"NSString"`), typed block pointers `@?<signature>`, and error messages for types not supported by `ctypes` (128-bit integers, complex numbers, atomic types).

Most of the details about the encoding format are taken from the `ASTContext::getObjCEncodingForTypeImpl` method from [`AST/ASTContext.h`](https://llvm.org/svn/llvm-project/cfe/trunk/lib/AST/ASTContext.cpp) from the Clang source code.

### DIFF
--- a/rubicon/objc/types.py
+++ b/rubicon/objc/types.py
@@ -55,7 +55,8 @@ _encoding_for_ctype_map = {}
 
 def _end_of_encoding(encoding, start):
     """Find the end index of the encoding starting at index start.
-    The encoding is not validated very extensively. There are no guarantees what happens for invalid encodings - an error may be raised, or a bogus end index may be returned.
+    The encoding is not validated very extensively. There are no guarantees what happens for invalid encodings;
+    an error may be raised, or a bogus end index may be returned.
     """
     
     if start < 0 or start >= len(encoding):
--- a/rubicon/objc/types.py
+++ b/rubicon/objc/types.py
@@ -53,6 +53,172 @@ def get_ctype_for_type_map():
 _ctype_for_encoding_map = {}
 _encoding_for_ctype_map = {}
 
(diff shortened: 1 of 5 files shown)


rubicon/objc/types.py
```diff
@@ -53,6 +53,172 @@ def get_ctype_for_type_map():
 _ctype_for_encoding_map = {}
 _encoding_for_ctype_map = {}
 
+def _end_of_encoding(encoding, start):
+    """Find the end index of the encoding starting at index start.
+    The encoding is not validated very extensively. There are no guarantees what happens for invalid encodings - an error may be raised, or a bogus end index may be returned.
```
phildini: Nit: Can we break this into multiple lines? We try to follow at least the Django Style Guide (https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/)

phildini:
I'm happy there are tests for this, and overall I don't see any huge issues, although @freakboy3742 is best equipped to review this, I think.

PR #50, opened 2017-07-24
Added rudimentary objc block support for method arguments

ojii:
Adds support to call objc-blocks that are arguments to nsobjects defined in Python or are properties of objects in objc. Does *not* add support to create your own objc-blocks to pass to objc.

Usage:

```
# `instance` is an instance of ObjCInstance we got from objc
# The block signature is assumed to be `(void (^)(int, int))`
block = ObjCBlock(instance, c_void_p, c_int, c_int)
block(1, 2)
```

### DIFF
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -1652,15 +1652,51 @@ class ObjCBlockStruct(Structure):
         ('flags', c_int),
         ('reserved', c_int),
         ('invoke', CFUNCTYPE(c_void_p, c_void_p)),
+        ('descriptor', c_void_p),
     ]
+
+
+def cast_block_descriptor(block):
+    descriptor_fields = [
+        ('reserved', c_ulong),
+        ('size', c_ulong),
+    ]
+    if block.has_helpers:
+        descriptor_fields.extend([
+            ('copy_helper', CFUNCTYPE(c_void_p, c_void_p, c_void_p)),
+            ('dispose_helper', CFUNCTYPE(c_void_p, c_void_p)),
+        ])
+    if block.has_signature:
+        descriptor_fields.extend([
+            ('signature', c_char_p),
+        ])
+    descriptor_struct = type(
+        'ObjCBlockDescriptor',
+        (Structure, ),
+        {'_fields_': descriptor_fields}
+    )
+    return cast(block.struct.contents.descriptor, POINTER(descriptor_struct))
 
 
 class ObjCBlock:
     def __init__(self, instance, return_type, *arg_types):
         self.instance = instance
-        self.block = cast(self.instance.ptr, POINTER(ObjCBlockStruct))
-        self.block.contents.invoke.restype = return_type
-        self.block.contents.invoke.argtypes = arg_types
+        self.struct = cast(self.instance.ptr, POINTER(ObjCBlockStruct))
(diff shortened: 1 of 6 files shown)


rubicon/objc/objc.py
```diff
@@ -1644,3 +1644,23 @@ def finalize(self, cmd) -> None:
...
+class ObjCBlockStruct(Structure):
+    _fields_ = [
+        ('isa', c_void_p),
+        ('flags', c_int),
+        ('reserved', c_int),
+        ('invoke', CFUNCTYPE(c_void_p, c_void_p)),
+    ]
+
+
+class ObjCBlock:
```
dgelessus: Would it be possible to extend `ObjCInstance` here? That way the `ObjCBlock` object can be passed into methods, and can have methods called on it (which is useful for `retain`, `release` and `copy`).

tests/objc/Blocks.m
```diff
@@ -0,0 +1,39 @@
...
+-(int) blockExample {
+  BlockDelegate *delegate = self.delegate;
+  NSLog(@"Delegate is: %@", delegate);
+
+  [delegate exampleMethod:^(int a, int b){
+    self.value = a + b;
+  }];
+  return self.value;
+}
+@end
```
dgelessus: Nitpick: the two methods above are indented with 2 spaces, they should use 4 spaces. There's also no newline at the end of the file.

rubicon/objc/objc.py
dgelessus: Update: After discussing this on Gitter, we found out that this isn't as easy as I thought.

Because of `ObjCInstance`'s caching mechanism, it's not possible to have two `ObjCInstance`s for the same object, even if they have different classes. This means that in this case the `ObjCBlock` constructor would return a normal `ObjCInstance` if one already existed for the object (which is very common).

It's not possible to handle blocks inside the `ObjCInstance` constructor, like we do with `ObjCClass`, `ObjCListInstance`, etc., because there is no way to see at runtime whether an object is a block or not.

tests/test_rubicon.py
```diff
@@ -1160,3 +1160,21 @@ def test_update3(self):
         self.assertEqual(d['four'], 'FIVE')
         self.assertEqual(len(d), len(self.py_dict) + 1)
 
+
+class BlockTests(unittest.TestCase):
+    def test_block_property(self):
+        BlockPropertyExample = ObjCClass("BlockPropertyExample")
+        instance = BlockPropertyExample.alloc().init()
+        result = ObjCBlock(instance.blockProperty, c_int, c_int, c_int)(1, 2)
```
freakboy3742: This part of the API gives me pause; it's the only place in public API that we'd be publicly referencing ctypes raw types. Could/should this refer to Python types that we then convert to their native variants?

ojii:
> The test cases you've covered are good, and other than the native Ctypes thing I flagged, this patch looks good.

I will change the code to accept python types as well and do the usual conversion.

> But doesn't this miss the use case of an ObjC API that accepts a block as an argument, and you want to pass in a Python callable as that argument?

This PR only covers the "receiving" blocks part. "sending" blocks is a lot more complicated (though I plan to try to implement that in the future).

ojii:
closed in favor of #51.

PR #51, opened 2017-07-25
Support for receiving objc blocks as argument to python methods

ojii:
Improved version of #50

This PR implements:

* API to manually transform an `ObjCInstance` which is an objc block to a callable object, using `ObjCBlock(objc_instance, return_type, *arg_types)(*args)`.
* API to automatically transform objc blocks given as arguments to python methods into a callable, as long as that block has a signature (and not bitfields), using the type annotation `objc_block`.

This PR does **not** implement:

* Passing Python callables as blocks to objc.

### DIFF
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -1652,15 +1652,51 @@ class ObjCBlockStruct(Structure):
         ('flags', c_int),
         ('reserved', c_int),
         ('invoke', CFUNCTYPE(c_void_p, c_void_p)),
+        ('descriptor', c_void_p),
     ]
+
+
+def cast_block_descriptor(block):
+    descriptor_fields = [
+        ('reserved', c_ulong),
+        ('size', c_ulong),
+    ]
+    if block.has_helpers:
+        descriptor_fields.extend([
+            ('copy_helper', CFUNCTYPE(c_void_p, c_void_p, c_void_p)),
+            ('dispose_helper', CFUNCTYPE(c_void_p, c_void_p)),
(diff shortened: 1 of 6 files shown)


rubicon/objc/objc.py
```diff
@@ -1644,3 +1660,201 @@ def finalize(self, cmd) -> None:
...
+    'L': c_ulong,
+    'Q': c_ulonglong,
+    'f': c_float,
+    'd': c_double,
+    'B': c_bool,
+    'v': c_void_p,
+    '*': c_char_p,
+    '#': Class,
+    ':': SEL,
+}
```
freakboy3742: Is this duplicated data? Can't we use the registered encoding map from types.py?

ojii: @dgelessus mentioned that I might've duplicated some of the existing work. I will look into it once I'm back at a mac. I initially tried to re-use stuff from types.py, but then opted to just implement stuff myself first to get a better understanding of how things work and get to a working state from which I can improve.

dgelessus: #53 adds a `ctypes_for_method_encoding` function to decode a method signature into its return/argument types. This should also work for block signatures, which are in exactly the same format as method signatures (as far as I can tell).

Previously we used some Objective-C runtime functions to get the return/argument types from method signatures, but they only work with a `Method` object, and not with a "standalone" signature like we have for blocks. So I wrote a similar function in Python, based on the existing type decoding infrastructure, which works with a simple bytestring instead of a `Method` object.

PR #54, opened 2017-08-21
Added support for sending blocks to objc

ojii:
This adds support for sending blocks to objc. It supports both "automagical python callable to objc block style" (which in non-trivial cases can segfault due to python GC) as well as "explicit create a block object style" through the `rubicon.objc.Block`. 

Magical style (functions need to be fully annotated):

```python

def myblock(arg1: arg1type, arg2: arg2type) -> return_type:
    ...

objc_instance.ObjcMethod_(myblock)
```

Explicit style:

```python
from rubicon.objc import Block

myblock = Block(lambda a, b: ..., return_type, argtype1, argtype2)
objc_instance.ObjcMethod_(myblock)
```

### DIFF
--- a/rubicon/objc/objc.py
+++ b/rubicon/objc/objc.py
@@ -1728,7 +1728,7 @@ class ObjCBlockInstance(ObjCInstance):
         return self.block(*args)
 
 
-_NSConcreteGlobalBlock = ObjCClass("__NSGlobalBlock__")
+_NSConcreteGlobalBlock = (c_void_p * 32).in_dll(c, "_NSConcreteGlobalBlock")
 
 
 NOTHING = object()
@@ -1767,7 +1767,7 @@ class Block:
         self.cfunc_type = CFUNCTYPE(restype, c_void_p, *signature)
 
         self.literal = BlockLiteral()
-        self.literal.isa = _NSConcreteGlobalBlock.ptr
+        self.literal.isa = addressof(_NSConcreteGlobalBlock)
         self.literal.flags = BlockConsts.HAS_STRET | BlockConsts.HAS_SIGNATURE
         self.literal.reserved = 0
         self.cfunc = self.cfunc_type(self.wrapper)
@@ -1776,7 +1776,6 @@ class Block:
         self.descriptor.reserved = 0
         self.descriptor.size = sizeof(BlockLiteral)
 
-        encoding_for_ctype(restype)
         self.descriptor.signature = encoding_for_ctype(restype) + b'@?' + b''.join(
             encoding_for_ctype(arg) for arg in signature
         )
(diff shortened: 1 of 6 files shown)


rubicon/objc/objc.py
```diff
@@ -1696,3 +1726,62 @@ def __call__(self, *args):
 class ObjCBlockInstance(ObjCInstance):
     def __call__(self, *args):
         return self.block(*args)
+
+
+_NSConcreteGlobalBlock = ObjCClass("__NSGlobalBlock__")
```
dgelessus: The `__NSGlobalBlock__` class isn't actually documented anywhere. `_NSConcreteGlobalBlock` is [documented](https://clang.llvm.org/docs/Block-ABI-Apple.html#high-level), but not required to be an Objective-C class (you're only supposed to use its address).

So the officially supported way to use this would be:

```python
_NSConcreteGlobalBlock = (c_void_p * 32).in_dll(c, "_NSConcreteGlobalBlock")
# ... when setting the isa ...
self.literal.isa = addressof(_NSConcreteGlobalBlock)
```

(`_NSConcreteGlobalBlock` is declared as a `void *[32]` in [`Block.h`](https://opensource.apple.com/source/libclosure/libclosure-67/Block.h.auto.html).)

rubicon/objc/objc.py
```diff
@@ -1696,3 +1726,62 @@ def __call__(self, *args):
...
+        self.literal.isa = _NSConcreteGlobalBlock.ptr
+        self.literal.flags = BlockConsts.HAS_STRET | BlockConsts.HAS_SIGNATURE
+        self.literal.reserved = 0
+        self.cfunc = self.cfunc_type(self.wrapper)
+        self.literal.invoke = cast(self.cfunc, c_void_p)
+        self.descriptor = BlockDescriptor()
+        self.descriptor.reserved = 0
+        self.descriptor.size = sizeof(BlockLiteral)
+
+        encoding_for_ctype(restype)
```
dgelessus: This discards the return value from the `encoding_for_ctype` call - is this intentional? If so, it would be good to add a comment.

rubicon/objc/types.py
```diff
@@ -235,7 +238,8 @@ def ctype_for_encoding(encoding):
 
 def encoding_for_ctype(ctype):
     """Return the Objective-C type encoding for the given ctypes type."""
-    
+    if isinstance(ctype, POINTER_TYPE):
+        return b'^' + encoding_for_ctype(ctype._type_)
```
dgelessus: This should probably be in the `except KeyError` block below, so that pointer types that are in `_encoding_for_ctype_map` are looked up there. For example `POINTER(c_char)` should encode as `*`, not as `^c`.

rubicon/objc/objc.py
ojii: looks like leftover debugging code from me.

PR #55, opened 2017-09-03
Add objc_const

dgelessus:
As suggested on Gitter. Nothing major, just a convenience method so `objc_id` doesn't need to be used directly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,10 +1,17 @@
 Changelog
 =========
 
+(next version)
+--------------
+
+* Added support for creating Objective-C blocks from Python callables. (ojii)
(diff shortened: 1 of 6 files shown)


dgelessus:
I didn't have a strong reason for the `in_dll` name other than sticking with the `ctypes` convention. I like the `ObjCConst` suggestion, it's shorter to type, and naming `ObjCInstance` explicitly in every call is probably not necessary.

(Making it a class method on `ObjCInstance` technically allows you to call it on subclasses too, like `ObjCClass.in_dll(...)`. This adds some extra type checking, but it's almost never usable in practice - I've never seen a global `Class` variable in any Objective-C header.)

dgelessus:
Done. (Also ninja-updated the changelog.)

PR #57, opened 2017-09-05
Reorganize test directory.

freakboy3742:


freakboy3742:
The `test_` bit is indeed a unittest requirement. The test discovery process requires that the filename start with `test`, `test_`, or end with `Test` or `_test` - even if it's in a `tests` directory. It is configurable, but it's a helpful convention - it means you can have `utils.py` and other non-test files in the same directory without getting picked up as tests.

PR #58, opened 2017-09-06
Improved handling of native Python data types

freakboy3742:
This is necessary to make sure retain/release operations occur on the right objects.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 (next version)
 --------------
 
+* Corrected support to storing strings and other objects as properties on Python-define Objective-C classes.
 * Added support for creating Objective-C blocks from Python callables. (ojii)
 * Added an ``objc_const`` convenience function to look up global Objective-C object constants in a DLL.
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 (next version)
 --------------
 
+* Added support for passing Python lists as Objective-C NSArray arguments, and Python dicts as Objective-C NSDictionary arguments.
 * Corrected support to storing strings and other objects as properties on Python-defined Objective-C classes.
 * Added support for creating Objective-C blocks from Python callables. (ojii)
 * Added an ``objc_const`` convenience function to look up global Objective-C object constants in a DLL.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,7 +4,7 @@ Changelog
 (next version)
 --------------
 
-* Corrected support to storing strings and other objects as properties on Python-define Objective-C classes.
+* Corrected support to storing strings and other objects as properties on Python-defined Objective-C classes.
 * Added support for creating Objective-C blocks from Python callables. (ojii)
 * Added an ``objc_const`` convenience function to look up global Objective-C object constants in a DLL.
 
--- a/rubicon/objc/core_foundation.py
+++ b/rubicon/objc/core_foundation.py
@@ -218,15 +218,13 @@ def from_value(value):
         return NSDecimalNumber.from_decimal(value)
     elif isinstance(value, dict):
         dikt = NSMutableDictionary.alloc().init()
-        if value is not None:
-            for k, v in value.items():
-                dikt.setObject_forKey_(v, k)
+        for k, v in value.items():
+            dikt.setObject_forKey_(v, k)
         return dikt
     elif isinstance(value, list):
         array = NSMutableArray.alloc().init()
-        if value is not None:
-            for v in value:
-                array.addObject(v)
+        for v in value:
+            array.addObject(v)
         return array
     else:
         return value
--- a/rubicon/objc/core_foundation.py
+++ b/rubicon/objc/core_foundation.py
@@ -197,6 +197,13 @@ class NSDecimalNumber(object):
         return ObjCInstance(cls.constructor(cast(cls.objc_class, objc_id), cls.selector, at(value.to_eng_string())))
 
 
+NSArray = ObjCClass('NSArray')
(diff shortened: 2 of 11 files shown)


CHANGELOG.rst
```diff
@@ -4,6 +4,7 @@ Changelog
 (next version)
 --------------
 
+* Corrected support to storing strings and other objects as properties on Python-define Objective-C classes.
```
dgelessus: `Python-define` -> `Python-defined`

setup.cfg
```diff
@@ -0,0 +1,16 @@
...
+[flake8]
+# https://flake8.readthedocs.org/en/latest/
+exclude=docs/*,build/*
+max-complexity = 15
+max-line-length = 119
+
+[pep8]
+# https://pep8.readthedocs.org/en/latest/
+# W503 raised if operators are after line break
+ignore = W503
```
dgelessus: If we're going to use flake8, there are a few more warnings that need to be disabled, such as the ones about star-imports. I have a flake8 config lying around in a git stash somewhere, I should make a PR for that later...

rubicon/objc/objc.py
```diff
@@ -948,8 +948,7 @@ def setter(_self, new):
         def _objc_getter(objc_self, objc_cmd, *args):
             from .core_foundation import at
             py_self = ObjCInstance(objc_self)
-            args = convert_method_arguments(getter_encoding, args)
-            result = getter(py_self, *args)
+            result = getter(py_self, *[ObjCInstance(a) for a in args])
```
dgelessus: I'm a bit confused why the `*args` are needed for the getter and setter. A getter always takes no arguments, and a setter always takes exactly one argument, right?

setup.cfg
freakboy3742: I haven't tied flake8 into the build process (yet); so this is just groundwork. If you've got any suggestions for things to turn on/off, PRs welcome.

rubicon/objc/objc.py
freakboy3742: Yeah - I couldn't think of any reason that `*args` would be used in a getter; I preserved it because it was effectively a no op. But I suppose not passing the argument at all would be even more of a no-op :-)

The setter can also be cleaned up in the same vein, since it should always have exactly 1 argument.

rubicon/objc/objc.py
```diff
@@ -948,8 +948,7 @@ def setter(_self, new):
         def _objc_getter(objc_self, objc_cmd, *args):
```
dgelessus: Can the `*args` here and on `_objc_setter` be replaced with zero/one explicit parameters as well?

freakboy3742: Turns out - yes! And the method is a lot cleaner for it.

rubicon/objc/core_foundation.py
```diff
@@ -209,6 +216,18 @@ def from_value(value):
         return at(value.decode('utf-8'))
     elif isinstance(value, Decimal):
         return NSDecimalNumber.from_decimal(value)
+    elif isinstance(value, dict):
+        dikt = NSMutableDictionary.alloc().init()
+        if value is not None:
```
dgelessus: This is always true, because of the `isinstance` check before.

rubicon/objc/core_foundation.py
```diff
@@ -209,6 +216,18 @@ def from_value(value):
...
+            for k, v in value.items():
+                dikt.setObject_forKey_(v, k)
+        return dikt
+    elif isinstance(value, list):
+        array = NSMutableArray.alloc().init()
+        if value is not None:
```
dgelessus: Same as above.

CHANGELOG.rst
```diff
@@ -4,6 +4,7 @@ Changelog
 (next version)
 --------------
 
+* Added support for passing Python lists as Objective-C NSArray arguments, and Python dicts as Objective-C NSDictionary arguments.
```
dgelessus: :+1:

rubicon/objc/core_foundation.py
```diff
@@ -209,6 +216,18 @@ def from_value(value):
         return at(value.decode('utf-8'))
     elif isinstance(value, Decimal):
         return NSDecimalNumber.from_decimal(value)
+    elif isinstance(value, dict):
```
dgelessus: Should we perhaps check for `collections.abc.Mapping` instead of `dict` here, so it works with other mapping types (`defaultdict`, `OrderedDict`, `Counter`, etc.) as well? On the other hand that would require an extra check so we don't "convert" `ObjCDictionaryInstance`s to Python and back again. Maybe at the top we should just add `if isinstance(value, ObjCInstance): return value`, so that Objective-C objects are never touched, no matter how much they pretend to be native Python objects.

rubicon/objc/core_foundation.py
freakboy3742: Good point.

rubicon/objc/core_foundation.py
freakboy3742: Turns out is isn't needed (at least for the examples you give):
```
>>> from collections import OrderedDict, defaultdict, Counter
>>> isinstance(OrderedDict(), dict)
True
>>> isinstance(defaultdict(), dict)
True
>>> isinstance(Counter(), dict)
True
```
They're all subclasses of the builtin dict, so they'll get converted automatically.

dgelessus: Okay, `collections.ChainMap`. :P In any case, I would be fine with either option. Even if we require an actual `dict` or subclass, users can use an explicit `dict(non_dict_mapping)`.

freakboy3742: Exactly - plus, it's a little more explicit, which helps avoid nasty surprises when you have a dict-like class that isn't *quite* a dict.

PR #59, opened 2017-09-09
List and dict improvements

freakboy3742:
This improves handling for the bridging of Python lists and dictionaries, including the use of primitive values as keys and values.

### DIFF
--- a/rubicon/objc/core_foundation.py
+++ b/rubicon/objc/core_foundation.py
@@ -67,7 +67,7 @@ libcf.CFAttributedStringCreate.argtypes = [CFAllocatorRef, CFStringRef, CFDictio
 
 # Core Foundation type to Python type conversion functions
 def CFSTR(string):
-    return ObjCInstance(cf.CFStringCreateWithCString(
+    return ObjCInstance(libcf.CFStringCreateWithCString(
         None, string.encode('utf-8'), kCFStringEncodingUTF8,
     ))
 
--- a/rubicon/objc/core_foundation.py
+++ b/rubicon/objc/core_foundation.py
@@ -16,7 +16,7 @@ from .types import CFIndex, CFRange, CGFloat
 
 # CORE FOUNDATION
 
-cf = cdll.LoadLibrary(util.find_library('CoreFoundation'))
+libcf = cdll.LoadLibrary(util.find_library('CoreFoundation'))
 
 CFTypeID = c_ulong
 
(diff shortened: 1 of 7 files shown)


dgelessus:
I like the automatic boxing of primitive values. This allows automatic conversion of lists/dicts containing Python primitives, right? (e. g. Python `{"zero": 0, "one": 1, "two": 2}` is converted to Objective-C `@{@"zero": @0, @"one": @1, @"two": @2}`)

I'm not sure about the custom `__repr__` and `__str__` for the Objective-C collections. `ObjCInstance` already provides custom `__repr__` and `__str__` methods, which call the Objective-C `debugDescription` method. (`__str__` returns the `debugDescription` unmodified, `__repr__` includes some additional info about the Python and Objective-C classes and addresses.) The `debugDescription` for arrays and dictionaries already provides a good string representation IMHO, so the custom `__repr__` and `__str__` seem a bit redundant to me.

freakboy3742:
Yes - the autoboxing of primitives means you can pass in almost any Python list, dict, or primitive and it (or the things it contains) will get converted to "object form".

As for `__str__` and `__repr__` - Those came out of the debugging process for the auto boxing :-) Yes - the Objective C objects fall back onto debugDescription et al, but the output format leaves a lot to be desired. It tells you the memory address of a list, but not the things that are *in* the list. And OK, sure - you can easily find that out and print it, but it doesn't make for a very Pythonic experience.

If you think it's worth preserving memory details in the __repr__, I can definitely see the benefit - but in Python space, `__str__` really should be a "useful" output, and the `debugDescription` definitely isn't that.

dgelessus:
That's odd, I have no issues with the default `__repr__` and `__str__` showing me the array contents:

```python
>>> NSMutableArray = objc.ObjCClass("NSMutableArray")
>>> arr = NSMutableArray.array()
>>> arr.append(objc.at("spam"))
>>> arr.append(objc.at("ham"))
>>> arr.append(objc.at("eggs"))
>>> arr
Out[9]: 
<rubicon.objc.runtime.ObjCMutableListInstance 0x104be3e10: __NSArrayM at 0x10076cba0: <__NSArrayM 0x10076cba0>(
spam,
ham,
eggs
)
>
>>> print(str(arr))
<__NSArrayM 0x10076cba0>(
spam,
ham,
eggs
)
```

To be honest, the default `__repr__` is a bit long, maybe we should shorten that. For example I don't think the Python `id` matters in practice, and we can probably leave out the Objective-C object's address as well (if it matters, the `debugDescription` probably includes it anyway).

freakboy3742:
For me, the `__repr__` is probably OK; it's verbose, yes - but then it's a repr - it's meant to be a peek into internals. But `__str__` is meant to be human readable - something you could put in a user interface. The format currently being used doesn't meet that requirement.

dgelessus:
Should we change the generic `__str__` to call `description` instead of `debugDescription`? The `description` is usually a bit less verbose, for example for arrays it doesn't include the class name and address:

```python
print(arr.description)
(
    spam,
    ham,
    eggs
)
```

freakboy3742:
That sounds like it might work. I'll make that adjustment.

rubicon/objc/core_foundation.py
```diff
@@ -67,8 +67,9 @@
 
 # Core Foundation type to Python type conversion functions
 def CFSTR(string):
-    return ObjCInstance(libcf.CFStringCreateWithCString(
-            None, string.encode('utf-8'), kCFStringEncodingUTF8))
+    return ObjCInstance(cf.CFStringCreateWithCString(
```
dgelessus: That should be `libcf` instead of `cf` (the change got lost in the merge).

tests/test_NSArray.py
```diff
@@ -25,8 +25,6 @@
 
 class NSArrayMixinTest(unittest.TestCase):
     py_list = ['one', 'two', 'three']
-    name = 'NSArray'
-
     def make_array(self, contents=None):
```
dgelessus: Missing blank line, `flake8` does not like that.

tests/test_NSArray.py
```diff
@@ -23,7 +25,6 @@
 
 class NSArrayMixinTest(unittest.TestCase):
     py_list = ['one', 'two', 'three']
-
     def make_array(self, contents=None):
```
dgelessus: Missing blank line, `flake8` does not like that.

dgelessus:
The Travis log also shows [a segfault](https://travis-ci.org/pybee/rubicon-objc/jobs/276231034#L1707), but that might just be a consequence of the `NameError` right before.

PR #61, opened 2017-09-14
Emulate object_isClass if unavailable

dgelessus:
`object_isClass` only exists since OS X 10.10 (Yosemite) and iOS 8, but it is very easy to emulate on older systems. I tested this on my OS X 10.7 (Lion) iMac and it works. Yes I know that's old, but I don't have any other systems to test on - the next newer system I have after that is my OS X 10.11 MacBook.

Fun fact: this is the *only* fix needed to get rubicon-objc working on OS X 10.7. (The test suite doesn't run fully, but that's mostly because of missing property declarations.)

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -336,13 +336,21 @@ libobjc.objc_setAssociatedObject.argtypes = [objc_id, c_void_p, objc_id, c_int]
 
 ######################################################################
 
-# BOOL object_isClass(id obj)
-libobjc.object_isClass.restype = c_bool
-libobjc.object_isClass.argtypes = [objc_id]
-
 # Class object_getClass(id object)
 libobjc.object_getClass.restype = Class
 libobjc.object_getClass.argtypes = [objc_id]
+
+# object_isClass exists as a native function only since OS X 10.10 and iOS 8.
+# If unavailable, we emulate it: an object is a class iff its class is a metaclass.
+try:
+    object_isClass = libobjc.object_isClass
+except AttributeError:
+    def object_isClass(obj):
+        return libobjc.class_isMetaClass(libobjc.object_getClass(obj))
+else:
+    # BOOL object_isClass(id obj)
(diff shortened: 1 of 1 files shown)


freakboy3742:
That's pretty neat! Does Travis still provide a 10.[789] test environment? If so, it would be nice to turn those on to verify and maintain compatibility with earlier macOS versions.

dgelessus:
[The docs](https://docs.travis-ci.com/user/reference/osx/#OS-X-Version) list `xcode6.4` (OS X 10.10) as the earliest version. There were older versions before that, but they are "retired", whatever that means. See [the history of the data file for the table](https://github.com/travis-ci/docs-travis-ci-com/commits/master/_data/xcodes.yml). I have no idea if older versions can still be used, I think I'll make a test PR that asks for a few old test environments and see what happens.

Any opinions on which old OS X versions we should "officially" support? OS X 10.7 seems reasonable in terms of work that need to be done - it's really mostly declaring a bunch of properties. There are also a couple of issues that appear on 10.7 but not current versions, due to some code that is not quite correct, but happens to work in most cases. I plan to fix those no matter what, since they could "reappear" on newer versions too.

freakboy3742:
I don't have any specific opinions on what we should support, other than the utilitarian "as old as we can without expending excessive effort." Based on a quick survey, 10.7 seems like a good point for the cutoff, with between 2 and 5% of market share depending on the data you're reading.

dgelessus:
I agree with "as old as we can without expending excessive effort". 10.7 seems to be the first version where the libobjc API is mostly complete, on earlier versions random bits simply don't exist yet. If the availability version numbers in the headers are to be trusted, 10.6 and 10.5 might also be possible with some extra work. 10.4 is an absolute mess and probably dead enough that we don't care.

dgelessus:
After testing in #62 I found out that it isn't possible to use the "retired" Travis environments. When trying to do so, Travis simply uses the default environment instead.

PR #62, opened 2017-09-14
[WIP] Travis testing

dgelessus:
Please do not merge. Just testing some things with Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,16 @@
-matrix:
-  include:
-    # OSX 10.10 Yosemite
-    - os: osx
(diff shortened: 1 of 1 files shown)


dgelessus:
So far Travis hasn't complained. If this runs through, it means we can have Travis environments for 10.9 and 10.8 as well. (I'll squash before merging.)

dgelessus:
Nevermind. It seems that when using an unsupported version number, Travis silently uses the default version (which at the moment is Xcode 7.3 on OS X 10.11 El Capitan). No old Xcodes for us then.

PR #63, opened 2017-09-15
Update and shorten .travis.yml

dgelessus:
This adds an Xcode 9 test environment, and shortens the `matrix` section by not repeating the parts that are always the same.


dgelessus:
Okay, I'm guessing that merge was supposed to go the other way. :)

PR #65, opened 2017-09-16
Remove F403 (star imports) from flake8 ignore list

dgelessus:
There are no more star imports in the code, they were all replaced with explicit `from`-`import`s, and I apparently overlooked that.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,12 +10,11 @@ exclude=*.egg-info/*,.git/*,.tox/*,__pycache__,build/*,docs/*
 max-complexity = 20
 max-line-length = 119
 
-# Currently ignored:
-# E133: closing bracket missing indentation  # does not match code style
-# E226: missing whitespace around arithmetic operator  # does not match code style
-# F403: 'from module import *' used; unable to detect undefined names  # currently used almost everywhere
-# W503: line break occurred before a binary operator  # does not match code style
-ignore = E133,E226,F403,W503
+# The following issues are ignored because they do not match our code style:
+# E133: closing bracket missing indentation
+# E226: missing whitespace around arithmetic operator
+# W503: line break occurred before a binary operator
+ignore = E133,E226,W503
 
 [isort]
 combine_as_imports = true


dgelessus:
On this topic, would it be desirable to replace `from ctypes import (...)` with `import ctypes` and use fully qualified names instead? That way we wouldn't have to maintain a long list of names that we use at the top of the file. PyCharm can do this automatically, but IMHO it's much cleaner to use a single module import rather than lots of `from`-`import`s.

freakboy3742:
Personally, I'm less fussed about the difficulty of managing a long list of imports than I am about the fidelity of reading code that says `ctypes.foo(ctypes.bar + ctypes.ctypes.ctypes.ctypes...`. For me, namespace level imports are good when you're dealing with occasional function calls, or you need to disambiguate two identical (or even very similar) named symbols; but in this case, the ctypes imports are all core symbols relating to the use of C in Python, so they are both unambiguous, and frequently used.

dgelessus:
True, you end up writing `ctypes` a lot. (I am disappointed that `ctypes.ctypes.ctypes.ctypes` is not a real module.) Though I would argue that we shouldn't do `from ctypes import util`, because `util` is extremely generic. Should we change that to `from ctypes.util import find_library` (which is the only thing that we use from `ctypes.util`)?

dgelessus:
Somewhat related, should we provide [`__all__` lists](https://docs.python.org/3.6/tutorial/modules.html#importing-from-a-package) for the `rubicon.objc` submodules? They are useful to ensure that users who star-import our modules only get the symbols that we define ourselves, and not everything that we import as a dependency (like all the `ctypes` stuff, which will change depending on what each module does or doesn't use internally). Or are the submodules purely internal, and users should only ever import `rubicon.objc`?

freakboy3742:
To my reading, the submodules are internals that shouldn't be imported by themselves; however, it certainly couldn't hurt to add `__all__` definitions for the exact reason that you've mentioned.

Issue #70, opened 2017-09-17
Should ObjCClass stay a subclass of type?


dgelessus:
At the moment `ObjCClass` extends `ObjCInstance` and `type`. As far as I can tell, the fact that it extends `type` is not used or needed for anything at the moment. A metaclass don't need to subclass `type`, it's enough to have its constructor accept the signature of `type`'s constructor (name, bases, namespace).

It's also misleading, because users might expect to be able to define Python methods on Objective-C classes (by not decorating them with `@objc_method`). This is not possible at the moment, because instances of user-defined Objective-C classes are actually instances of `ObjCInstance`, and not of the user's `ObjCClass` subclass. We could theoretically support this though - similar to how an ObjC `NSArray` becomes a Python `ObjCListInstance`, we could make an ObjC `UserSubclass` become a Python `UserSubclass` too. (The logic in `ObjCInstance._select_mixin` would probably need to be optimized though, I don't think a chain of `isKindOfClass:` scales well.)

PR #81, opened 2017-11-10
Support mapping Objective-C classes to custom ObjCInstance subclasses

dgelessus:
Yes, it's another type mapping registry. This is a replacement for `ObjCClass._select_mixin`, which is not extensible and not very efficient as it grows (as I noted in #70). The new system uses a mapping to keep track of which Python type corresponds to which Objective-C class, which is more efficient and allows easy addition of extra types. This is now used internally for `ObjCProtocol` and `ObjC[Mutable](List|Dict)Instance`, but theoretically it can also be used externally.

`ObjCClass`, `ObjCMetaClass` and `ObjCBlockInstace` are still special-cased in `ObjCInstance.__new__`. They cannot be registered using the new system, because there is no single Objective-C base class for all classes, metaclasses or blocks.

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -1143,6 +1143,76 @@ def objc_rawmethod(f):
 
 ######################################################################
 
+_type_for_objcclass_map = {}
+
+
+def type_for_objcclass(objcclass):
+    """Look up the ObjCInstance subclass used to represent instances of the given Objective-C class in Python.
+
+    If the exact Objective-C class is not registered, each superclass is also checked, defaulting to ObjCInstance
+    if none of the classes in the superclass chain is registered. Afterwards, all searched superclasses are registered
+    for the ObjCInstance subclass that was found.
+    """
+
+    if isinstance(objcclass, ObjCClass):
+        objcclass = objcclass.ptr
+
+    superclass = objcclass
+    traversed_classes = []
+    pytype = ObjCInstance
+    while superclass.value is not None:
+        try:
+            pytype = _type_for_objcclass_map[superclass.value]
+        except KeyError:
+            traversed_classes.append(superclass)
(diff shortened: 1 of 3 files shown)


rubicon/objc/runtime.py
```diff
@@ -1208,7 +1259,7 @@ def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
             if is_block:
                 cls = ObjCBlockInstance
             else:
-                cls = cls._select_mixin(object_ptr)
+                cls = type_for_objcclass(libobjc.object_getClass(object_ptr))
```
freakboy3742: Is there any reason that the `is_block` check couldn't be merged into `type_for_objcclass`?

dgelessus: Yes, unfortunately. `is_block` is assigned at the very top of `ObjCInstance.__new__`, and is used to remember whether the `object_ptr` that was originally passed to the constructor was of type `objc_block`. That information would be lost otherwise, because `object_ptr` is cast to `objc_id` in the next line.

Even if that wasn't the case, `type_for_objcclass` takes a *class* pointer, and the block check happens on the *instance* pointer. And the reason why we have to rely on the type of the instance pointer at all is because there's no good way to tell at runtime whether an object is a block or not.

dgelessus:
Yeah, Mac Travis is as reliable as usual. Four of the five builds ran fine, the remaining one failed to `brew update` so I whacked it with the restart wrench a few minutes ago. The [backlog](https://traviscistatus.com/) is acceptable at the moment ("only" 200 builds), so you could wait for that one build to finish. Or just merge right away, I doubt the last build would fail if the others all ran fine.

dgelessus:
Nevermind, the build ran just now. That was much quicker than I expected.

PR #84, opened 2017-11-20
Fixed #1 -- Allow returning structures by copy.

### LINKED ISSUE #1: Can't define callback with method that returns structure
ctypes allows you to define a method that returns structures. However, a _callback_ method can't return a structure - only primitive types.

This is reported as [Python issue5710](https://bugs.python.org/issue5710). 

However, there are some methods in the Cocoa and iOS APIs that require returning NSSize or NSRect. `NSWindow windowWillResize:toSize:` is one notable example.

So - we either need to fix the Python issue to allow structures to be returned by callbacks, or we need a workaround. One thought for a workaround - it may be possible to do this by casting a structure to a void pointer to get it through ctypes, then reversing the cast on the other side.

The return_struct branch [contains a test](https://github.com/pybee/rubicon-objc/blob/return_struct/tests/test_rubicon.py#L347) for this problem; the test fails when the callback method on the handler class is registered.

freakboy3742:
This workaround essentially treats "return structure by copy" as a Python object pointer - which means it's treated like a block of memory; but the Structure subclasses know how to interpret that block of memory, so it seems to work.

It all seems a little *too* easy... so i'm wondering what I've missed...

### DIFF
--- a/rubicon/objc/async.py
+++ b/rubicon/objc/async.py
@@ -1,18 +1,17 @@
 """PEP 3156 event loop based on CoreFoundation"""
 
+import threading
 from asyncio import (
-    DefaultEventLoopPolicy, coroutines, events, tasks, unix_events
+    DefaultEventLoopPolicy, coroutines, events, tasks, unix_events,
 )
 from ctypes import CFUNCTYPE, POINTER, Structure, c_int, c_void_p
-import threading
 
 from .core_foundation import (
-    CFAbsoluteTime, CFAllocatorRef, CFDataRef, CFOptionFlags,
-    CFStringRef, CFTimeInterval, kCFAllocatorDefault, libcf
+    CFAbsoluteTime, CFAllocatorRef, CFDataRef, CFOptionFlags, CFStringRef,
+    CFTimeInterval, kCFAllocatorDefault, libcf,
 )
 from .runtime import objc_const, objc_id
 from .types import CFIndex
-
 
 __all__ = [
     'EventLoopPolicy',
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -5,7 +5,7 @@ from ctypes import (
     CDLL, CFUNCTYPE, POINTER, ArgumentError, Array, Structure, addressof,
     alignment, byref, c_bool, c_char_p, c_double, c_float, c_int, c_int32,
     c_int64, c_longdouble, c_size_t, c_uint, c_uint8, c_ulong, c_void_p, cast,
-    sizeof, util,
+    py_object, sizeof, util,
 )
 from enum import Enum
 
@@ -629,6 +629,9 @@ def send_super(receiver, selName, *args, **kwargs):
     restype = kwargs.get('restype', c_void_p)
     argtypes = kwargs.get('argtypes', None)
 
+    if type(restype) == type(Structure):
+        restype == py_object
+
     send = libobjc['objc_msgSendSuper']
     send.restype = restype
     if argtypes:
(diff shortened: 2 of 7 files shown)


dgelessus:
This is an interesting solution :) Unfortunately, while it does work in this case, it doesn't solve our main problem.

By changing the return type to `py_object`, you allow the method to return any Python object. In this case, the Python object happens to be a `NSRect` instance, but it could just as well be a `str` or a `complex` or a plain `object()`. `ctypes` gets the address of the object, passes it through C as a pointer, and converts the pointer back to an object on the other side.

This doesn't work when you need to implement/override a method that has to return a structure. On the C side, a `py_object` is just a pointer to a structure (namely CPython's `PyObject`). If you return a `PyObject *` instead of (for example) a `NSRect`, the actual value doesn't match the expected type, and the caller probably gets garbage.

freakboy3742:
AFAICT, this problem exists right now. On the Python side, annotations are just annotations; you're free to return whatever object you want. The following code will *run*:
```    
class MyObject(NSObject):
    @objc_method
    def tester1(self) -> int:
        return "Hello"

    @objc_method
    def tester2(self):
        return 42
```

`tester1` returns the memory address of the string; the `tester2` causes a segfault.

So, sure - if you didn't return an `NSSize` on a method annotated with `NSSize`, you're going to get unpredictable behavior - but that's true regardless.

However, I have found a problem; I've just updated the test case to demonstrate. 

I *really* don't want to have to patch ctypes to make this work... :-(

dgelessus:
The reason why both methods of `MyObject` "work" is because both `rubicon.objc` and `ctypes` do a lot of "automatic casting".

In `tester1`, the string `"Hello"` is converted to a `NSString` `ObjCInstance` by `rubicon.objc`. The internal form of an `ObjCInstance` is an `objc_id`, which is a `c_void_p` subclass, and `ctypes` "helpfully" takes the address and truncates it to an integer. So the result of `tester1()` are the lower 32 bits of the address of an `NSString` containing the text "Hello". If we change the return type to `ctypes.c_long`, you get the full address and can see what's actually happening:

```python
>>> o2 = MyObject2.new()
>>> o2.tester1()
4312358736
>>> hex(_)
'0x101095f50'
>>> objc.ObjCInstance(0x101095f50)
<rubicon.objc.runtime.ObjCInstance 0x103b68e48: __NSCFString at 0x101095f50: Hello>
```

In `tester2`, `ctypes` converts the integer `42` into a `c_void_p`, which `rubicon.objc` then tries to use as the address of an Objective-C object, which of course segfaults. If we disable result conversion, we can again see what's really going on:

```python
>>> o2.tester2(convert_result=False)
<objc_id object at 0x103bcd620>
>>> _.value
42
```

This doesn't work with *any* Python object though. For example, if you try to return a `complex`, this happens:

```python
>>> class MyObject3(objc.NSObject):
...     @objc.objc_classmethod
...     def tester3(self):
...         return 1j
...     
>>> MyObject3.tester3()
  File "_ctypes/callbacks.c", line 259, in 'converting callback result'
TypeError: cannot be converted to pointer
Exception ignored in: <function objc_classmethod.<locals>._objc_classmethod at 0x103bced90>

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

However this has nothing to do with the `py_object` issue. Here's an example of what happens when you use `py_object` as the return type:

```python
>>> import ctypes
>>> from rubicon import objc
>>> thing = objc.NSMakeRect(1, 2, 3, 4)
>>> class PyReturner(objc.NSObject):
...     @objc.objc_classmethod
...     def getPy(self) -> ctypes.py_object:
...         return thing
...     
>>> PyReturner.getPy()
<rubicon.objc.types.NSRect object at 0x103b13950>
>>> objc.send_message(PyReturner, "getPy", restype=ctypes.c_void_p, argtypes=[])
c_void_p(4356913488)
>>> id(thing)
4356913488
>>> objc.send_message(PyReturner, "getPy", restype=objc.NSRect, argtypes=[])

Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
```

When calling `getPy`, the default return type is `py_object` (this is determined from the Objective-C type encoding), and all conversions happen correctly in that case. If we manually change the return type to `c_void_p`, we can see what happens internally - returning a Python object from a function with return type `py_object` is equivalent to returning the Python object's `id` as a `c_void_p`. And if we try to use `NSRect` as the return type, we get a segfault, because a pointer cannot be converted to a structure.

This explanation is a bit long, so if you have any questions, please do ask (here or on Gitter).

freakboy3742:
@dgelessus has a much more robust (and working!) solution in #85.

PR #86, opened 2017-11-23
Fixes for return-by-copy for larger structures

freakboy3742:
Three features here:

* Clarify the error message when structures have already been registered
* Require pre-registration of structures that will be returned, rather than on-demand. The on-demand approach fails as soon as you have 2 methods returning the same structure (even if they're on different classes)
* Use the -stret variant of `objc_msgSendSuper` when returning a structure. For some reason, the non-stret version works fine for NSPoint and NSSize, but fails on NSRect.

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -715,10 +715,6 @@ def add_method(cls, selName, method, encoding):
     signature = [ctype_for_type(tp) for tp in encoding]
     assert signature[1] == objc_id  # ensure id self typecode
     assert signature[2] == SEL  # ensure SEL cmd typecode
-    if signature[0] is not None and issubclass(signature[0], (Structure, Union)):
-        # Patch struct/union return types to make them work in callbacks.
-        # See the source code of the ctypes_patch module for details.
-        ctypes_patch.make_callback_returnable(signature[0])
     selector = SEL(selName)
     types = b"".join(encoding_for_ctype(ctype) for ctype in signature)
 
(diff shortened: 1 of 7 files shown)


rubicon/objc/runtime.py
```diff
@@ -715,10 +718,6 @@ def add_method(cls, selName, method, encoding):
     signature = [ctype_for_type(tp) for tp in encoding]
     assert signature[1] == objc_id  # ensure id self typecode
     assert signature[2] == SEL  # ensure SEL cmd typecode
-    if signature[0] is not None and issubclass(signature[0], (Structure, Union)):
-        # Patch struct/union return types to make them work in callbacks.
-        # See the source code of the ctypes_patch module for details.
-        ctypes_patch.make_callback_returnable(signature[0])
```
dgelessus: I think the `make_callback_returnable` call should still happen automatically, it's an internal issue with `ctypes` that the user shouldn't have to care about. That way we also don't have to decorate every structure in Rubicon. We could change `make_callback_returnable` so it does nothing if the ctype has already been patched (i. e. if it has the `_rubicon_objc_ctypes_patch_getfunc` and `..._setfunc` attributes), then it could be called multiple times on the same ctype, but would still error out if called on a ctype that has a "native" `getfunc`/`setfunc`.

freakboy3742: Fair point. It also makes it easier to optionally include (and eventually remove) the `make_callback_returnable` call. Update incoming shortly.

Issue #74, opened 2017-09-28
Remove use of Core Foundation?


dgelessus:
Should we eventually remove the use of Core Foundation for operations like creating strings and collections? As far as I can tell, there are no bootstrapping issues where Core Foundation would be needed to access Foundation APIs. For example, `NSString`s can be created from bare C data using the [`+[NSString stringWithUTF8String:]`](https://developer.apple.com/documentation/foundation/nsstring/1497379-stringwithutf8string?language=objc) or [`+[NSString stringWithCharacters:length:]`](https://developer.apple.com/documentation/foundation/nsstring/1497248-stringwithcharacters?language=objc) methods.

The advantage of using only Foundation is that we don't need to interface with an additional library. It's not necessary to use Core Foundation directly even with APIs that only accept Core Foundation types, because the corresponding Foundation types can be used instead, using toll-free bridging. Foundation is also more high-level and easy to use - you don't need to declare all the functions and manually keep track of object types, the Objective-C runtime takes care of everything.

freakboy3742:
Sounds good to me - simplification is always a good thing, and as I understand it, moving to a CoreFoundation-less implementation would mean Rubicon could be used on Objective-C implementations that aren't Apple's.

cculianu:
Ha, just don't remove some of the low-level types such aas CGPoint and CGSize and CGRect, obviously. 👍

PR #88, opened 2017-11-30
Rewrite most of the core_foundation module based on Foundation

dgelessus:
Implements #74.

This is a rewrite of most of the `core_foundation` module's functions based on Foundation classes, resulting in more readable code with no manual function declarations needed. The only things left in the `core_foundation` module are the basic types/constants/functions used by the ~~`async`~~ `eventloop` module.

The conversion functions have also been restructured a bit. `from_value` and `to_value` have been renamed to `ns_from_py` and `py_from_ns`, because the old names weren't very clear on what direction the conversion went. The `at` function, which previously only produced strings, is now exactly equivalent to `ns_from_py`. This allows things like `at(42)` or `at([1, 2, 3])`, similar to the `@` prefix/operator in Objective-C. The type-specific conversion functions (`to_str`, `to_number`, etc.) have been removed, because `py_from_ns` can be used in place of all of them.

I don't know if the old functions from `core_foundation` are used much externally. I did a quick search through the Toga repo and couldn't find anything except `at` being used. (`at` has only been moved internally, so that shouldn't be an issue.) Are there any other users of Rubicon that I should check for this sort of thing? I know Rubicon isn't officially stable yet, but I still prefer not to break other people's code.

As usual, this PR includes some general cleanup and fixes, so the individual commit diffs may be easier to read than the full diff.

### DIFF
--- a/rubicon/objc/__init__.py
+++ b/rubicon/objc/__init__.py
@@ -1,14 +1,10 @@
 __version__ = '0.2.9'
 
-from .core_foundation import (  # noqa: F401
-    NSArray, NSDictionary, NSMutableArray, NSMutableDictionary, at, to_list,
-    to_number, to_set, to_str, to_value,
-)
 from .runtime import (  # noqa: F401
-    IMP, SEL, Block, Class, Ivar, Method, NSObject, NSObjectProtocol, ObjCBlock, ObjCClass,
-    ObjCInstance, ObjCMetaClass, ObjCProtocol, objc_classmethod, objc_const, objc_id,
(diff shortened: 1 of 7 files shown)


dgelessus:
Thanks :) I appreciate your reviews, it's always good to have someone else look for logic errors and such, and to get feedback on the API.

Yeah, the naming of `string_at` is unfortunate. I think it's a leftover from Python 2, where `str` was 8-bit, so it was the natural counterpart to C `char` strings. The Unicode version is `wstring_at`, which converts C `wchar_t` strings.

Issue #13, opened 2016-10-31
Crash under GCC's libobjc due to lack of class_setSuperclass


ElSaico:
When trying to run Podium on Linux for further development I faced this issue when it tried to use this library:

```
Traceback (most recent call last):
  File "/home/bruno/anaconda3/bin/podium", line 11, in <module>
    load_entry_point('podium', 'console_scripts', 'podium')()
  File "/home/bruno/anaconda3/lib/python3.4/site-packages/pkg_resources/__init__.py", line 565, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/bruno/anaconda3/lib/python3.4/site-packages/pkg_resources/__init__.py", line 2589, in load_entry_point
    return ep.load()
  File "/home/bruno/anaconda3/lib/python3.4/site-packages/pkg_resources/__init__.py", line 2249, in load
    return self.resolve()
  File "/home/bruno/anaconda3/lib/python3.4/site-packages/pkg_resources/__init__.py", line 2255, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/bruno/PycharmProjects/podium/podium/__main__.py", line 1, in <module>
    from .app import Podium
  File "/home/bruno/PycharmProjects/podium/podium/app.py", line 5, in <module>
    from podium.deck import *
  File "/home/bruno/PycharmProjects/podium/podium/deck.py", line 3, in <module>
    from rubicon.objc import ObjCClass, objc_classmethod, objc_method
  File "/home/bruno/anaconda3/lib/python3.4/site-packages/rubicon/objc/__init__.py", line 3, in <module>
    from .objc import (
  File "/home/bruno/anaconda3/lib/python3.4/site-packages/rubicon/objc/objc.py", line 144, in <module>
    objc.class_setSuperclass.restype = c_void_p
  File "/home/bruno/anaconda3/lib/python3.4/ctypes/__init__.py", line 364, in __getattr__
    func = self.__getitem__(name)
  File "/home/bruno/anaconda3/lib/python3.4/ctypes/__init__.py", line 369, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
AttributeError: /usr/lib/x86_64-linux-gnu/libobjc.so.4: undefined symbol: class_setSuperclass
```

After some research I found out that not only my version of `libobjc` (4.9.2-10, from Debian jessie) lacks the aforementioned call but it's not present on GCC's trunk yet to this day.
[...]

PR #91, opened 2017-12-03
Add support for asyncio subprocess calls.

freakboy3742:


tests/test_async.py
```diff
@@ -156,3 +156,40 @@ def echo_client(message, loop):
...
+        def list_dir():
+            self.proc = yield from asyncio.create_subprocess_shell(
+                'ls',
+                stdout=asyncio.subprocess.PIPE,
+            )
+
+            entries = set()
+            line = yield from self.proc.stdout.readline()
+            while line:
+                entries.add(line.decode('utf-8').strip())
```
dgelessus: It *might* be more correct to use `locale.getpreferredencoding()` as the encoding here, because the output of `ls` will be encoded as whatever the current locale specifies. On the other hand, practically everything uses UTF-8 as the default encoding nowadays, especially on Apple platforms, so it's unlikely to matter in practice.

rubicon/objc/eventloop.py
```diff
@@ -569,6 +569,32 @@ def _new_default_loop(self):
...
+        If not yet set, a SafeChildWatcher object is automatically created.
+        """
+        if self._watcher is None:
+            self._init_watcher()
+
+        return self._watcher
+
+    def set_child_watcher(self, watcher):
+        """Set the watcher for child processes."""
+        assert watcher is None or isinstance(watcher, AbstractChildWatcher)
```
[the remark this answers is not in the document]
freakboy3742: The assertion isn't really needed at all, so I've removed it.

tests/test_async.py
freakboy3742: I think you're right - this is one of those "technically correct, but not in a way that is actually ever helpful" thing. Is there any platform other than Apple where this code could run, even in theory?

rubicon/objc/eventloop.py
```diff
@@ -569,6 +569,32 @@ def _new_default_loop(self):
         loop._policy = self
         return loop
 
+    def _init_watcher(self):
+        with events._lock:
+            if self._watcher is None:  # pragma: no branch
+                self._watcher = SafeChildWatcher()
+                if isinstance(threading.current_thread(), threading._MainThread):
```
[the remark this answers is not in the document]
freakboy3742: Fair call. This implementation was cribbed from Python's own internal implementation - but AFAICT there's no reason not to just check if "current thread is main thread", rather than the type check.

tests/test_async.py
dgelessus: Yes, Linux. clang, libobjc, libclosure (blocks) and Core Foundation are open-source and work on Linux, and there is [GNUstep](http://gnustep.org/). Not exactly a common scenario, but at least one person has attempted to run Rubicon on Linux in the past (#13).

It's probably not worth worrying about though - the likelihood of someone using Rubicon on Linux with a non-UTF-8 locale is *very* small, in the Linux world UTF-8 is basically standard as far as I know. And even if someone wants to use such a setup, the only thing that would break is a unit test.

Alternatively, just don't decode the output and compare against bytes. That will work for all ASCII-based encodings.

PR #92, opened 2017-12-07
Fix various issues in the collection ObjCInstance subclasss

dgelessus:
See the commit messages for a full list of fixes. I tried to add unit tests for the bugs I fixed if possible.

I also moved the collection classes out of the `runtime` module into a new `collections` module.


rubicon/objc/__init__.py
```diff
@@ -1,5 +1,9 @@
 __version__ = '0.2.10'
 
+from . import types  # noqa: F401
+from . import runtime  # noqa: F401
+from . import collections  # noqa: F401
```
freakboy3742: What's the purpose behind these imports?

dgelessus: The `types` and `runtime` imports are not strictly necessary - I just added them for clarity. The `from`-`import`s below would have implicitly added them anyway. (When you do `from a.b import c`, and `c` is a module, `a.b.c` is imported and added as the attribute `c` of the parent module `a.b`.)

The `collections` import is necessary to make sure the collection classes are registered (with the `for_objcclass` decorators). Otherwise the module would never be imported, since the collection classes aren't meant to be used explicitly.

freakboy3742: Ah - registration makes sense. Also provides a partial explanation for why the code was originally in runtime, rather than in a standalone module. 

Ironically, pre-registration requirement aside, it seems odd to import a module that would be visible by a normal import path. Could the pre-registration requirement be met by importing the specific classes that need to be registered, and then *not* including them in an `__all__` declaration (which we've avoided having by using `NOQA`)?

dgelessus: Hm, I don't think importing the classes from `rubicon.objc.collections` separately makes much sense. There's nothing useful you can do with the classes themselves, so there's no reason to add them to the main `rubicon.objc` namespace. IMHO it also doesn't make it any clearer that the import is just to have the classes registered.

We could add an `__all__`, but then we would have to list each imported name there again. Alternatively we could `from`-`import *` each submodule, and then combine all of their `__all__`s. We'd need to adjust the submodule `__all__`s a bit, because at the moment I think there are a few things listed there that are only meant to be used internally.

freakboy3742: Agreed - those options (and mine) aren't much better.  

One last option - we could separate the code out into `collections` (which I absolutely agree is a refactoring worth doing - but use an import at the *end* of `runtime` to trigger the registration process. That would give the improved logical separation of code, but without polluting the `rubicon.objc` namespace. However, it does break the "all imports at the top" pattern.

Maybe just having the "from . import collections" is the best approach, after all - as long as we document why it's being done so we don't accidentally clean it up later "because it's clearly not needed"...

dgelessus:
Just did a bit of psychic debugging, since my Mac has Xcode 8.2.1 and the issue *of course* only appears on Xcode 8.3 or higher.

Here's what's happening. I'm trying to call the method [`-[NSMutableArray replaceObjectsInRange:withObjectsFromArray:]`](https://developer.apple.com/documentation/foundation/nsmutablearray/1416902-replaceobjectsinrange?language=objc). However, [the error](https://travis-ci.org/pybee/rubicon-objc/jobs/313424140#L1139) says that the only method whose name starts with `replaceObjectsInRange` is `replaceObjectsInRange:withObjects:count:`. There's no such method listed in [Apple's docs for `NSMutableArray`](https://developer.apple.com/documentation/foundation/nsmutablearray?language=objc#1654082), so I assume it's a private method added by the implementation's `NSMutableArray` subclass.

Based on this info, I found a bug in the kwarg method call implementation when inheritance was involved. For example, if a method `initWithFoo:bar:` existed in the superclass,
and the subclass had a method `initWithFoo:spam:`, only the latter
could be called on the subclass.

dgelessus:
OK, it seems that did the trick. I also added some comments to the `__init__.py` to explain what the submodule imports are for, and updated the Travis build matrix to use Xcode 9.2. Assuming the next bunch of Travis builds runs through, that's everything done I think.

PR #95, opened 2018-01-03
Fix properties behind an accessor method. (?)

Longhanks:
I'm really puzzled. I'm using rubicon in an iOS project ([Ekans](https://github.com/Longhanks/Ekans)), and recently updated rubicon to the newest version.

Since then, the iOS application failed to start: iOS is asking my app delegate for the `window` property and assigns an UIWindow instance, then calls `[window setRootViewController:]` on said window, assigning the root view controller. This fails, because iOS was previously unable to correctly assign the UIWindow instance to the `window` property of my app delegate.

Digging into rubicon, the `_objc_getter` Python method calls `result = ns_from_py(getter(py_self))`, which raises a TypeError, because `getter(py_self)` returns an ObjCBoundMethod instead of the ObjCInstance object wrapping the UIWindow instance. Calling the returned ObjCBoundMethod, which is a getter to the property, returns the UIWindow, which is what this PR implements. But I am sure this is more of a workaround than desired behavior.

I don't understand enough of the ObjC runtime to be able to understand what's happening here. If someone has a better understanding, I'd be glad for instructions on how to fix this.

(Also, and this may or may not be related to the issue, `cache_property_methods` asks the ObjC runtime if the class responds to a given property, but in the final check, `if responds or (accessor and mutator) or forced:`, the `if responds` part is always `False`: responds is an `objc_property_t`, which is not truthy. Turning the part into `if responds is not None or ...` fixes the true check, but raises many more bugs, such as methods responding to `libobjc.class_getProperty`, making them ObjCInstances instead of ObjCBoundMethods... but again, I don't know enough about the runtime why this is the behavior, methods shouldn't be properties, right?)

Furthermore, I was unable to reproduce the issue on my Mac via a test - It's only happening on the iOS device when iOS calls `[UIWindow setRootViewController:]` and the window property is somehow not correctly working.

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -1107,21 +1107,24 @@ class objc_property(object):
 
     def register(self, cls, attr):
         def getter(_self) -> ObjCInstance:
-            return getattr(_self, '_' + attr, None)
+            getter_result = getattr(_self, '_' + attr, None)
+            if isinstance(getter_result, ObjCBoundMethod):
+                getter_result = getter_result()
+            return getter_result
 
         def setter(_self, new):
             if not hasattr(_self, '_' + attr):
                 setattr(_self, '_' + attr, None)
-            if getattr(_self, '_' + attr) is None:
+            if getter(_self) is None:
                 setattr(_self, '_' + attr, new)
                 if new is not None:
                     new.retain()
             else:
-                if not getattr(_self, '_' + attr).isEqualTo_(new):
-                    getattr(_self, '_' + attr).release()
+                if not getter(_self).isEqualTo_(new):
+                    getter(_self).release()
                     setattr(_self, '_' + attr, new)
                     if new is not None:
-                        getattr(_self, '_' + attr).retain()
+                        getter(_self).retain()
 
         getter_encoding = encoding_from_annotation(getter)
         setter_encoding = encoding_from_annotation(setter)


freakboy3742:
Thanks for the report (and potential patch). A couple of questions to help us narrow down what is going on here:

* When you say "latest" Rubicon - can you clarify exactly which version that is?

* I understand that you can't reproduce this into a unit test - What's the simplest example (on iOS, I presume) that demonstrates the problem? If you can start with the [BeeWare Python-iOS-template](https://github.com/pybee/Python-iOS-template) that would be even more helpful.

* What iOS versions are you seeing this on? Is it an iOS11-specific issue?

(Unrelated, but a matter of interest - what is Ekans? I'm personally excited to see anyone using Rubicon et al in the wild!)

dgelessus:
Here's what I *think* is going on - though I don't have a proper iOS dev setup, so I can't verify if this is correct.

The value of an `objc_property` is stored as a *Python attribute* on the `ObjCInstance` object, and the attribute name is the property name with an underscore prepended. The attribute getting and setting is naturally done using `getattr` and `setattr`.

The issue here is that `UIResponder` has a private method called `_window`. This means that when the getter of `window` tries to get the value of the `_window` attribute, the custom `__getattr__` of `ObjCInstance` kicks in and returns a method object for the `_window` method.

But I think even if the private `_window` method didn't exist, you would get an `AttributeError` here, for accessing the property before setting it. If the property was set before, it would exist in the instance `__dict__`, which would take precedence over `__getattr__`. The correct fix would be to add an `init` method to your delegate that sets the `window` property to `None`, so that getting it never tries to access an unset attribute.

However I think storing `objc_property` values as Python attributes isn't a good idea in the first place. How an `objc_property` is named shouldn't matter, it should never clash with a Python attribute with a different name. IMHO it would be better to create an `objc_ivar` with the same name and use that to hold the property value. `objc_ivar` values are only accessible from Python via the `(get|set)_instance_variable` functions, so there are no possible name clashes with Python attributes.

dgelessus:
@Longhanks An `objc_property_t` is falsey iff if it is a `NULL` pointer:

```python
>>> bool(objc.runtime.objc_property_t(0))
False
>>> bool(objc.runtime.objc_property_t(8))
True
```

Changing the check to `responds is not None` won't work - `responds` is *never* `None`, only `responds.value` can be `None`. (For some reason, `ctypes` decided that the `value` of a `NULL` `c_void_p` is `None` rather than `0`, even though all other `c_void_p` `value`s are integers.)

Longhanks:
Hi again, thank you for your feedback and your input! Sorry for my late answer.

@freakboy3742:

- I'm talking about the latest commit on the master branch, which is [4618db6f6edd76b91ec94c58016da2b6e1002b65](https://github.com/pybee/rubicon-objc/commit/4618db6f6edd76b91ec94c58016da2b6e1002b65).
- I've created a small  bash script for creating a sample project [here](https://gist.github.com/Longhanks/16004615ec2b95657122c8db2aac7a88). Essentially, it creates an empty project via cookiecutter, downloads Python-Apple-support frameworks and rubicon, creates a bare bone AppDelegate class, a storyboard with an empty view controller (containing an empty view) and patches the project files to point to that storyboard. This project can be run in the simulator, the issue will reproduce.
- I've tested it on 11.1.2 in the Simulator and on my real iPhone X. Because the window property is required since Storyboards can be used as initial view controller providers, I assume the issue is reproducible on many earlier iOS versions, too.

Ekans is, right now, a text input and a text output view, where you can enter text which will then be eval'd by the Python interpreter, stdout is captured and shown on the output view - it's kind of a (very) barebone foundation of what could become an open source Pythonista clone :)

@dgelessus:

This seems to do the trick. Creating a new method like this:
```Python
    @objc_method
    def init(self):
        self._window = None
        return self
```
in the AppDelegate class makes the issue disappear. However, assigning `None` to `window` itself doesn't. So this seems like another workaround for the same underlying problem...

(I guess this workaround is useless then, should the PR be closed?)

dgelessus:
@Longhanks That's interesting - if you try to set `self.window` instead of `self._window`, what happens? If possible, can you post a traceback?

Also, a small nitpick regarding ObjC `init` methods - you need to call the superclass' `init` inside your own `init`, otherwise the instance might not be initialized correctly ([explanation on Stack Overflow](https://stackoverflow.com/questions/2956943/why-should-i-call-self-super-init)). The correct procedure for calling the super `init` looks like this:

```python
class MySubclass(UIResponder):
    @objc_method
    def init(self):
        # self needs to be reassigned - some inits return a different self than the one they were originally called on
        self = send_super(self, "init")
        # self may be null if the super init failed, our own init code cannot be run in that case
        if self is not None:
            # Our init code comes here
            self.window = None
        return self
```

`send_super` is like `send_message`, except that it calls the method on the superclass. (`send_super(self, "name:", arg)` is equivalent to `[self name:arg]`.)

Longhanks:
So, if my AppDelegate looks like this (self is not `None` after calling `send_super`, I've checked that):

```Python
from rubicon.objc import ObjCClass, ObjCInstance, objc_method, objc_property, send_super
from UIKit import *


class EKANSAppDelegate(UIResponder):
    window = objc_property()
    
    @objc_method
    def init(self):
        self = ObjCInstance(send_super(self, 'init'))
        if self is not None:
            self.window = None
        return self

    @objc_method
    def application_didFinishLaunchingWithOptions_(self, application, launchOptions):
        return True
```

...then this is the traceback:

```
  File "_ctypes/callbacks.c", line 234, in 'calling callback function'
  File "/var/containers/Bundle/Application/7EB439B0-8131-4F73-92B0-F0E3292D1329/Ekans.app/Library/Python.framework/Resources/lib/python3.6/site-packages/rubicon/objc/runtime.py", line 1139, in _objc_setter
    setter(py_self, ObjCInstance(name))
  File "/var/containers/Bundle/Application/7EB439B0-8131-4F73-92B0-F0E3292D1329/Ekans.app/Library/Python.framework/Resources/lib/python3.6/site-packages/rubicon/objc/runtime.py", line 1120, in setter
    if not getattr(_self, '_' + attr).isEqualTo_(new):
AttributeError: 'ObjCBoundMethod' object has no attribute 'isEqualTo_'
  File "_ctypes/callbacks.c", line 234, in 'calling callback function'
  File "/var/containers/Bundle/Application/7EB439B0-8131-4F73-92B0-F0E3292D1329/Ekans.app/Library/Python.framework/Resources/lib/python3.6/site-packages/rubicon/objc/runtime.py", line 1131, in _objc_getter
    result = ns_from_py(getter(py_self))
  File "/var/containers/Bundle/Application/7EB439B0-8131-4F73-92B0-F0E3292D1329/Ekans.app/Library/Python.framework/Resources/lib/python3.6/site-packages/rubicon/objc/runtime.py", line 1800, in ns_from_py
    .format(cls=type(pyobj))
TypeError: Don't know how to convert a rubicon.objc.runtime.ObjCBoundMethod to a Foundation object
2018-01-06 16:16:47.540612+0100 Ekans[14022:4136171] -[UIStoryboard setRootViewController:]: unrecognized selector sent to instance 0x1d4068000
2018-01-06 16:16:47.540765+0100 Ekans[14022:4136171] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[UIStoryboard setRootViewController:]: unrecognized selector sent to instance 0x1d4068000'
*** First throw call stack:
(0x184e96364 0x1840dc528 0x184ea3828 0x184e9bd10 0x184d80ecc 0x18e5ddf2c 0x18e3e6008 0x18e5dc984 0x18e8782d4 0x18eb5012c 0x18e877f38 0x18e878798 0x18effd7d0 0x18effd674 0x18ed6b2dc 0x18ef043dc 0x18ed6b18c 0x18eb4f78c 0x18e5db1ec 0x18e9f0de8 0x18755a1f0 0x187562af8 0x100b3928c 0x100b459e4 0x18758e7f8 0x18758e49c 0x18758ea38 0x184e3e97c 0x184e3e8fc 0x184e3e184 0x184e3bd5c 0x184d5be58 0x186c08f84 0x18e3db67c 0x100086610 0x18487856c)
libc++abi.dylib: terminating with uncaught exception of type NSException
(lldb) 
```

But without setting `window` to `None`, this is the result:

```Python
from rubicon.objc import ObjCClass, ObjCInstance, objc_method, objc_property, send_super
from UIKit import *


class EKANSAppDelegate(UIResponder):
    window = objc_property()

    @objc_method
    def application_didFinishLaunchingWithOptions_(self, application, launchOptions):
        return True
```

Traceback:

```
  File "_ctypes/callbacks.c", line 234, in 'calling callback function'
  File "/var/containers/Bundle/Application/0C98F510-E814-43DB-ABF5-9D6B2D6D381B/Ekans.app/Library/Python.framework/Resources/lib/python3.6/site-packages/rubicon/objc/runtime.py", line 1131, in _objc_getter
    result = ns_from_py(getter(py_self))
  File "/var/containers/Bundle/Application/0C98F510-E814-43DB-ABF5-9D6B2D6D381B/Ekans.app/Library/Python.framework/Resources/lib/python3.6/site-packages/rubicon/objc/runtime.py", line 1800, in ns_from_py
    .format(cls=type(pyobj))
TypeError: Don't know how to convert a rubicon.objc.runtime.ObjCBoundMethod to a Foundation object
2018-01-06 16:19:45.314611+0100 Ekans[14028:4137659] -[UIStoryboard setRootViewController:]: unrecognized selector sent to instance 0x1d4272a80
2018-01-06 16:19:45.314778+0100 Ekans[14028:4137659] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[UIStoryboard setRootViewController:]: unrecognized selector sent to instance 0x1d4272a80'
*** First throw call stack:
(0x184e96364 0x1840dc528 0x184ea3828 0x184e9bd10 0x184d80ecc 0x18e5ddf2c 0x18e3e6008 0x18e5dc984 0x18e8782d4 0x18eb5012c 0x18e877f38 0x18e878798 0x18effd7d0 0x18effd674 0x18ed6b2dc 0x18ef043dc 0x18ed6b18c 0x18eb4f78c 0x18e5db1ec 0x18e9f0de8 0x18755a1f0 0x187562af8 0x1014a928c 0x1014b59e4 0x18758e7f8 0x18758e49c 0x18758ea38 0x184e3e97c 0x184e3e8fc 0x184e3e184 0x184e3bd5c 0x184d5be58 0x186c08f84 0x18e3db67c 0x1009fe610 0x18487856c)
libc++abi.dylib: terminating with uncaught exception of type NSException
(lldb) 
```

dgelessus:
Ah, I see, the `objc_property` setter gets the attribute before setting it. I didn't see that at first, so I thought setting the property would work. In that case, setting `self._window = None` in `init` is probably the best solution for now, until `objc_property` is fixed.

Issue #98, opened 2018-02-13
Issue using objc_property


cculianu:
I have an class roughly like this:

```
class MyObj(NSObject):
    foo = objc_property()

    @objc_method
    def init(self):
        self = ObjCInstance(send_super(self, 'init'))
        self.foo = True
        return self

```

Later on whenever I want to write to self.foo, I get errors such as:
```

app_packages/rubicon/objc/runtime.py", line 1382, in __getattr__
    type(self).__module__, type(self).__qualname__, self.objc_class.name, name)
AttributeError: rubicon.objc.runtime.ObjCInstance __NSCFNumber has no attribute isEqualTo_
(lldb) 
```

I basically can't really overwrite any properties.  I suspect there is some bug in runtime.py.  Any thoughts?

I'm on rubicon_objc-0.2.10-py3.5

Please help!

cculianu:
I figured out why this is happening.  IsEqualTo: is only available on macOS.  I'm using rubicon on iOS. An alternate method needs to be used rather than 'IsEqualTo:'...

Rubicon is broken for iOS until this is fixed.  Perhaps I can issue a pull request to fix this?

Or care to fix it?

freakboy3742:
Please turn down the emotional hyperbole. Rubicon for iOS isn't "broken". It's "broken for your specific use case". I can point at many places where it works fine.

And pull requests are always welcome :-)

cculianu:
Sorry, I didn't mean to offend you.  I wasn't being emotional, it's just the way I talk, man. :)

Rubicon is great.  I am happy it exists and it does actually work like magic most of the time.  It's just this 1 little bug.

I'm going to submit a PR now, but I am not sure you'll be entirely happy with it.  It's worth a shot, though.  Basically isEqualTo: is missing from NSObject on iOS.  Here's the Apple developer page for this method.. notice how platforms is *only* "macOS".

https://developer.apple.com/documentation/objectivec/nsobject/1393823-isequalto?language=objc

So my workaround is to basically not call it, and instead do the assignment using a .autorelease()/.retain() pair.  It works.. but maybe you will find it breaks semantics.  We'll see.

Coming shortly...

riazrizvi:
If I may, I don't see why it is desirable to inject a MacOS function to this iOS interface, it would make the API confusing. Why not just use the available IsEqual function? Personally I don't see any need for a change.

dgelessus:
Fixed in #99.

PR #99, opened 2018-02-14
Fixes bug on iOS (and other non-macOS) where 'isEqualTo:' is missing for NSObject

cculianu:
Fixes issue #98 .

I hope you are ok with it.  It works great in my testing -- but you may be upset that it slightly slightly alters when objects get release and/or semantics.

Another approach would be to check platform -- but on ios simulator platform returns the same thing as on macOS.  

Could also check at startup if NSObject has that selector and set a global flag.. but that would be more invasive.

I went with this approach.  Let me know if it's ok.

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -1117,11 +1117,10 @@ class objc_property(object):
                 if new is not None:
                     new.retain()
             else:
-                if not getattr(_self, '_' + attr).isEqualTo_(new):
-                    getattr(_self, '_' + attr).release()
-                    setattr(_self, '_' + attr, new)
-                    if new is not None:
-                        getattr(_self, '_' + attr).retain()
+                getattr(_self, '_' + attr).autorelease()
+                setattr(_self, '_' + attr, new)
+                if new is not None:
+                    getattr(_self, '_' + attr).retain()
 
         getter_encoding = encoding_from_annotation(getter)
         setter_encoding = encoding_from_annotation(setter)


dgelessus:
Technically there's no need to remove the equality check completely - the [`NSObject`](https://developer.apple.com/documentation/objectivec/1418956-nsobject?language=objc) protocol includes an [`isEqual:`](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418795-isequal?language=objc) method, which is supported on all platforms. The [`isEqualTo:`](https://developer.apple.com/documentation/objectivec/nsobject/1393823-isequalto?language=objc) method comes from the [NSComparisonMethods](https://developer.apple.com/documentation/foundation/scripting_support/nscomparisonmethods?language=objc) informal protocol, which is related to [scripting support](https://developer.apple.com/documentation/foundation/scripting_support?language=objc) and as such doesn't exist on platforms other than macOS. (See also [this StackOverflow question](https://stackoverflow.com/q/7096691).)

However I would be fine with removing the equality check no matter what. As far as I can tell, property setters synthesized by a real Objective-C compiler don't check for equality, at least not for `retain` properties. On my system, [this test program](https://gist.github.com/dgelessus/d943f2d1f477b046fd32b89e551ca23f) produces the following output:

```sh
$ xcodebuild -version
Xcode 8.2.1
Build version 8C1002
$ clang --version
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
$ clang -fobjc-link-runtime -o DGPropertyEqualTest DGPropertyEqualTest.m
$ ./DGPropertyEqualTest
2018-02-14 12:31:19.762 DGPropertyEqualTest[16842:1085715] thing1 thing2
2018-02-14 12:31:19.763 DGPropertyEqualTest[16842:1085715] thing1
2018-02-14 12:31:19.763 DGPropertyEqualTest[16842:1085715] thing2
```

cculianu:
Wow, great analysis.  I actually didn't think to do the rigorous test you did, and I'm really glad you did it.  I learned something more concrete which I felt intuitively would be true anyway -- about whether synthesized properties care about equality. Intuitively I sort of felt they don't actually check for it (since the default implementation of isEqual just checks pointer value, if I recall correctly), which was my motivation for doing away with the test.  

Also checking for equality as a rule about whether you want to overwrite properties in obj-c, which is a language that has references, is problematic anyway.  Maybe you really want to assign object X as a property, overwriting object Y (which at the time of the assignment is 'equal' in terms of the value it holds) -- but you know what you're doing.  You did it because you want X to be shared across objects, for example (or, rather, references to X).  So it should be allowed.

Anyway -- I digress. 

I'm glad you're ok with removing the test, in light of the potential problems it causes on non-macOS, and also the limited usefulness of the default implementation of isEqual (I belive on NSObject default just tests for pointer equality).  And also there's the potential semantic/behavior difference of the code when the test is in place.

Thanks so much for doing the check on whether the synthesized properties actually check for equality.  I hadn't thought to do that!

👍

cculianu:
Sorry for comment spam.  Just to 100% convince myself I also ran my own test (dunno where the DGPPropertyEqualTest.m you got is from -- where is it from?).

The below program (ARC disabled):

```
//
//  main.m
//  TstStuff
//
//  Created by calin on 2/14/18.
//  Copyright © 2018 c3-soft.com. All rights reserved.
//

#import <Foundation/Foundation.h>

@interface MyObj : NSObject {
    NSString *val;
}
@property (retain) NSString *val;
@end

@implementation MyObj
@synthesize val;

    - (id) init {
        self = [super init];
        self.val = @"";
        NSLog(@"MyObj init");
        return self;
}

    - (void) dealloc {
        NSLog(@"MyObj dealloc (%@)", val);
        self.val = nil;
        [super dealloc];
    }

    - (BOOL) isEqual:(id)other {
        NSLog(@"MyObj isEqual called for (%@,%@)!", self.val, ((MyObj *)other).val);
        return [super isEqual:other];
    }
    - (BOOL) isEqualTo:(id)other {
        NSLog(@"MyObj isEqualTo called for (%@,%@)!", self.val, ((MyObj *)other).val);
        return [super isEqualTo:other];
    }
@end

@interface Tst : NSObject {
    MyObj *o;
}
    @property (retain) MyObj *o;
@end
@implementation Tst
    @synthesize o;
@end

int main(int argc, const char * argv[]) {
    @autoreleasepool {
        // insert code here...
        Tst *t = [Tst new];
        MyObj *a = [MyObj new], *b = [MyObj new];
        a.val = @"A";
        b.val = @"B";
        t.o = a;
        [a release];
        NSLog(@"t.o.val=%@", t.o.val);
        t.o = b;
        [b release];
        NSLog(@"t.o.val=%@", t.o.val);
        t.o = nil;
    }
    return 0;
}

```

Produces this output:

```

2018-02-14 15:04:10.400 TstStuff[50616:10338505] MyObj init
2018-02-14 15:04:10.401 TstStuff[50616:10338505] MyObj init
2018-02-14 15:04:10.401 TstStuff[50616:10338505] t.o.val=A
2018-02-14 15:04:10.401 TstStuff[50616:10338505] t.o.val=B
2018-02-14 15:04:10.401 TstStuff[50616:10338505] MyObj dealloc (B)
2018-02-14 15:04:10.401 TstStuff[50616:10338505] MyObj dealloc (A)
Program ended with exit code: 0
```

So yeah.. no isEqual (or isEqualTo:) is ever called for synthesized retain properties.

I just did this to totally convince myself.  This means it's safe to remove the test altogether.

Interestingly I think obj-c also does an 'autorelease' when clearing out retained properties (hence the deallocs all happening together at the end).

dgelessus:
`DGPropertyEqualTest.m` is from nowhere, I wrote it when I made my comment so I could test how Clang behaves here. :)

I'm not entirely sure if the synthesized setter does a release or an autorelease. If you compile the program to LLVM code (`clang -fobjc-link-runtime -S -emit-llvm -o DGPropertyEqualTest.llvm DGPropertyEqualTest.m`) and look at `-[DGPropertyHolder setThing:]`, you can see that it calls a function `objc_setProperty_atomic`. The source code for that function can be found in [runtime/objc-accessors.mm](https://opensource.apple.com/source/objc4/objc4-723/runtime/objc-accessors.mm.auto.html) in the Objective-C runtime. The implementation calls `objc_release` and not `objc_autorelease`, so I'm not sure why the deallocation is delayed. It might be some sort of optimization that I missed, or that's performed by the compiler.

Anyway, removing the equality check should be fine, so I'll go ahead and merge this.

cculianu:
You're right -- setting the property does the retain/release thing only.

On further inspection I figured out what's going on!

The observed behavior can be explained, still, because if you look at the getter function, objc_getProperty() in that very same file -- it looks like in the cases of atomic properties (the default if unspecified) there is a retain-then-autorelease happening. 

```
id objc_getProperty(id self, SEL _cmd, ptrdiff_t offset, BOOL atomic) {
    if (offset == 0) {
        return object_getClass(self);
    }

    // Retain release world
    id *slot = (id*) ((char*)self + offset);
    if (!atomic) return *slot;
        
    // Atomic retain release world
    spinlock_t& slotlock = PropertyLocks[slot];
    slotlock.lock();
    id value = objc_retain(*slot);
    slotlock.unlock();
    
    // for performance, we (safely) issue the autorelease OUTSIDE of the spinlock.
    return objc_autoreleaseReturnValue(value);
}
```

Sure enough, if I make all the properties in my code 'nonatomic', the behavior behaves no as before but as expected from superficial reading of the code -- the releases happen immediately, and not at the end of the innermost autorelease pool scope.

Switch back to 'atomic' (the default), and it does autorelease.  

This is because the properties are being 'read' for NSLog() output.  Commenting out the output also has the effect of no longer putting the object in the autorelease pool as well...

So moral of the story: objc can sometimes autorelease, rather than release.. for atomic properties. 
 

This also means my little change (we are dealing with atomic properties anyway), seems to be ok, even if it does do an autorelease, since autorelease would happen anyway regardless.

Thanks a lot for the discussion!  I learned something new!  And I liked that you found this objc implementation code.  I didn't think to read how it's all implemented.  It's nice code to read and sometimes illustrates things more clearly than any docs you could ever read!

Much appreciated and thanks for looking at this so quickly and expertly!  This little bug was tripping up my iOS project!  Now I can get back to coding.. 👍

freakboy3742:
@cculianu I see you've now met @dgelessus 😄 He's a unsung wonder of the world, and the reason that Rubicon went from being a passable Objective C bridge into a truly compelling one. I swear he's forgotten more about the Objective C runtime than I have ever known.

cculianu:
Ah, yeah this was a good experience overall.  Including this PR stuff. 

Rubicon is very nice indeed and I am so glad to be using it.  It's elegant and simple and powerful. 👍

PR #100, opened 2018-02-17
Documentation updates

danyeaw:
Rewrap line length of documentation and add additional documentation about NSSize and preventing type conversion that I learned from implementing the Canvas widget.

### DIFF
--- a/docs/how-to/type-mapping.rst
+++ b/docs/how-to/type-mapping.rst
@@ -82,5 +82,30 @@ means you can treat the return value as if it were a dict - iterating over
 keys, values or items, retrieving objects by key, and so on.
 
 `NSPoint`, `NSSize`, and `NSRect`
----------------------------------
+ -------------------------------
 
+On instances of an Objective C structure, each field is exposed as a Python
+attribute. For example, if you create an instance of an `NSSize` object you can
+access its width and height by calling `NSSize.width`.
+
+When you need to pass an Objective C structure to an Objective C method,
+you can pass a tuple instead. For example, if you pass (10.0, 5.1) where a
+`NSSize` is expected, it will be converted automatically in the appropriate
+width, height for the structure.
+
+Prevent type conversion
+-----------------------
+
+For some use cases you may not actually want to do an automatic type conversion.
+For example if you need to make use of an actual `NSString` object in a Python
+program, you need the ability to prevent automatic conversion in to `str`.
+
+To prevent type conversion, pass `convert_result=False` as a parameter. An
+example of this in action would be to create a text string in Python::
+
+    text_string = ObjCInstance(
+        ObjCInstance(NSString.alloc(convert_result=False)).initWithString_(text, convert_result=False)
+    )
+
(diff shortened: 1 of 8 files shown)


docs/how-to/type-mapping.rst
```diff
@@ -82,5 +82,30 @@ means you can treat the return value as if it were a dict - iterating over
 keys, values or items, retrieving objects by key, and so on.
 
 `NSPoint`, `NSSize`, and `NSRect`
----------------------------------
+ -------------------------------
```
dgelessus: Is this valid RST? I think the underline has to be at least as long as the title, or you'll get a warning or error.

docs/how-to/type-mapping.rst
```diff
@@ -82,5 +82,30 @@ means you can treat the return value as if it were a dict - iterating over
...
+For some use cases you may not actually want to do an automatic type conversion.
+For example if you need to make use of an actual `NSString` object in a Python
+program, you need the ability to prevent automatic conversion in to `str`.
+
+To prevent type conversion, pass `convert_result=False` as a parameter. An
+example of this in action would be to create a text string in Python::
+
+    text_string = ObjCInstance(
+        ObjCInstance(NSString.alloc(convert_result=False)).initWithString_(text, convert_result=False)
+    )
```
dgelessus: Should this be a `.. code-block:: python`?

docs/how-to/type-mapping.rst
```diff
@@ -82,5 +82,30 @@ means you can treat the return value as if it were a dict - iterating over
...
+
+This is admittedly really ugly, because this gets unwrapped twice and there
+is no ObjCStringInstance yet like there is for dictionaries above.
```
dgelessus: While I completely agree with what you're saying here, I don't think this belongs in the documentation. Maybe instead it should be a note saying that this use case will be better supported in the future.

dgelessus:
I'm not sure why GitHub is showing two of my comments as "outdated". Might be because I was looking at a single comment diff instead of the whole PR diff. In any case, the comments do appply to the current version.

docs/how-to/type-mapping.rst
freakboy3742: The `.. code-block` part is optional; it's a hint to the markup to sell it what language highlighting scheme to use, but as long as the code is indented, and the previous paragraph ends with `::`, it will be rendered as code.

dgelessus: Ah, I wasn't sure if `::` blocks would always be rendered as code, good to know. But since it's Python code, it should be marked as such, I think.

danyeaw: reST defaults to Python code, but I agree to be consistent with the other parts of the documentation I'll spell it out.

docs/how-to/type-mapping.rst
danyeaw: Yes, I agree I could have worded this better.

PR #103, opened 2018-02-27
Fix and extend implementation of ivars and properties

dgelessus:
This changes the implementation of `objc_property` so it stores the property value in an ivar instead of a Python attribute. This should fix the issues described in #95.

While at it, I also fixed the implementation of the ivar get/set functions so they work correctly for non-object types, and added support for properties with types other than `ObjCInstance`.

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -7,13 +7,13 @@ from ctypes import (
     CDLL, CFUNCTYPE, POINTER, ArgumentError, Array, Structure, Union,
     addressof, alignment, byref, c_bool, c_char_p, c_double, c_float, c_int,
     c_int32, c_int64, c_longdouble, c_size_t, c_uint, c_uint8, c_ulong,
-    c_void_p, cast, sizeof, string_at, util,
+    c_void_p, cast, memmove, sizeof, string_at, util,
 )
 
 from . import ctypes_patch
 from .types import (
     __arm__, __i386__, __x86_64__, compound_value_for_sequence,
-    ctype_for_type, ctypes_for_method_encoding, encoding_for_ctype,
+    ctype_for_encoding, ctype_for_type, ctypes_for_method_encoding, encoding_for_ctype,
     register_ctype_for_type, with_encoding, with_preferred_encoding,
 )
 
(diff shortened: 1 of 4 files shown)


rubicon/objc/runtime.py
```diff
@@ -186,6 +186,13 @@ class objc_property_t(c_void_p):
     pass
 
 
+class objc_property_attribute_t(Structure):
```
phildini: I don't understand why this was moved?

dgelessus: e344fa50b878f766336e5c5780897512212268ea added the declaration for `class_addProperty`, whose signature includes an `objc_property_attribute_t`. `class_addProperty` is declared together with all other `class_*` functions, above the previous location of `objc_property_attribute_t`, so I had to move the type up so the new function declaration could use it.

tests/test_core.py
```diff
@@ -844,16 +836,41 @@ def test_no_duplicate_protocols(self):
...
+        class Ivars(NSObject):
+            object = objc_ivar(objc_id)
+            int = objc_ivar(c_int)
+            rect = objc_ivar(NSRect)
+
+        ivars = Ivars.alloc().init()
+
+        set_ivar(ivars, 'object', at('foo').ptr)
+        set_ivar(ivars, 'int', c_int(12345))
+        set_ivar(ivars, 'rect', NSMakeRect(12, 34, 56, 78))
```
freakboy3742: Is this a deliberately explicit test of`set_ivar`, rather than using "Pythonic" assignment to `ivars.object` etc? I *think* it is; but regardless, for the sake of clarity, a comment explaining what exactly is being tested would be helpful.

dgelessus: Python attribute syntax doesn't work with ivars, only with properties, which are tested separately. The only way to get and set ivars is with the `get_ivar` and `set_ivar` functions. (This behavior is not new by the way, attribute syntax has never worked with ivars before, as far as I know.)

freakboy3742: ACK - makes sense. At some point in the future, we should probably document when you should use an `objc_property`, when you should use an `objc_ivar`, and when you should use a Python property.

Issue #102, opened 2018-02-23
Blocks wrapping Python callables should keep the callable alive


cculianu:
I'm getting bugginess with objc blocks passed as parameters.  I noticed rubicon has support for them automagically as such:
```

def blockTes1t() -> None:
    # ....

# .... and then maybe do:
vc.dismissViewControllerAnimated_completion_(True,blockTest1)

```

According to #54 , this should work.  However I get a runtime error half the time:

`SystemError: null argument to internal routine`

.. and the other half of the time I get a random crash and a debugger-level stack trace -- indicating to me something is not quite right in the Python/ObjC glue code layer. :)

No matter how hard I try I can't consistently get this to work at all.  Am I doing something wrong? Does the Block code for rubicon-objc need to be updated?  I'm willing to go in and update it but I just want to make sure it's not ME doing something wrong (some guidance is appreciated!)

Also, definitely a feature is missing: I can't call methods that expect Blocks as parameters but accept 'nil' blocks because the type 'None' for blocks should be accepted, but the current Block code doesn't accept it (some methods in ObjC do indeed accept nil blocks).   If I pass 'None' for a block parameter, I get this error:

```

    arg = Block(arg).block
  File "/Users/calin/Library/Developer/CoreSimulator/Devices/80F9328C-18C7-43C1-9E93-F4CB128AD3CA/data/Containers/Bundle/Application/27B85C08-FEFC-45C5-9D30-E192F20DA04C/Electron-Cash.app/Library/Application Support/com.c3-soft.ElectronCash/app_packages/rubicon/objc/runtime.py", line 2299, in __init__
    raise TypeError('Blocks must be callable')
TypeError: Blocks must be callable
```

cculianu:
UPDATE: I think I figured out why I was getting crashes!  Duh! With Asynch block code, sometimes Python cleans up temporaries by the time your callable gets called  - the Block type needs to be stored somewhere.  Then I don't get crashes.

Still.. it would be nice to support 'None' as a block param.  That definitely doesn't work quite right.

Anyway -- it would also be nice to have the Block wrapper have the same semantics as obj-c -- namely keep the Block wrapper around until it's released.  Is that possible?  Perhaps some obj-c level observer on the obj-c level Block object type?

dgelessus:
Right, that's the usual problem with passing local variables to C functions - if you're not careful, the Python object can get deleted while the C code still has a pointer to it.

[...]

PR #105, opened 2018-03-02
Properly implement block copy and dispose helpers to allow Python blo…

cculianu:
…ck objects to have the same semantics as objc.

This means you can do Block(somePyFun) and it will stay around if obj-c side retained it (as in an asynch callback).

Previously this lead to a crash.

Also added a fixup so you can pass 'None' to methods expecting a block parameter -- as this is supported by many of the Foundation/iOS/etc APIs.

The only thing I am unsure of with this PR is whether on the copy, we need to explicitly copy back in the descriptor, etc structs.  I am fuzzy on C/Python ctypes,etc semantics so it's not that easy for me to do it, and it didn't seem necessary, so I left it out.

Any comments or suggestions are appreciated.

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -817,7 +817,11 @@ class ObjCMethod(object):
                     elif isinstance(arg, Block):
                         arg = arg.block
                     else:
-                        arg = Block(arg).block
+                        # This usage below will leverage the ctypes _as_parameter_
+                        # mechanism.
+                        # Note: We need to keep te Block instance around at least until
+                        # the objc method is called else bad things may happen!
+                        arg = Block(arg)
                 elif issubclass(argtype, objc_id):
                     # Convert Python objects to Foundation objects
                     arg = ns_from_py(arg)
@@ -2126,6 +2130,10 @@ class Block:
         )
         self.literal.descriptor = cast(byref(self.descriptor), c_void_p)
         self.block = cast(byref(self.literal), objc_block)
+
+    @property
+    def _as_parameter_(self):
+        return self.block
 
     def wrapper(self, instance, *args):
         return self.func(*args)
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -2128,6 +2128,7 @@ class Block:
 
     def wrapper(self, instance, *args):
         return self.func(*args)
+
     def disposeHelper(self, dst):
         Block._keep_alive_blocks_.pop(self.block.value, None)
 
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -1957,13 +1957,16 @@ def objc_const(dll, name):
 
     return ObjCInstance(objc_id.in_dll(dll, name))
 
+_cfunc_type_block_invoke  = CFUNCTYPE(c_void_p, c_void_p)
+_cfunc_type_block_dispose = CFUNCTYPE(c_void_p, c_void_p)
+_cfunc_type_block_copy    = CFUNCTYPE(c_void_p, c_void_p, c_void_p)
 
 class ObjCBlockStruct(Structure):
     _fields_ = [
         ('isa', c_void_p),
         ('flags', c_int),
         ('reserved', c_int),
-        ('invoke', CFUNCTYPE(c_void_p, c_void_p)),
+        ('invoke', _cfunc_type_block_invoke),
         ('descriptor', c_void_p),
     ]
 
@@ -1972,8 +1975,8 @@ class BlockDescriptor(Structure):
     _fields_ = [
         ('reserved', c_ulong),
         ('size', c_ulong),
-        ('copy_helper', c_void_p),
-        ('dispose_helper', c_void_p),
+        ('copy_helper', _cfunc_type_block_copy),
+        ('dispose_helper', _cfunc_type_block_dispose),
         ('signature', c_char_p),
     ]
 
@@ -1983,7 +1986,7 @@ class BlockLiteral(Structure):
         ('isa', c_void_p),
         ('flags', c_int),
         ('reserved', c_int),
-        ('invoke', c_void_p),
+        ('invoke', c_void_p), # NB: this must be c_void_p due to polymorphic nature
         ('descriptor', c_void_p)
     ]
 
@@ -1995,8 +1998,8 @@ def create_block_descriptor_struct(has_helpers, has_signature):
     ]
     if has_helpers:
         descriptor_fields.extend([
-            ('copy_helper', CFUNCTYPE(c_void_p, c_void_p, c_void_p)),
-            ('dispose_helper', CFUNCTYPE(c_void_p, c_void_p)),
+            ('copy_helper', _cfunc_type_block_copy),
+            ('dispose_helper', _cfunc_type_block_dispose),
         ])
     if has_signature:
         descriptor_fields.extend([
@@ -2107,18 +2110,16 @@ class Block:
         self.literal.isa = addressof(_NSConcreteGlobalBlock)
         self.literal.flags = BlockConsts.HAS_STRET | BlockConsts.HAS_SIGNATURE | BlockConsts.HAS_COPY_DISPOSE
         self.literal.reserved = 0
-        self.cfunc = self.cfunc_type(self.wrapper)
-        self.literal.invoke = cast(self.cfunc, c_void_p)
+        self.cfunc_wrapper = self.cfunc_type(self.wrapper)
+        self.literal.invoke = cast(self.cfunc_wrapper, c_void_p)
         self.descriptor = BlockDescriptor()
         self.descriptor.reserved = 0
         self.descriptor.size = sizeof(BlockLiteral)
 
-        self.copy_helper_type = CFUNCTYPE(c_void_p, c_void_p, c_void_p)
-        self.dispose_helper_type = CFUNCTYPE(c_void_p, c_void_p)
-        self.copy_helper = self.copy_helper_type(self.copyHelper)
-        self.dispose_helper = self.dispose_helper_type(self.disposeHelper)
-        self.descriptor.copy_helper = cast(self.copy_helper, c_void_p)
-        self.descriptor.dispose_helper = cast(self.dispose_helper, c_void_p)
+        self.cfunc_copy_helper = _cfunc_type_block_copy(self.copy_helper)
+        self.cfunc_dispose_helper = _cfunc_type_block_dispose(self.dispose_helper)
+        self.descriptor.copy_helper = self.cfunc_copy_helper
+        self.descriptor.dispose_helper = self.cfunc_dispose_helper
 
         self.descriptor.signature = encoding_for_ctype(restype) + b'@?' + b''.join(
             encoding_for_ctype(arg) for arg in signature
@@ -2129,10 +2130,16 @@ class Block:
     def wrapper(self, instance, *args):
         return self.func(*args)
 
-    def disposeHelper(self, dst):
-        Block._keep_alive_blocks_.pop(self.block.value, None)
+    def dispose_helper(self, dst):
+        Block._keep_alive_blocks_.pop(dst, None)
 
-    def copyHelper(self, dst, src):
-        Block._keep_alive_blocks_.pop(self.block.value, None)
-        self.block = objc_block(dst)
-        Block._keep_alive_blocks_[self.block.value] = self
+    def copy_helper(self, dst, src):
+        # Update our keepalive table because objc just informed us that it
+        # took ownership of a block/copied a block we are concerned with.
+        # Note that sometime later we can expect calls to dispose_helper
+        # for each of the 'dst' blocks objc told us about, but until then we
+        # need to make sure the python code they reference stays in memory,
+        # so basically put self in a class variable dictionary so it is
+        # guaranteed to stay around until dispose_helper tells us they are all
+        # gone.
+        Block._keep_alive_blocks_[dst] = self
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -2065,7 +2065,7 @@ class ObjCBlockInstance(ObjCInstance):
         return self.block(*args)
 
 
-_NSConcreteGlobalBlock = (c_void_p * 32).in_dll(libc, "_NSConcreteGlobalBlock")
+_NSConcreteStackBlock = (c_void_p * 32).in_dll(libc, "_NSConcreteStackBlock")
 
 
 NOTHING = object()
@@ -2107,7 +2107,7 @@ class Block:
         self.cfunc_type = CFUNCTYPE(restype, c_void_p, *signature)
 
         self.literal = BlockLiteral()
-        self.literal.isa = addressof(_NSConcreteGlobalBlock)
+        self.literal.isa = addressof(_NSConcreteStackBlock)
         self.literal.flags = BlockConsts.HAS_STRET | BlockConsts.HAS_SIGNATURE | BlockConsts.HAS_COPY_DISPOSE
         self.literal.reserved = 0
         self.cfunc_wrapper = self.cfunc_type(self.wrapper)
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -2129,10 +2129,7 @@ class Block:
         )
         self.literal.descriptor = cast(byref(self.descriptor), c_void_p)
         self.block = cast(byref(self.literal), objc_block)
-
-    @property
-    def _as_parameter_(self):
-        return self.block
+        self._as_parameter_ = self.block
 
     def wrapper(self, instance, *args):
         return self.func(*args)
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -811,7 +811,10 @@ class ObjCMethod(object):
                     arg = arg.value
 
                 if issubclass(argtype, objc_block):
-                    if isinstance(arg, Block):
+                    if arg is None:
+                        # allow for 'nil' block args, which some objc methods accept
+                        arg = ns_from_py(arg)
+                    elif isinstance(arg, Block):
                         arg = arg.block
                     else:
                         arg = Block(arg).block
@@ -1969,6 +1972,8 @@ class BlockDescriptor(Structure):
     _fields_ = [
         ('reserved', c_ulong),
         ('size', c_ulong),
+        ('copy_helper', c_void_p),
+        ('dispose_helper', c_void_p),
         ('signature', c_char_p),
     ]
 
@@ -2064,6 +2069,9 @@ NOTHING = object()
 
 
 class Block:
+
+    _keep_alive_blocks_ = {}
+
     def __init__(self, func, restype=NOTHING, *arg_types):
         if not callable(func):
             raise TypeError('Blocks must be callable')
@@ -2097,13 +2105,20 @@ class Block:
 
         self.literal = BlockLiteral()
         self.literal.isa = addressof(_NSConcreteGlobalBlock)
-        self.literal.flags = BlockConsts.HAS_STRET | BlockConsts.HAS_SIGNATURE
+        self.literal.flags = BlockConsts.HAS_STRET | BlockConsts.HAS_SIGNATURE | BlockConsts.HAS_COPY_DISPOSE
         self.literal.reserved = 0
         self.cfunc = self.cfunc_type(self.wrapper)
         self.literal.invoke = cast(self.cfunc, c_void_p)
         self.descriptor = BlockDescriptor()
         self.descriptor.reserved = 0
         self.descriptor.size = sizeof(BlockLiteral)
+
+        self.copy_helper_type = CFUNCTYPE(c_void_p, c_void_p, c_void_p)
+        self.dispose_helper_type = CFUNCTYPE(c_void_p, c_void_p)
+        self.copy_helper = self.copy_helper_type(self.copyHelper)
+        self.dispose_helper = self.dispose_helper_type(self.disposeHelper)
+        self.descriptor.copy_helper = cast(self.copy_helper, c_void_p)
+        self.descriptor.dispose_helper = cast(self.dispose_helper, c_void_p)
 
         self.descriptor.signature = encoding_for_ctype(restype) + b'@?' + b''.join(
             encoding_for_ctype(arg) for arg in signature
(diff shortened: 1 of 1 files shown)


cculianu:
Oh, this relates to issue #102

rubicon/objc/runtime.py
```diff
@@ -2097,14 +2105,21 @@ def __init__(self, func, restype=NOTHING, *arg_types):
...
+        self.literal.flags = BlockConsts.HAS_STRET | BlockConsts.HAS_SIGNATURE | BlockConsts.HAS_COPY_DISPOSE
         self.literal.reserved = 0
         self.cfunc = self.cfunc_type(self.wrapper)
         self.literal.invoke = cast(self.cfunc, c_void_p)
         self.descriptor = BlockDescriptor()
         self.descriptor.reserved = 0
         self.descriptor.size = sizeof(BlockLiteral)
 
+        self.copy_helper_type = CFUNCTYPE(c_void_p, c_void_p, c_void_p)
+        self.dispose_helper_type = CFUNCTYPE(c_void_p, c_void_p)
```
dgelessus: Minor thing: since the signatures of the helpers are always the same, these `CFUNCTYPE`s could be global constants instead of attributes. This would also allow setting the correct types for the `BlockDescriptor` struct's `copy_helper` and `dispose_helper` fields.

rubicon/objc/runtime.py
```diff
@@ -2097,14 +2105,21 @@ def __init__(self, func, restype=NOTHING, *arg_types):
...
+        self.copy_helper = self.copy_helper_type(self.copyHelper)
+        self.dispose_helper = self.dispose_helper_type(self.disposeHelper)
```
dgelessus: I'm not sure about the naming here - `self.copy_helper` and `self.copyHelper` are identical, except that one is snake_case and the other is dromedaryCase. I think it would be better to give them different names, such as `copy_helper_c` and `copy_helper_py` or something, that would make it less confusing to read.

(Same goes for `dispose_helper` and `disposeHelper`.)

rubicon/objc/runtime.py
```diff
@@ -2097,14 +2105,21 @@ def __init__(self, func, restype=NOTHING, *arg_types):
...
+        self.descriptor.copy_helper = cast(self.copy_helper, c_void_p)
+        self.descriptor.dispose_helper = cast(self.dispose_helper, c_void_p)
```
dgelessus: See above, these casts can be removed if the struct field types are set correctly.

rubicon/objc/runtime.py
```diff
@@ -2113,3 +2128,11 @@ def __init__(self, func, restype=NOTHING, *arg_types):
...
     def wrapper(self, instance, *args):
         return self.func(*args)
+
+    def disposeHelper(self, dst):
+        Block._keep_alive_blocks_.pop(self.block.value, None)
+
+    def copyHelper(self, dst, src):
+        Block._keep_alive_blocks_.pop(self.block.value, None)
+        self.block = objc_block(dst)
+        Block._keep_alive_blocks_[self.block.value] = self
```
dgelessus: I'm not sure if modifying `self` is the right approach here. When a block is copied, the original stays alive and valid, so I think `self` should continue to point to the old block. `Block._keep_alive_blocks_[dst.value] = self` should be enough - that creates an extra reference that stays even if the old block is deleted. And since the new block has the same descriptor as the old one, it also has the same dispose helper, so once the new block is deleted, its entry in the table will be removed as well.

(Actually, that might require a small modification to the dispose helper - it should remove the entry for `dst.value` from the table, not `self.block.value`, otherwise the same entry would be deleted twice.)

cculianu: Ok, I'll look at your suggestions about Python CFUNCTYPE stuff -- I'm not a Python expert so much as I am an ObjC guy.

As for your suggestion about not updating the pointer -- I think this would lead to incorrect behavior, for what we want. 

After ObjC calls block copyHelper, it no longer ever refers to that block by the old pointer in subsequent calls to us on the Python side.  Its concept of the Block is now 'that' new ptr.  

All subsequent calls refer to it by the new pointer (the one passed into dst), which is why I updated the entry.  Otherwise later when disposeHelper is called, we will get the new ptr passed to us, and we won't know what to do with it.

copyHelper is to inform Python that objc took ownership of a copy.  ObjC thereafter always refers to it by the ptr it has, the one it owns.

If we were to dispose of our own fake blocks (which we never do in the actual way we are supposed to), a second call to disposeHelper would occur.  But since we don't actually do this (do we??) it's fine.

Actually, either way it's fine -- since our table is only concerned with keeping code alive that ObjC may call into.  Python's own garbage collection stuff is fine for us.

But I digress:

So I need to update the dict in copyHelper else disposeHelper never works right.  That's the summary.

Basically the contract is as follows, as a sequence of events in time:

1. Python owns the block B right after it creates it.  If it deallocates it, it is responsible for disposing of it (we never actually end up calling block dispose on any blocks we own -- we just let them disappear into the ether -- which is fine since it's only Python data).

2. Some time later that block is passed into an objc function and is retained (basically copied via Block_copy()) -- at which time it is 'taken over' by objective C.  Python is notified of the new copy created in copyHelper.   We don't care about our own private copy -- but we do care about ObjC's copy because we need to remember the pointer to it for later when we want to kill our Python block wrapper from memory..

3. Now, as far as ObjC is concerned, there potentially are up to 2 copies of the blocks -- the one Python owns and the one it took control over.  However the one python owns is not really important and it gets auto-cleaned from memory anyway when the Block instance dies.  The one ObjC owns is of vital concern to us because we need to be informed about when it does so the code it points to (in Python) can get cleaned up.

I am pretty sure the above is accurate.  Note I'm an ObjC guy and not a Python guy (although obviously I do know Python).

I'll take a look at your other suggestions but I am pretty sure this is correct.  Let me know if you disagree.

cculianu: Oh wait -- I just re-read what you wrote.  HA!  Ok -- you're right.  I guess what you say makes tons of sense -- we keep our private pointer for our block the same as it always was.

But -- we insert into our "remember THESE" table the pointer that ObjC is concerned with.

Got it!

Makes sense! :)

rubicon/objc/runtime.py
cculianu: OK, will do.  Note: I'm a C++/ObjC guy so snakeCase is always nicer to me.. ha ha.  Will do.

rubicon/objc/runtime.py
cculianu: Ah I see.  Yeah no sense in keeping that data around in the instance when it's always the same.  Got it.

cculianu:
You know -- you're right. My approach had bugs if Obj ended up sharing that object further down the line.

Your approach is correct.  I just ran some tests. Now.. to tackle your second point about function types in python.. :)

> 
> Copy Helper 106102874398976 -> 106102874374304
> Helper table size now: 1
> Dispose Helper 106102874374304
> Helper table size now: 0
> Copy Helper 106652630203360 -> 106652630202432
> Helper table size now: 1
> Dispose Helper 106652630202432
> Helper table size now: 0
> Copy Helper 105553120594944 -> 105553118612960
> Helper table size now: 1
> Dispose Helper 105553118612960
> Helper table size now: 0
> Copy Helper 106652630224032 -> 106652630221792
> Helper table size now: 1
> Dispose Helper 106652630221792
> Helper table size now: 0
> Copy Helper 106102874373760 -> 106102874399744
> Helper table size now: 1
> Copy Helper 106102878525120 -> 106102874367680
> Helper table size now: 2
> Dispose Helper 106102874399744
> Helper table size now: 1
> Copy Helper 106652630235584 -> 106652630237728
> Helper table size now: 2
> Dispose Helper 106102874367680
> Helper table size now: 1
> Dispose Helper 106652630237728
> Helper table size now: 0
> Copy Helper 107202385982144 -> 107202385987552
> Helper table size now: 1
> Copy Helper 105553120672416 -> 105553120669152
> Helper table size now: 2
> Dispose Helper 107202385987552
> Helper table size now: 1
> Copy Helper 106102874396704 -> 106102874396032
> Helper table size now: 2
> Dispose Helper 105553120669152
> Helper table size now: 1
> Dispose Helper 106102874396032
> Helper table size now: 0
>

cculianu:
Ok, updated.  Note that I'm not entirely sure my usage of CFUNCTYPE is correct -- since like I said I'm a Py->C interop NUB.  But.. I infer it's right.. based on intuition. 

Any thoughts and comments are as always appreciated!

dgelessus:
I'm not sure what's going on with the Travis build right now - all builds except Xcode 6.4 are erroring, because the Python 3 install isn't working. Homebrew is complaining that it can't install Python 3, because Python 2 is already installed. Apparently it's considering Python 2 and 3 as different versions of the same package, and because of that it won't let both be installed at once. This is very strange, because installing Python 2 and 3 side-by-side is a perfectly normal thing to do. I'm not familiar with Homebrew at all - @freakboy3742 do you have any idea what could be going on here?

In any case, this is almost definitely *not* caused by your PR, so you don't need to worry. (But if you have an idea what could be the problem, any help is appreciated 😃)

The Xcode 6.4 build seems to work properly though, and all the tests run fine there. flake8 found [a couple of issues](https://travis-ci.org/pybee/rubicon-objc/jobs/348414486#L2112-L2116), you might want to look at those.

rubicon/objc/runtime.py
```diff
@@ -2057,13 +2065,16 @@ def __call__(self, *args):
         return self.block(*args)
 
 
-_NSConcreteGlobalBlock = (c_void_p * 32).in_dll(libc, "_NSConcreteGlobalBlock")
+_NSConcreteStackBlock = (c_void_p * 32).in_dll(libc, "_NSConcreteStackBlock")
```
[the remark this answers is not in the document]
cculianu: Yeah and it actually caused crash bugs in some cases! (Which is what inspired me to change it).  The stack isa type tells the objc runtime to issue a copy for real ASAP and not do any optimizations omitting it.  I was actually getting dangling-pointer crashes (from the obj-c side).

👍

cculianu:
See -- at the risk of sounding pedantic -- this is why I don't use HomeBrew.  It's great because everybody uses it.  But MacPorts is good too.  And it allows you to install lots of software side-by-side elegantly (including Python).  I had like 4 versions of Python on my system at one point.

Hey man thanks for approving the changes.  I thoroughly tested them (am actually building a full UI in Python calling into UIKit so I end up tripping up the bugs!).

Am really glad it was so "easy" relatively (only a few lines of code) -- this will be a big boon -- the ability to actually use Python closures seamlessly with objc calls is super super great for my productivity! 👍 

As for your HomeBrew woes: No idea. Ha ha. :) People that use HomeBrew generally use 'pyenv' to get around its side-by-side limitations.  And yes, I agree.. not sure why HomeBrew is so difficult about Python .. TBH it's annoying. I switched to MacPorts long ago and never looked back.  I guess Travis is on HomeBrew...?

Good luck.  And thanks for approving this PR so quickly! The limitation on blocks irked me for a few weeks now. :)

rubicon/objc/runtime.py
```diff
@@ -811,7 +811,10 @@ def __call__(self, receiver, *args, convert_args=True, convert_result=True):
...
 
                 if issubclass(argtype, objc_block):
-                    if isinstance(arg, Block):
+                    if arg is None:
+                        # allow for 'nil' block args, which some objc methods accept
+                        arg = ns_from_py(arg)
+                    elif isinstance(arg, Block):
                         arg = arg.block
                     else:
                         arg = Block(arg).block
```
cculianu: I have a concern about this line here -- is it possible that the Block() temporary created here gets destroyed before obj-c has a chance to 'claim' it?  In my testing it works ok -- but I am not sure why and/or if it's guaranteed to always work with this temporary in place.  

I could always research more deeply the guarantees Python offers about how long a temporary stays around -- but the reference count on that temporary Block() instance drops to 0 pretty quickly.. so it makes me nervous.

Is this line 100% ok?

dgelessus:
Yep, the Travis Mac environments all come with Homebrew and a bunch of packages preinstalled. I'm a MacPorts person myself as well, but that wasn't really a conscious decision. At some point I needed to install Wine on OS X, some random tutorial recommended MacPorts, so I used that and stuck with it. I don't even use it to manage my Python installations, I use the prebuilt installers from python.org instead.

I think technically there's nothing stopping us from installing Python on Travis with something other than Homebrew, but since Homebrew is preinstalled it's simpler and much faster to use that. The build issue is probably not that hard to solve either, if I'm reading the error correctly we could probably replace `install` with `upgrade` and be done with it. However I would prefer to at least roughly know what's causing the error, so we can fix the build config properly in a way that won't quickly break again.

rubicon/objc/runtime.py
dgelessus: I'm probably missing something, but right now I don't see why this code would work at all - `Block` objects don't have a `block` attribute, only `ObjCBlockInstance`s do. And I think you're right, there's nothing stopping the block from being collected before its pointer is passed to a method. To prevent this, I think the pointer object would need to keep a reference to the block object - or maybe the block object should simply have an `_as_parameter_` attribute, so it can be passed into C functions as-is.

cculianu: Oh man, thanks for responding.

I can attest to it magically working. I have tested that codepath by passing naked Python callables to obj-c methods with all sorts of semantics from immediate to asynchronous -- and they work -- the copy_helper is invoked pretty much right away without crashing despite everything -- I guess we got lucky -- I'm surprised Python allows for objects with 0 refcount, marked for deletion, to be resurrected later!  Go Python! :)

Indeed the Block class does have the .block attribute, akin to the .ptr attribute for ObjCInstance, which is basically the raw pointer. It's assigned in the very last line of the constructor: 

``self.block = cast(byref(self.literal), objc_block)``


Anyway -- I'll go ahead and update this PR to mitigate the risk of crash created here by doing either what I propose (a list), or I can investigate this _as_parameter_ mechanism you speak of (not familiar with it but hey, an opportunity to learn!)

Thanks for responding quickly!

cculianu:
Ha man I wish I could help. I'm a total Travis nub, to be honest.  Glad to find another MacPorts user! We are inexplicably few and far between! :)

PS: I updated the PR to reflect your suggestion about using \_as_parameter\_. Neato!  I like the elegance involved!

rubicon/objc/runtime.py
```diff
@@ -811,10 +811,16 @@ def __call__(self, receiver, *args, convert_args=True, convert_result=True):
                     arg = arg.value
 
                 if issubclass(argtype, objc_block):
-                    if isinstance(arg, Block):
+                    if arg is None:
+                        # allow for 'nil' block args, which some objc methods accept
+                        arg = ns_from_py(arg)
+                    elif isinstance(arg, Block):
                         arg = arg.block
```
dgelessus: Now that `Block`s have `_as_parameter_`, this `elif` branch should not be needed anymore - I think we can remove this and change the `else` below to `elif callable(arg)`.

rubicon/objc/runtime.py
```diff
@@ -2096,20 +2110,43 @@ def __init__(self, func, restype=NOTHING, *arg_types):
...
+
         self.descriptor.signature = encoding_for_ctype(restype) + b'@?' + b''.join(
             encoding_for_ctype(arg) for arg in signature
         )
         self.literal.descriptor = cast(byref(self.descriptor), c_void_p)
         self.block = cast(byref(self.literal), objc_block)
 
+    @property
+    def _as_parameter_(self):
+        return self.block
```
dgelessus: If the value for `_as_parameter_` is always the same for each instance, you can also use a regular attribute instad of a property. Doing `self._as_parameter_ = self.block` in `__init__` should work - that's how it's done [in `ObjCInstance`](https://github.com/pybee/rubicon-objc/blob/master/rubicon/objc/runtime.py#L1284) as well.

There are situations where a property is better than an attribute, for example if `block` was meant to be changeable by the user. But since that isn't the case here, I think an attribute is better, because it's shorter and consistent with our other code.

rubicon/objc/runtime.py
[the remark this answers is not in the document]
dgelessus: Ah, sometimes I was getting random crashes related to blocks as well when running the unit tests. I'll need to check if they are fixed once this PR is merged - if so, that would be great! 😃

rubicon/objc/runtime.py
cculianu: Ok, done.  Old C habit I guess that I did it that way -- sort of defensive programming was the rationale.  But yeah, if it's consistent with how it's already done, sure.  Changed.  Will do your other suggestion shortly too.

rubicon/objc/runtime.py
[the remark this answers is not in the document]
cculianu: Do let me know if that's the case! 👍

rubicon/objc/runtime.py
cculianu: Ok, done.

cculianu:
Something just occurred to me.  

All of the logic in that method marshalling code could be simplified simply by adding callables/Blocks/etc to ns_from_py which returns a Block.. can't it??

Wouldn't that make everything more flexible going forward?

...

Only thing I'm unclear about is the difference between ObjCBlock (which I presume is a facade around actual objc pointers to blocks) and our Block.  I suppose in either case we just return back the passed-in arg in ns_from_py, if they are passed in.

Or am I missing something here?

cculianu:
I just implemented another approach to this -- moving all this logic into ns_from_py (which is where it belongs IMHO).  What do you think?  I may create another PR for it.  Or should I commit it to this branch?

It basically has a lot fewer lines of code and is much more flexible IMHO.  Only issue I can think of is that in the case where a callable is passed to ns_from_py(), it returns a Block.. which *technically* isn't a foundation object but can be used interchangably as one on the python side.

Thoughts?

dgelessus:
Hm, I don't remember why exactly the callable-to-block conversion isn't done in `ns_from_py`. I think part of the reason is that the conversion only happens when the parameter type is a block type, and that can't be checked in `ns_from_py` obviously.

I'm not quite sure whether this special check is worth it or not. One could argue that it's an unnecessary special case, and that there's no reason to *not* convert callables to blocks in other contexts as well. On the other hand, passing a block into a non-block argument is very unusual - in Objective-C there's no way to tell at runtime whether an object is a block, let alone what its signature is, so it's very unlikely that you'd want to pass a block into an `NSObject *` or `id` parameter.

There's also the issue that a `Block` isn't an `ObjCInstance`, unlike all other types returned by `ns_from_py`. This would be an issue if the code calling `ns_from_py` expects to be able to send messages to the returned object. A possible solution to that would be to merge the three `ObjCBlock`, `ObjCBlockInstance` and `Block` classes into one class that has the features of all three combined - an `ObjCInstance` that can be constructed from an existing block object or a Python callable, and that can be called from Python.

But in any case I think it would be better to do that in a separate PR, since this would be an additional feature and not a necessary fix for this PR.

cculianu:
Yep, in obj-c, blocks are pretty much obj-c classes.  They accept messages such as 'class' and 'description', etc.  I even think they take a retain message if you aren't using ARC, in which case Block_copy() is called (I think).

I felt a little weird to be honest adding Block conversion to ns_from_py.  But the odd thing is that our block object is a pseudo-NS object.  It does have a pointer (accessible as .block) which NS knows how to use.  In fact I even think you can sent it messages using the low-level api (and using its pointer).

It's not *quite* a real NS object from our Python scheme of things, though, like you say, because it doesn't fit into the ObjCInstance heirarchy (unfortunately)... but you can pass it around and pass it back to ObjC methods.  So it sort of makes sense.

I agree that ideally all 3 of those types get merged into 1 intelligent type which behaves like a first class obj-c object on the python and obj-c side symmetrically and uniformly.

That being said, I'm ok with removing the conversion to Block from ns_from_py and going back to that special case in the method call marshalling code-- but it sounds like you think this conversion is a good idea, provided we merge the object hierarchy for Blocks to be instances of ObjCInstance. 

With or without the merger I sort of like the conversion as it makes a lot of automatic conversion code easier to write.  It modifies the contract slightly though -- ns_from_py may return an ObjCInstance.. or it may return a Block.  Before it always returned an ObjCInstance (or failed).  But all return types at least are suitable to be passed back to ObjC.  I guess in my mind ns_from_py returns types suitable to be passed to NS, but I see your point in that you ideally want ObjCInstance to encapsulate everything-NS.

So if you're ok with keeping the conversion for now -- I can work on a separate PR that merges all 3 of those Block classes to inherit from ObjCInstance.. does that sound like a deal? (My preference is this.)

Or, alternatively, if you don't want this conversion in there right now (without said-PR), I can remove it too (and add it later once that separate PR is ready).  Let me know what to do!

I am excited to make this library more useful.  It's scratching a personal itch.  I am using it heavily to build a UIKit-based app and the more niceties in it, the better for my productivity. :)

dgelessus:
> but it sounds like you think this conversion is a good idea, provided we merge the object hierarchy for Blocks to be instances of ObjCInstance.

Exactly. `ns_from_py` is supposed to always return an `ObjCInstance` of some kind (or `None`), but at the moment `Block` isn't one. I don't know if it would cause any problems in practice, but I think it would be better not to do it until `Block` is an `ObjCInstance`, just to be sure.

Also, since this PR is finished and working, it's cleaner and safer to implement additional block-related features in a separate PR.

Also, I think I now have a working configuration for Travis agian, see #106. Once that PR is merged, you can merge pybee/master into your branch to get the Travis fix.

cculianu:
Ah, ok. Looking forward to the working changes to Travis.  I noticed you started that.

OK, understood and agreed.  I'll modify my branch to take out the recent modifications to ns_from_py, and we'll put that on the table for another PR where I volunteer to take a stab at unifying the many faces of the BlockInstance god. :)

Sounds great man. Thanks so much for your guidance and thoughts and work on all of this!! 👍

dgelessus:
OK, the Travis build is now fixed in the master branch - if you merge pybee/master into your PR branch, the Travis build here should work again too.

cculianu:
Ok, merged

cculianu:
Hmm. It appears to have failed on some 'flake8' command.  Is that due to the formatting/style warnings it issued or is there still an actual problem with the travis environment?

dgelessus:
Yes, `flake8` is only the style checker. To make that succeed, the style problems listed [in the log](https://travis-ci.org/pybee/rubicon-objc/jobs/349637569#L2574-L2580) need to be fixed.

The unit tests ran successfully, which is a good sign - when I was testing Travis, I often encountered a build that failed because of the random block-related crash mentioned above. It's hard to say for sure with bugs that only appear at random, but it looks like you've probably fixed that! :+1:

cculianu:
Ah, ok.  

I installed flake8, ran it on the runtime.py file, did what it said.  No errors now here.

Ok, committed one last revision (I hope!)

As for the tests passing -- awesome!  Yeah -- it was crashing in my code too before (I was using blocks heavily) .. but.. no more!

dgelessus:
And done. Thank you very much for the PR! 😃

cculianu:
Dude, thank YOU for working so closely and being so responsive on it.  I learned a lot working on it (about Python, mostly, and a little obj-c internals).  To me, this bug was a killer because it severely limited which parts of the UIKit API were open to me -- and that slowed me down in my work.  I am so glad it's merged. Thanks!

PR #106, opened 2018-03-03
Install and test with multiple Python versions on Travis using pyenv

dgelessus:
This changes the Travis config to use `pyenv` instead of `brew` to install Python.

The original motivation for this change was that our Travis builds are currently broken. `brew install python3` is failing, because the Travis environment has Python 2 installed by default, and for some reason Homebrew is refusing to install Python 3 because of that. (I think it's because Homebrew's handling of Python 2/3 has changed - `python3` is now an alias for `python`, which refers to Python 3, and if you want Python 2 you need to explicitly use `python@2`. However the Travis environment has Python 2.7 installed as `python` rather than `python@2`, which is why it causes a conflict when we try to install Python 3.)

However Travis also comes with `pyenv` preinstalled, which provides much more control about what Python versions are installed than Homebrew does. And since, unlike Homebrew, it allows installing multiple Python versions at once, I decided to try installing all Python versions that we have `tox` configurations for. This should allow us to run `tox` on Travis.

~~This PR is still WIP while I figure out how `pyenv` works (haven't used it before) - please don't merge yet!~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ install:
   # The different Travis environments have different pyenv versions, and not all have the Python versions we need,
   # so we upgrade pyenv to get a consistent setup in all environments.
   - brew update
-  - brew upgrade pyenv
+  - brew outdated pyenv || brew upgrade pyenv
   # pyenv's shims directory is not in the PATH by default.
   - export PATH="$(pyenv root)/shims:${PATH}"
   # These are the Python versions that we want to test on.
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,28 @@ branches:
 os: osx
 matrix:
   include:
+    # Note: The ordering of the environments is optimized so that the complete matrix runs as quickly as possible,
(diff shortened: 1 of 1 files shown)


dgelessus:
OK, this should be working and finished now. I've squashed away all my trial-and-error commits, no need to keep those around.

I ended up not testing on PyPy, since for some reason it's being very annoying to build/install on Travis. I also enabled Travis's cache mechanism for the pyenv directory, so that the Python versions don't need to be rebuilt from scratch for every build. (I wasn't able to test that part yet - pull request builds don't update the cache, they only fetch it, which does nothing since no cache has been uploaded yet.)

dgelessus:
Turns out I was wrong - PR builds do update the cache after all, but they have their own cache that's separate from the main repo. So it looks like the caching works properly.

I've also optimized the order of the environments slightly, so that the fastest environment finishes first, while using as little time as possible for the entire build matrix. See the comments for details. (I *may* have overthought this a little, it probably won't make a huge difference in practice. 😛)

Issue #107, opened 2018-03-09
Is inheritance working fully on the Python side?


cculianu:
Can I inherit form a class that inherits from an ObjC object?

What I'm thinking of is the following code:

```
class MyA(NSObject):
    
    aProp = objc_property()
    
    @objc_method
    def init(self) -> ObjCInstance:
        self = ObjCInstance(send_super(self, 'init'))
        print ("MyA init")
        self.aProp = "MyA Property"
        return self
    
    @objc_method
    def dealloc(self) -> None:
        print ("MyA dealloc")
        send_super(self, 'dealloc')

class MyB(MyA):
    
    
    @objc_method
    def init(self) -> ObjCInstance:
        self = ObjCInstance(send_super(self, 'init'))
        print ("MyB init")
        self.aProp = "MyB Property"
        return self
    
    @objc_method
    def dealloc(self) -> None:
        print ("MyB dealloc")
        send_super(self, 'dealloc')
```

Then, I should be able to do:

```
a = MyB.new().autorelease()
print("a prop = %s"%(str(a.aProp)))
```

If I do that, I get the following error and crash at runtime:

```
Traceback (most recent call last):
  File "_ctypes/callbacks.c", line 234, in 'calling callback function'
  File "/Users/calin/Library/Developer/CoreSimulator/Devices/80F9328C-18C7-43C1-9E93-F4CB128AD3CA/data/Containers/Bundle/Application/009B9605-E170-441C-AE5A-DF4CFBC80BDE/Electron-Cash.app/Library/Application Support/com.c3-soft.ElectronCash/app_packages/rubicon/objc/runtime.py", line 1040, in _objc_method
    result = f(py_self, *args)
  File "/Users/calin/Library/Developer/CoreSimulator/Devices/80F9328C-18C7-43C1-9E93-F4CB128AD3CA/data/Containers/Bundle/Application/009B9605-E170-441C-AE5A-DF4CFBC80BDE/Electron-Cash.app/Library/Application Support/com.c3-soft.ElectronCash/app/ElectronCash/electroncash_gui/ios_native/send.py", line 25, in init
    self = ObjCInstance(send_super(self, 'init'))
[...]

PR #108, opened 2018-03-10
Proposed fix to send_super 

cculianu:
Related: Issue #107 

I'm not sure if my approach is the most elegant.. but at least it has the same semantics as what obj-c does for you at compile-time.

The idea is the caller tells send_super where to start looking, which 100% emulates what the compiler does in obj-c for you.  

If they omit the keyword, they get the (possibly incorrect) automatic behavior we had before.

Not sure how elegant this is though.  Maybe we can do better.  But this is better than nothing IMHO.

Let me know what you think!

Tested on the following code and it works like a charm:

Code:

```
class MyA(NSObject):
    
    aProp = objc_property()
    
    @objc_method
    def init(self) -> ObjCInstance:
        #self = ObjCInstance(send_super(self, 'init', superclass='NSObject'))
        self = ObjCInstance(send_super(__class__, self, 'init'))
        print ("MyA init")
        self.aProp = "MyA Property"
        return self
    
    @objc_method
    def dealloc(self) -> None:
        print ("MyA dealloc")
        self.aProp = None
        #send_super(self, 'dealloc', superclass='NSObject')     
        send_super(__class__, self, 'dealloc')     

class MyB(MyA):
    
    bProp = objc_property()
    
    @objc_method
    def init(self) -> ObjCInstance:
        #self = ObjCInstance(send_super(self, 'init', superclass='MyA'))
        self = ObjCInstance(send_super(__class__, self, 'init'))
        print ("MyB init")
        self.aProp = "MyB Property, overwritten"
        self.bProp = "MyB's own property"
        return self
    
    @objc_method
    def dealloc(self) -> None:
        print ("MyB dealloc")
        self.bProp = None
        #send_super(self, 'dealloc', superclass='MyA')
        send_super(__class__, self, 'dealloc')

# ... then, somewhere in a function or whatever ... 

    a=MyB.new().autorelease()
    print("MyB propA=%s propB=%s"%(str(a.aProp),str(a.bProp)))



```
Output:

```
MyA init
MyB init
MyB propA=MyB Property, overwritten propB=MyB's own property
MyB dealloc
MyA dealloc

```

__EDIT__: Updated this comment to reflect changed API as discussed below (so that someone googling has good usage to look at in front of their eyes rather than non-existant usage).

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -681,9 +681,20 @@ def send_super(receiver, selName, *args, **kwargs):
     else:
         raise TypeError("Invalid type for receiver: {tp.__module__}.{tp.__qualname__}".format(tp=type(receiver)))
 
+    # Note that in objective-c, the compiler provides this information implicitly at compile-time, but we must
+    # have the caller do it at runtime in order to get 100% correct behavior. See github issue #107.
     superclass = kwargs.get('superclass', None)
-    if superclass is not None and not isinstance(superclass, Class):
-        raise TypeError("Invalid type for 'superclass' kwarg, expected type 'Class', got: %" % (str(type(superclass))))
+    if superclass is not None:
+        if isinstance(superclass, Class):
+            pass  # ok, accept Class instances, I guess
+        elif not isinstance(superclass, str):  # expecting a string here.. complain
+            raise TypeError("Invalid type for 'superclass' kwarg, expected type str, got: %"
+                            % (str(type(superclass))))
+        else:
+            classname = superclass
+            superclass = get_class(classname)
+            if superclass.value is None:
+                raise TypeError("Could not find an obj-c class named '%s'" % (classname))
     superclass = get_superclass_of_object(receiver) if superclass is None else superclass
     super_struct = objc_super(receiver, superclass)
     selector = SEL(selName)
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -688,7 +688,7 @@ def send_super(receiver, selName, *args, **kwargs):
         if isinstance(superclass, Class):
             pass  # ok, accept Class instances, I guess
         elif not isinstance(superclass, str):  # expecting a string here.. complain
-            raise TypeError("Invalid type for 'superclass' kwarg, expected type str, got: %"
+            raise TypeError("Invalid type for 'superclass' kwarg, expected type str, got: %s"
                             % (str(type(superclass))))
         else:
             classname = superclass
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -682,14 +682,12 @@ def send_super(cls, receiver, selName, *args, **kwargs):
 
     cls may also be a naked Class pointer.
     """
-    if isinstance(cls, Class) or type(cls) is ObjCClass:
-        pass
-    else:
+    if not isinstance(cls, (ObjCClass, Class)):
         # Kindly remind the caller that the API has changed
         raise TypeError("Missing/Invalid cls argument: '{tp.__module__}.{tp.__qualname__}' -- "
                         .format(tp=type(cls))
                         + "send_super now requires its first argument be an"
-                        + " ObjClass or an objc raw Class pointer."
+                        + " ObjCClass or an objc raw Class pointer."
                         + " To fix this error, use Python's __class__ keyword as the first argument to"
                         + " send_super.")
 
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -674,12 +674,14 @@ def send_super(cls, receiver, selName, *args, **kwargs):
     may find the appropriate superclass method implementation (if any) to call. (See issue #107)
 
     As such, cls can be a string, which is the class name of the lexical scope of the caller,
-    such as given by the __class__ keyword in Python.  Eg: send_super(__class__, self, 'init')
-    would be the typical usage pattern.
+    such as given by the __class__.__name__ built-in in Python, or a Python 'class' type as
+    obtained from the __class__ built-in.
+
+    Eg: send_super(__class__, self, 'init') would be the typical usage pattern.
 
     cls may also be an ObjCClass instance or a naked Class pointer.
     """
-    if not isinstance(cls, (str, ObjCClass, Class)):
+    if not isinstance(cls, (str, ObjCClass, Class)) and not inspect.isclass(cls):
         # Kindly remind the caller that the API has changed
         raise TypeError("Missing/Invalid cls argument: {tp.__module__}.{tp.__qualname__} -- "
                         + "send_super now requires its first argument be a"
@@ -689,8 +691,11 @@ def send_super(cls, receiver, selName, *args, **kwargs):
                         .format(tp=type(cls)))
 
     if not isinstance(cls, ObjCClass):
-        # Convert class name argument to an ObjCClass from either a name or a naked Class pointer
-        cls = ObjCClass(cls)  # raises a NameError if class is not found and/or pointer was nil
+        if inspect.isclass(cls):
+            # __class__ was passed-in, convert to str for obj-c lookup
+            cls = cls.__name__
+        # Convert class name/ptr argument to an ObjCClass (from either a name or a naked Class pointer)
+        cls = ObjCClass(cls)  # raises an exception if class is not found by name and/or pointer was nil
 
     try:
         receiver = receiver._as_parameter_
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -664,11 +664,34 @@ class objc_super(Structure):
 
 
 # http://stackoverflow.com/questions/3095360/what-exactly-is-super-in-objective-c
-def send_super(receiver, selName, *args, **kwargs):
+def send_super(cls, receiver, selName, *args, **kwargs):
     """Send a message named selName to the super of the receiver.
 
     This is the equivalent of [super selname:args].
+
+    Since proper 'super' semantics requires lexical scoping, the cls argument is now required.
+    It basically instructs this runtime what the lexical scope was of the caller, so that it
+    may find the appropriate superclass method implementation (if any) to call. (See issue #107)
+
+    As such, cls can be a string, which is the class name of the lexical scope of the caller,
+    such as given by the __class__ keyword in Python.  Eg: send_super(__class__, self, 'init')
+    would be the typical usage pattern.
+
+    cls may also be an ObjCClass instance or a naked Class pointer.
     """
+    if not isinstance(cls, (str, ObjCClass, Class)):
+        # Kindly remind the caller that the API has changed
+        raise TypeError("Missing/Invalid cls argument: {tp.__module__}.{tp.__qualname__} -- "
+                        + "send_super now requires its first argument be a"
+                        + " class name (str) or an instance of ObjCClass and/or Class."
+                        + " To fix this error, use Python's __class__ keyword as the first argument to"
+                        + " send_super."
+                        .format(tp=type(cls)))
+
+    if not isinstance(cls, ObjCClass):
+        # Convert class name argument to an ObjCClass from either a name or a naked Class pointer
+        cls = ObjCClass(cls)  # raises a NameError if class is not found and/or pointer was nil
+
     try:
         receiver = receiver._as_parameter_
     except AttributeError:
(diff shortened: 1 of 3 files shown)


dgelessus:
I think this is the right approach - while I agree that explicitly passing in the superclass isn't very pretty, it's the most straightforward way of implementing this. `send_super` isn't used very frequently either, so I don't think the extra typing is a big issue.

I have a couple of comments/questions though:

* Any particular reason why `superclass` is the class name as a string? I saw that you originally made `superclass` be a `Class` pointer, but changed that to avoid needing to call `get_class` every time. Wouldn't it be easier to allow `superclass` to be either an `ObjCClass` or `Class`, and unwrap it if necessary? That's how `receiver` is handled (it can be an `ObjCInstance` or `objc_id`).
* For symmetry with the long form of Python's `super`, I think it would be better if the extra `send_super` argument would be the current class, rather than its superclass. This would also make refactoring less error-prone - you don't need to change every `send_super` call when the superclass changes.
  * This would also allow us to make use of a somewhat obscure Python feature: all functions inside a Python class definition have access to a special variable named `__class__`, which is set to the class object that's currently being created. For example inside a `class Foo` block, `__class__` is `Foo`. This means that all `objc_super` calls can be written as `objc_super(self, "method", ..., cls=__class__)` and they will work regardless of what the current class is called.
* I think it would be a good idea to make the "current class" argument required, because without it `send_super` won't behave correctly. If we make use of `__class__` as explained above, it wouldn't be a lot of extra typing work: `send_super(self, "method")` becomes `send_super(__class__, self, "method")`.
  * If we do this, old code that doesn't pass `__class__` will stop working - either because of too few arguments, or because `receiver` ends up being a string (the method name) due to the parameters shifting. I think it's okay to break backwards compatibility here - Rubicon is not stable yet, `send_super` is not used very often, and the old behavior is broken. We can even detect "old-style" calls that are missing the class, by checking whether `receiver` is a string, and throw an error with a helpful message in that case.

cculianu:
First, your question:

> Any particular reason why superclass is the class name as a string? I saw that you originally made superclass be a Class pointer, but changed that to avoid needing to call get_class every time. Wouldn't it be easier to allow superclass to be either an ObjCClass or Class, and unwrap it if necessary? That's how receiver is handled (it can be an ObjCInstance or objc_id).

Sure, I agree we should support that.  Although I see advantages to supporting a string too (less typing!).  In fact -- I think you'd be ok with that given your \_\_class\_\_ idea.  (I figured a string was easier to type, and I noticed 'get_class' isn't exported to the rubicon.objc namespace by default, and I wanted to be minimally intrusive with this change... so that was the rationale for going with the string).

As for the test of your comment:

Oh man!  I love this idea!  I was *super* bothered by the fact that manually typing the classname for superclass= is very very bug prone.  This is doubly true if you go around changing the inheritance heirarchy.  You have to also remember to change all calls to send_super.  I couldn't figure out how else to lexically scope this.  Turns out Python has this \_\_class\_\_ keyword which is lexically scoped perfectly for us!

I love this approach that you propose and can see only benefit to it.

It's true, it breaks existing calls.  If we issue a friendly message, as you say, all calls can quickly be found and corrected.

I'm all for this approach you propose.  It fixes the potential maintainability nightmare of having subtle errors because you changed your object hierarchy but missed the odd "send_super" call.

In ObjC, even though they don't use that term in their docs -- 'super' is *lexically* scoped.  We need a lexically scoped mechanism to get it right.  Enter ther \_\_class\_\_ keyword which is just that.  I love it.

So send_super isn't used much eh?  Ha -- I use it all the time.  But I can see how not everyone is crazy enough to actually inherit from UIKit classes in Python in their iOS project like I do. :)  I noticed the Toga framework uses an entirely different pattern to do its magic, for example.

I'll go ahead and change this PR to do what you say (I have to do a bit of digging on how to get from the 'class' to its superclass but I trust ObjC has easy methods to do it -- if not runtime.py having helpers to do it already).

Thanks so much for your very thoughtful guidance on this!  I'm really glad this bug is getting fixed! 

I also noticed you just made a PR that addresses a potential python warning related to this.  Thanks! 👍

cculianu:
Ok, done.  Let me know what you think!

Also note that flake8 on my system __*flaked*__ (heh) on the \_\_class\_\_ built-in.  It complained it doesn't exist.  When it does -- code runs fine and Python does the right thing.  Likely a bug in flake8.  Not sure how to handle this.. :/

cculianu:
Yeah, tests got that DeprecatedWarning, plus flake8.. barfs on \_\_class\_\_.. how to fix? :(

dgelessus:
The `__class__` issue has been fixed in `pyflakes` already (PyCQA/pyflakes#326), but there hasn't been a release containing the fix yet. I've updated #109 to change our `flake8` test environment to use a development version of `pyflakes`, that way it will understand `__class__`. So once that PR is merged, the tests should run again.

rubicon/objc/runtime.py
```diff
@@ -664,11 +664,39 @@ class objc_super(Structure):
...
+                        + "send_super now requires its first argument be a"
+                        + " class name (str) or an instance of ObjCClass and/or Class."
+                        + " To fix this error, use Python's __class__ keyword as the first argument to"
+                        + " send_super."
+                        .format(tp=type(cls)))
+
+    if not isinstance(cls, ObjCClass):
+        if inspect.isclass(cls):
+            # __class__ was passed-in, convert to str for obj-c lookup
+            cls = cls.__name__
```
dgelessus: This check isn't necessary; `__class__` is an `ObjCClass` object if you use it inside an `ObjCClass`:

```python
>>> from rubicon import objc
>>> class Foo(objc.NSObject):
...     @objc.objc_method
...     def foo(self) -> None:
...         print(type(__class__), __class__)
...         
>>> Foo.alloc().init().foo()
<class 'rubicon.objc.runtime.ObjCClass'> ObjCClass('Foo')
```

rubicon/objc/runtime.py
```diff
@@ -664,11 +664,39 @@ class objc_super(Structure):
...
+        # Convert class name/ptr argument to an ObjCClass (from either a name or a naked Class pointer)
+        cls = ObjCClass(cls)  # raises an exception if class is not found by name and/or pointer was nil
```
dgelessus: Since `cls` is only needed for its superclass, I think it would be better to leave it as a raw `Class` and use the `get_superclass_of_class` function to get its superclass.

I'm honestly surprised that using `ObjCClass` here is possible at all, without causing any bootstrapping issues - usually you cannot (unconditionally) use the high-level `ObjCInstance` and `ObjCClass` inside `send_message` and `send_super`, because these functions are used in the implementation of `ObjCInstance`.

rubicon/objc/runtime.py
```diff
@@ -664,11 +664,39 @@ class objc_super(Structure):
...
+def send_super(cls, receiver, selName, *args, **kwargs):
     """Send a message named selName to the super of the receiver.
 
     This is the equivalent of [super selname:args].
+
+    Since proper 'super' semantics requires lexical scoping, the cls argument is now required.
+    It basically instructs this runtime what the lexical scope was of the caller, so that it
+    may find the appropriate superclass method implementation (if any) to call. (See issue #107)
+
+    As such, cls can be a string, which is the class name of the lexical scope of the caller,
```
dgelessus: I'm still not sure when using a string for `cls` would be needed. In all normal use cases, you'll pass `__class__` as the first argument. If for whatever reason that doesn't work, you can use `ObjCClass` explicitly to get a class with a specific name.

rubicon/objc/runtime.py
cculianu: Whaa.. ok.  Yeah that makes sense.  You Python-Fu is strong. Much respect.

rubicon/objc/runtime.py
cculianu: Ok, so basically grab the cls.ptr and use that, got it.  No messing about with strings or other "smart" stuff. Only complain if passed-in cls is not an ObjCClass, right?

Got it!

rubicon/objc/runtime.py
cculianu: Yeah you're right.  Will fix.

Man your Python Kung Fu is strong.  Thanks for taking the time to explain this to me.  Like I said, I'm a C++/C/Obj-C guy -- Python is relatively new to me (really only learned it properly for this project I'm working on).

Very cool! Thanks!

cculianu:
Question: I noticed this in the code in class ObjCClass:

```
    def __instancecheck__(self, instance):
        if isinstance(instance, ObjCInstance):
            return bool(instance.isKindOfClass(self))
        else:
            return False

    def __subclasscheck__(self, subclass):
        if isinstance(subclass, ObjCClass):
            return bool(subclass.isSubclassOfClass(self))
        else:
            raise TypeError(
                'issubclass(X, {self!r}) arg 1 must be an ObjCClass, not {tp.__module__}.{tp.__qualname__}'
                .format(self=self, tp=type(subclass))
            )
```

Note the `isSubclassofClass` and `isKindOfClass` calls into objc missing the trailing `_`.  Is this a "feature" of rubicon-objc I'm unaware of or is this an oversight/bug?

dgelessus:
That is indeed a feature. Rubicon lets you use two different styles for calling methods with arguments. For example, if you want to call the method [`-[NSMutableArray replaceObjectsInRange:withObjectsFromArray:range]`](https://developer.apple.com/documentation/foundation/nsmutablearray/1409073-replaceobjectsinrange?language=objc), you can use either of the following two method calls:

* `arr1.replaceObjectsInRange_withObjectsFromArray_range_(range1, arr2, range2)`
* `arr1.replaceObjectsInRange(range1, withObjectsFromArray=arr2, range=range2)`

The first style is what you're probably used to - the Python method name is the Objective-C method name with all colons replaced with underscores.

The second style tries to mimic Objective-C's method call syntax a little, by placing the method arguments "in between" the parts of the method name. The first method name part becomes the Python method name, the first argument is passed as a positional argument (right after the opening paren), and all further method name parts and arguments become Python keyword arguments. It's a bit lengthy to explain, it's probably easier to understand by looking at the example code above.

When you call a single-argument method and leave off the underscore at the end, you're actually using the second method call style - but because the method only takes a single argument, the call doesn't use any keyword arguments.

cculianu:
Wow.. you just blew my miiiind.

Yeah, I get it.  Clever that you support both styles.  The second style is more "objective-c" ish.  Also kind of more Pythonesque if you think about it. HA!  Didn't know that.  Will have to read that relevant part of runtime.py to get a sense for how the trickery is done.

Awesome! Thanks for explaining that. 👍

rubicon/objc/runtime.py
```diff
@@ -664,11 +664,35 @@ class objc_super(Structure):
...
+    an instance method implementation of a class descending from ObjCClass -- which all your custom
+    obj-c classes will always descend from).
+
+    Eg: send_super(__class__, self, 'init') would be the typical usage pattern.
+
+    cls may also be a naked Class pointer.
     """
+    if isinstance(cls, Class) or type(cls) is ObjCClass:
+        pass
+    else:
```
dgelessus: Minor thing: this could be written as `if not isinstance(cls, (Class, ObjCClass))`, right?

rubicon/objc/runtime.py
```diff
@@ -664,11 +664,35 @@ class objc_super(Structure):
...
+        # Kindly remind the caller that the API has changed
+        raise TypeError("Missing/Invalid cls argument: '{tp.__module__}.{tp.__qualname__}' -- "
+                        .format(tp=type(cls))
+                        + "send_super now requires its first argument be an"
+                        + " ObjClass or an objc raw Class pointer."
```
dgelessus: Typo: `ObjClass` should be `ObjCClass`.

cculianu:
Awesome.  Can't wait.  Thank you so much for everything man! 👍

cculianu:
Hey man -- You been busy?  I noticed neither this PR nor the one fixing the deprecated warnings got merged.  Just was wondering if you've been busy or what's up?

Best regards,

-Calin

dgelessus:
Yep, things have been somewhat busy lately, so I wasn't able to work much on Rubicon for a few weeks - sorry about that.

The reason why this PR couldn't be merged yet is because it requires #109 for the CI build to pass, and that PR couldn't be merged yet because it hasn't been approved yet (because we have [required reviews](https://help.github.com/articles/about-required-reviews-for-pull-requests/) enabled). @freakboy3742 Apologies for pinging you again, I know you're busy with PyCon preparations and such at the moment - but when you have time, could you have a look at #109? Thanks 😃

freakboy3742:
@dgelessus Apologies for the delay reviewing #109; I've done that now, and kickstarted a new build (although now that I've done that... the #109 changes won't be included, so I don't know why I pushed that button... I'm chalking it up to being in transit for (checks watch) 35 hours... :-P)

dgelessus:
Well, the CI build passed anyway, go figure 😆 I'm not quite sure what's going on - the PR build already seems to be using the updated `tox.ini` somehow? In any case, since the only build failure before was the `flake8` check, there should be no issues with merging this.

Thank you for the quick response @freakboy3742! That's quite a long trip, hope you'll arrive soon 😃

PR #109, opened 2018-03-10
Fix DeprecationWarning when using __class__ in an Objective-C subclass 

dgelessus:
On Python 3.6 and later, when a class definition uses the special `__class__` variable, the metaclass receives a `__classcell__` entry in the class namespace. The metaclass must retain this entry in the namespace passed to `type.__new__`. ([Relevant Python docs](https://docs.python.org/3/reference/datamodel.html#creating-the-class-object).)

When creating an Objective-C subclass in Python, `ObjCClass` discards the entire namespace (because everything was translated to Objective-C methods, properties, etc.), which means that `__classcell__` is removed from the namespace and not passed on automatically. This PR adds a special case for `__classcell__` to keep it around.

This is relevant to #108.

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -1431,6 +1431,7 @@ class ObjCClass(ObjCInstance, type):
         protocols_ptr = libobjc.class_copyProtocolList(self, byref(out_count))
         return tuple(ObjCProtocol(protocols_ptr[i]) for i in range(out_count.value))
 
+    @classmethod
     def _new_from_name(cls, name):
         name = ensure_bytes(name)
         ptr = get_class(name)
@@ -1439,6 +1440,7 @@ class ObjCClass(ObjCInstance, type):
 
         return ptr, name
 
+    @classmethod
     def _new_from_ptr(cls, ptr):
         ptr = cast(ptr, Class)
         if ptr.value is None:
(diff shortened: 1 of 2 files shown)


dgelessus:
Update: I've temporarily changed the `flake8` `tox` environment to use a development version of `pyflakes`, so that it understands `__class__`. (The latest release of `pyflakes`, 1.6.0, doesn't know about `__class__` and complains when you use it.) This dependency should be removed again once a new version of `pyflakes` is released.

cculianu:
Awesome! 👍

rubicon/objc/runtime.py
```diff
@@ -1560,7 +1558,19 @@ class with those properties is created and returned. This form is usually called
             # which need to be kept from being garbage-collected.
             # It does not contain any other methods, do not use it for calling methods.
             'imp_keep_alive_table': {},
-        })
+        }
+
+        # On Python 3.6 and later, the class namespace may contain a __classcell__ attribute that must be passed on
+        # to type.__new__. See https://docs.python.org/3/reference/datamodel.html#creating-the-class-object
+        if '__classcell__' in attrs:
+            new_attrs['__classcell__'] = attrs['__classcell__']
```
cculianu: 👍

PR #117, opened 2018-05-30
Refactor low-level and high-level parts of rubicon.objc.runtime into two modules

dgelessus:
This is just an internal refactor to organize some things a little better, there should be no changes to API or behavior (although code that uses certain non-public names might break). None of the currently open PRs should be affected either.

* Split the `rubicon.objc.runtime` module into two modules, `.runtime` and `.api`. The former contains the `ctypes` declarations and helper functions for low-level interfacing with the Objective-C runtime; the latter contains the high-level `ObjCInstance`, `ObjCClass`, etc. classes and related functions.
  * The reason why I'm doing this (aside from trying to make the code cleaner) is that I was working on documenting the `runtime` module, and I found that the module was quite large and contained a mix of high- and low-level parts. This would have made the reference docs for the module longer and more difficult to organize. With two separate modules, it's easier to document the low-level (`.runtime`) and high-level (`.api`) parts separately. It will also allow users to get a detailed explanation of the available high-level APIs, without having to worry about the low-level parts unless they need/want to.
* Updated the `__all__` list of `.runtime`, and removed some private things that don't really belong into `__all__`.
* Refactored a couple of functions in `.runtime` (see individual commits).
* Removed some unused private helper functions.


dgelessus:
The `.runtime` module doesn't necessarily have to be private, there might be cases where users would want to access the low-level runtime functions. Also, all of the `CDLL`s are now in `.runtime`, and there are definitely cases where users want access to those.

I would be okay with removing some of the `from .runtime import` names though. Aside from `SEL`, `send_message` and `send_super`, they are only needed for low-level runtime things, so it would make sense to not import them by default.

dgelessus:
OK, done.

I also made the `_load_or_error` function from `.runtime` public as `load_library`, so others can use it to load system frameworks reliably on iOS. And I refactored the unit tests a little to eliminate the copy-pasted code at the top of almost every test module. (Just refactoring lots of minor things that I come across.)

PR #118, opened 2018-05-31
Reference documentation

dgelessus:
Title says it all, I'm working on the reference section of the docs, and also updating/reformatting docstrings so I can pull them in using autodoc.

This isn't finished yet - so far I've done the `.types` module, but I'm planning to document the other submodules as well, and I'll probably also add an official list of names that are imported into `rubicon.objc`. I also want to try and integrate the reference docs with the existing tutorials and how-tos a little, by adding links (in both directions) between reference sections and the relevant guides.

Feedback is welcome - if you have any comments or suggestions about the structure, content, formatting, code style, whatever, let me know.


dgelessus:
Note to self: convert [this](https://gitter.im/pybee/general?at=5b550ee3f10ffa0b48a5a522) into a howto on calling C functions.

dgelessus:
I've finally managed to get some work done on this again. The howto on C functions is now finished, but I've pushed it to a different branch ([c_functions_howto](https://github.com/dgelessus/rubicon-objc/tree/c_functions_howto) in my fork), because it's fairly long and not directly related to the reference documentation. I'll make a PR for that once this one is merged (GitHub acts really strangely when one PR is based on another unmerged PR).

The reference documentation itself is basically finished (if I remember correctly - it's been a while since I had a proper look) and ready for review. I need to fix a few merge conflicts due to the changes on master from the last year and a half, but that shouldn't have any effect on the content.

dgelessus:
I've force-pushed away my previous "clean up all the docs" commit (dgelessus/rubicon-objc@e14de79d216a02a427d081de9781aa2ab879087d) - it caused *a lot* of merge conflicts, because a bunch of the cleanups have already been done in master anyway (just *slightly* differently). As a bonus, this also makes the diff for this PR easier to read.

After this PR is merged, I'll probably make a separate PR just to clean up the formatting in the docs - which is how I should've done it in the first place, rather than doing it as part of this PR.

rubicon/objc/api.py
```diff
@@ -525,12 +651,18 @@ def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
...
+        """Return ``self.description`` converted to a Python :class:`str`."""
+
         desc = self.description
         if desc is None:
             raise ValueError('{self.name}.description returned nil'.format(self=self))
         return str(desc)
 
     def __repr__(self):
+        """Get a debugging representation of ``self``, which includes the Objective-C object's address, class,
+        and ``deubgDescription``.
```
freakboy3742: typo: `deubg`

rubicon/objc/types.py
```diff
@@ -509,12 +554,8 @@ def compound_value_for_sequence(seq, tp):
 register_preferred_encoding(b'v', None)
 register_preferred_encoding(b'B', c_bool)
 
-# Note, we prefer to map b'c' to c_byte rather than c_char, because otherwise
-# ctypes converts the value into a one-character string which is generally
-# not what we want at all, especially when the b'c' represents a BOOL var.
 register_preferred_encoding(b'c', c_byte)
 register_preferred_encoding(b'C', c_ubyte)
-register_encoding(b'c', c_char)
```
freakboy3742: It's not clear why this is being removed...

rubicon/objc/types.py
```diff
@@ -568,16 +609,12 @@ def compound_value_for_sequence(seq, tp):
 
 register_preferred_encoding(b'^v', c_void_p)
 
-# Anonymous structs/unions with unknown fields can't be decoded meaningfully,
-# so we treat pointers to them like void pointers.
-register_encoding(b'^{?}', c_void_p)
-register_encoding(b'^(?)', c_void_p)
```
freakboy3742: Similarly, it's not clear why this is being removed...

docs/how-to/type-mapping.rst
```diff
@@ -119,45 +177,157 @@ string::
...
+    >>> md = objc.NSMutableDictionary.dictionaryWithDictionary({"one": 1, "two": 2})
+    >>> md["three"] = 3
+    >>> md
+    <rubicon.objc.collections.ObjCMutableDictInstance 0x10b8a7860: __NSDictionaryM at 0x7f86fb164b60: {
+        one = 1;
+        three = 3;
+        two = 2;
+    }>
+
+Mapping methods like ``keys`` and ``values`` are also supported:
```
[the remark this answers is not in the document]
dgelessus: > Do we want to mention that these create a copy of the key/value list, rather than a MemoryView?

I've added a note about this to the reference docs, because I think in most cases it's unlikely to be relevant in practice - you don't often call `keys`/`values` repeatedly, and most dictionaries will be small enough that there's no noticeable performance impact.

> To that end - the docs for str, list and dict handling talk about converting from Python to ObjC, but don't mention how to convert from ObjC to a pure Python str/list/dict should you need to

You're right, that's not explained very well in the howto or in the reference documentation - it's only explicitly pointed out for strings, because of the UTF-16 thing. I'm not sure how often it's really necessary to convert an Objective-C collection to a "real" Python one (since our wrapper classes already implement the Python collection APIs), but I agree that it should be mentioned for completeness. I'll expand both the how-to and reference docs to include that.

docs/reference/index.rst
```diff
@@ -3,6 +3,13 @@ Reference
...
+    :maxdepth: 1
+
+    rubicon-objc
+    rubicon-objc-api
+    rubicon-objc-runtime
+    rubicon-objc-types
 
 This is the technical reference for public APIs provided by Rubicon.
+
+Note that the :mod:`rubicon.objc` package also contains other submodules not documented here. These are for internal use only and not part of the public API; they may change at any time without notice.
```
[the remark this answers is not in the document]
dgelessus: They should; I think I just skipped writing the docs for `rubicon.objc.eventloop` because I really don't know much about `asyncio`. I've added a placeholder documentation page for the `.eventloop` module for now - it only lists the public classes/methods and pulls in their docstrings via autodoc where possible.

rubicon/objc/types.py
dgelessus: This isn't visible in the diff, but the same registration call is repeated a few lines below, so I removed this one because it's redundant. https://github.com/dgelessus/rubicon-objc/blob/d2886bdd22d58ca6bb49ca46b00722677f22d262/rubicon/objc/types.py#L580-L585

rubicon/objc/types.py
dgelessus: These are moved to a `with_encoding` annotation on `UnknownPointer`: https://github.com/beeware/rubicon-objc/pull/118/files#diff-a6b0ed51cad5a8aeefbf32642c5811adR648-R651

This is technically a functionality change that I snuck into the documentation PR, because while documenting I thought about this behavior, and decided that it makes more sense to map "unknown struct" to `UnknownPointer` rather than `c_void_p`.

dgelessus:
> The first is async, and making the call as to whether we want to make that a formal supported API.

(see [this review comment](https://github.com/beeware/rubicon-objc/pull/118#discussion_r368303353))

> The second is about usage of basic functionality. The docs provided are great API docs - but they're fairly cursory when it comes to the practicalities

This is mostly intentional - I think that sort of information is better placed in a how-to rather than in the reference documentation. In fact, one of the reasons why I started working on the reference docs is because I needed a place for all the deep technical details that *shouldn't* go into the how-tos. This is why the reference docs currently focus so much on these details and less on the practical use cases that most users care about :)

I agree though that the reference docs should have *some* additional information on how to actually use things like blocks in practice, in the style of what's already there for subclassing Objective-C classes.

dgelessus:
I've expanded the type conversion and blocks documentation a bit, and added a short example for `objc_property` to demonstrate where/how it's supposed to be used.

The block-related classes in particular weren't documented all that well before. It's been a while since I last looked at the block internals, so I had to re-study the source code a bit while writing the documentation. If you find anything in that part of the docs that is inconsistent or doesn't match your experience as a Rubicon user, let me know - it's possible that I misread some part of the source code.

PR #119, opened 2018-06-04
Enforces Python version 3.5 and above

danyeaw:
Signed-off-by: Dan Yeaw <dan@yeaw.me>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(
     author_email='russell@keith-magee.com',
     url='http://pybee.org/rubicon',
     packages=find_packages(exclude=['tests']),
+    python_requires='>=3.5',
     namespace_packages=['rubicon'],
     license='New BSD',
     classifiers=[
@@ -33,8 +34,9 @@ setup(
         'License :: OSI Approved :: BSD License',
         'Programming Language :: Objective C',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3 :: Only',
         'Topic :: Software Development',
     ],
--- a/tox.ini
+++ b/tox.ini
@@ -4,13 +4,13 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = {py34,py35,py36,pypy3}-{default,i386},flake8
+envlist = {py35,py36,py37,pypy3}-{default,i386},flake8
 
 [testenv]
 basepython = 
-    py34: python3.4
     py35: python3.5
     py36: python3.6
+    py37: python3.7
     pypy3: pypy3
 
 whitelist_externals =


setup.py
```diff
@@ -25,6 +25,7 @@
     author_email='russell@keith-magee.com',
     url='http://pybee.org/rubicon',
     packages=find_packages(exclude=['tests']),
+    python_requires='>=3.5',
```
dgelessus: I don't have any strong opinions on minimum Python versions, but is there any particular reason why we should require 3.5? AFAIK, the current minimum Python version for Rubicon is 3.4, and I'd be inclined to leave it at that unless there's a reason to drop 3.4. CPython 3.4 seems to still be supported, 3.4.8 was released a couple of months ago, on 2018-02-05.

If we decide to bump the minimum Python version, I think it would be better to go straight to 3.6, which has some very useful features like f-strings and ordered kwargs. It's probably too early to require the absolute latest version of Python though.

Adding `python_requires` is a good idea though, regardless of what minimum version we decide on, so users get an error when they try to `pip install` on an unsupported Python version.

setup.py
```diff
@@ -33,8 +34,9 @@
         'License :: OSI Approved :: BSD License',
         'Programming Language :: Objective C',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
```
dgelessus: Python 3.7 isn't released yet, the release is [scheduled for 2018-06-27](https://www.python.org/dev/peps/pep-0537/#schedule), so it may be too early to officially support it.

tox.ini
```diff
@@ -4,13 +4,13 @@
...
 [tox]
-envlist = {py34,py35,py36,pypy3}-{default,i386},flake8
+envlist = {py35,py36,py37,pypy3}-{default,i386},flake8
 
 [testenv]
 basepython = 
-    py34: python3.4
     py35: python3.5
     py36: python3.6
+    py37: python3.7
```
dgelessus: This requires some changes to the .travis.yml as well: Python 3.7 needs to be installed using `pyenv`, and the `tox` call needs to be updated to include the `py37` environment.

setup.py
freakboy3742: I don't think there's any specific motivation, other than the fact that 3.4 is now reaching end of life, and supporting less things is generally easier. 3.4 is in security-only mode now; when 3.4.8 was released, only sources were published. 

Rubicon is in a good position in that it doesn't need to use the newer language features, but is mature enough that porting across versions isn't that big a deal. Some of the other BeeWare projects (Toga, VOC and Batavia most notably) are a lot more sensitive to version changes, and dropping 3.4 specifically opens up some options (specifically, being able to use the `async`/`await` keywords). 

So - in the case of Rubicon, I'd tend to agree that it's worth keeping 3.4 support, at least until it is either (a) fully EOL'd, or (b) becomes a maintenance hurdle for some other reason.

PR #122, opened 2018-06-30
Fix loading of the Foundation framework under GNUstep

### LINKED ISSUE #113: How to run rubicon-objc on linux
I got the following error

> AttributeError: python: undefined symbol: PyType_stgdict

when I tried to import `rubicon.objc` on Ubuntu. Is there a way to use this library on Linux? If not currently, what do I need to do to make it work?

dgelessus:
Fixes #113.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,8 +60,8 @@ install:
       brew outdated pyenv || brew upgrade pyenv
     else
       pushd "$(pyenv root)"
-      git fetch
       git checkout master
+      git pull
       popd
     fi
   # The existing pyenv shims from the build cache may no longer be valid, for example if pyenv was upgraded
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 branches:
   only:
     - master
-os: osx
 matrix:
   include:
     # Note: The ordering of the environments is optimized so that the complete matrix runs as quickly as possible,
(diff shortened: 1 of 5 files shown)


dgelessus:
Done. I've also taken the opportunity to move the `flake8` check into a separate Travis environment, so that style errors won't cause all builds to fail anymore (possibly hiding OS-specific test failures).

dgelessus:
OK, for some reason `gcc` on Linux can't find the Foundation headers. I'm not sure why, they should be included in the `gnustep-devel` package.

The Linux tests are also failing because #114 isn't merged yet.

freakboy3742:
Ok; I've been wrestling with this all morning, and I have reached the limit of my patience :-)

I've worked out how to locate the GNUstep headers; but then *blocks* don't compile. 

You can add `-fblocks` to CFLAGS, which stops the *syntax* errors, but then you start getting an error about a header file that doesn't exist (`objc/blocks_runtime.h`). The problem appears to be related to an "ObjectiveC 2 runtime" that needs to exist - but I can't find any details on how to *install* this runtime.

Making matters worse, Travis has been *extremely* flaky all morning, taking 10-15 minutes to spool up a Linux box, locking up during builds, timeouts during builds, failures due to random network and permissions errors. On top of that, there appears to be *no* documentation for the Xenial base image, which is required to get Python 3.7 working on Ubuntu.

I'd love for Rubicon to support GNUStep; but at this point, I think we need someone who understands how GNUstep works to step up and fix the problems we're seeing.

dgelessus:
I've noticed that you switched the compiler to `clang` on all platforms - I'm not sure if that's a good idea. As far as I know, GCC is *the* standard compiler on Linux, while Clang is an optional alternative. It seems to me like GNUstep is meant to integrate with GCC by default - if you look at the [`gnustep-core-runtime`](https://packages.ubuntu.com/bionic/gnustep-base-runtime) package for example, you can see a GCC library listed in the dependencies. The [`libobjc4`](https://packages.ubuntu.com/bionic/libobjc4) runtime that is pulled in by default appears to be GCC-specific as well (it depends on GCC 8).

> The problem appears to be related to an "ObjectiveC 2 runtime" that needs to exist - but I can't find any details on how to install this runtime.

This *might* be related to the GCC vs. Clang thing... I saw that GNUstep has the repo https://github.com/gnustep/libobjc2 on GitHub, whose description says "Objective-C runtime library intended for use with Clang". This appears to be a *different* library than the one installed by default with GNUstep (libobjc4). So maybe the default GNUstep install provides a runtime library that works only with GCC and not Clang, and if you want to use Clang you need to install a different runtime library? I have no idea really, maybe I'm jumping to conclusions here.

> but at this point, I think we need someone who understands how GNUstep works to step up and fix the problems we're seeing

Agreed, that would be very helpful - I don't use Linux a lot, and have no experience with GNUstep aside from attempting to get Rubicon to work with it, so I keep running into issues that are probably quite obvious to someone used to developing on Linux.

freakboy3742:
@dgelessus I switched to clang because gcc didn't seem to work - the issue appears to be Objective C *syntax* that doesn't appear to be supported on GCC. I could be completely wrong on that front, though - again, GNUstep's documentation is it's own worst enemy on this front.

This points at the biggest concern I have with this patch. 

In theory, I'd love to be able to support GNUstep. However, given the state of GNUstep's documentation, I have absolutely no confidence that we (as in, you and I, as the primary maintainers) would be able to *maintain* GNUstep support once it was added. I don't want to be in a situation where progress on Rubicon that is needed for iOS/macOS support is being constrained because GNUstep's documentation allow us to work out how to support a new feature.

freakboy3742:
This PR has undergone significant bitrot, and I don't have any personal motivation to continue trying to maintain it. I've opened #176 to track the feature request; if someone is sufficiently motivated, they can re-start this porting effort.

PR #133, opened 2019-03-14
Travis CI: Add High Sierra and remove Yosemite and Python 3.4

cclauss:
Python 3.4 end of life is in two days https://devguide.python.org/#status-of-python-branches

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
     # xcode7.3: 5 minutes
     # xcode8.3: 3 minutes
     # xcode9.2: 2 minutes
+    # https://docs.travis-ci.com/user/reference/osx/#macos-version
     # This data is based on the following build, which should be similar to most builds on the repo:
     # https://travis-ci.org/pybee/rubicon-objc/builds/348992202
     # Currently, Travis allows a maximum of two concurrent Mac builds per owner:
@@ -21,21 +22,22 @@ matrix:
 
     # OSX 10.10 Yosemite
     - osx_image: xcode6.4
+    # macOS 10.13 High Sierra
+    - osx_image: xcode10.1
     # macOS 10.12 Sierra
     - osx_image: xcode9.2
     - osx_image: xcode8.3
     # OSX 10.11 El Capitan
-    - osx_image: xcode7.3
+    - osx_image: xcode8
 language: generic
 cache:
   directories:
     ~/.pyenv
 env:
   global:
-    - PY34="3.4.8"
-    - PY35="3.5.5"
-    - PY36="3.6.5"
-    - PY37="3.7.0"
+    - PY35="3.5.6"
+    - PY36="3.6.8"
+    - PY37="3.7.2"
 install:
   # The different Travis environments have different pyenv versions, and not all have the Python versions we need,
   # so we upgrade pyenv to get a consistent setup in all environments.
@@ -48,11 +50,10 @@ install:
   # pyenv's shims directory is not in the PATH by default.
   - export PATH="$(pyenv root)/shims:${PATH}"
   # These are the Python versions that we want to test on.
-  - pyenv install --skip-existing ${PY34}
   - pyenv install --skip-existing ${PY35}
   - pyenv install --skip-existing ${PY36}
   - pyenv install --skip-existing ${PY37}
-  - pyenv global ${PY37} ${PY36} ${PY35} ${PY34} system
+  - pyenv global ${PY37} ${PY36} ${PY35} system
   # tox is not installed by default in some older Travis environments.
   - python3.6 -m pip install tox
   - make -f Makefile
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,29 +7,26 @@ matrix:
     # Note: The ordering of the environments is optimized so that the complete matrix runs as quickly as possible,
     # while also providing the first results as early as possible.
     # At the moment, build times for each environment are roughly as follows:
-    # xcode6.4: 13 minutes
-    # xcode7.3: 5 minutes
-    # xcode8.3: 3 minutes
-    # xcode9.2: 2 minutes
+    # xcode7.3: 16 minutes
+    # xcode8.3: 7 minutes
+    # xcode9.2: 7 minutes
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -21,21 +22,22 @@ matrix:
...
     # OSX 10.10 Yosemite
     - osx_image: xcode6.4
+    # macOS 10.13 High Sierra
+    - osx_image: xcode10.1
     # macOS 10.12 Sierra
     - osx_image: xcode9.2
     - osx_image: xcode8.3
     # OSX 10.11 El Capitan
-    - osx_image: xcode7.3
+    - osx_image: xcode8
```
dgelessus: Huh. I originally removed our `xcode8` build because Travis announced that they would deprecate that image (2b568ae4decb0b342dcbcce49009b58cae5b9ad0). But apparently they un-deprecated `xcode8` and changed their blog post without any comment, so my commit message now looks like nonsense... (I'm not making this up - compare the blog post [on 2017-11-27](https://web.archive.org/web/20171127041023/https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce#retiring-some-older-images) and [on 2018-06-14](https://web.archive.org/web/20180614162535/https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce#retiring-some-older-images).)

In that case re-adding `xcode8` is not a bad idea, but I think we should also keep `xcode7.3`, so that we have at least one build per major Xcode version.

setup.cfg
```diff
@@ -14,7 +14,8 @@ max-line-length = 119
 # E133: closing bracket missing indentation
 # E226: missing whitespace around arithmetic operator
 # W503: line break occurred before a binary operator
-ignore = E133,E226,W503
+# W504: line break occurred after a binary operator
+ignore = E133,E226,W503,W504
```
dgelessus: Wouldn't it be better to reformat the code that triggers W504 rather than ignoring it? IIRC in Rubicon we usually put line breaks before the operators, which is why W503 is ignored. I'm not sure why flake8 is detecting these W504 issues only now - perhaps they changed something in an update.

tox.ini
```diff
@@ -26,11 +25,6 @@ setenv =
 
 [testenv:flake8]
 basepython = python3.6
-# FIXME Remove the manual pyflakes dependency once flake8 depends on pyflakes >= 2.0.0.
-# Currently we need to manually update pyflakes in order to use __class__, because pyflakes <= 1.6.0
-# doesn't understand __class__. This has been fixed in pyflakes 2.0.0, but flake8's requirements don't allow that
-# version of pyflakes yet.
 deps =
     flake8
-    pyflakes >= 2.0.0
```
dgelessus: :+1:

dgelessus:
BTW, I'm not sure why there are errors on Travis about Homebrew already running. I've deleted the Travis build cache for this PR and restarted all the builds, hopefully that fixes it.

setup.cfg
cclauss: Indeed the flake8 did change something in the update so that when you have add ___ignore___ argument the both w503 and w504 are enabled.

.travis.yml
```diff
@@ -7,35 +7,35 @@ matrix:
...
-    # xcode7.3: 5 minutes
-    # xcode8.3: 3 minutes
-    # xcode9.2: 2 minutes
+    # xcode7.3: 16 minutes
+    # xcode8.3: 7 minutes
+    # xcode9.2: 7 minutes
+    # xcode10.1: 6 minutes
+    # https://docs.travis-ci.com/user/reference/osx/#macos-version
     # This data is based on the following build, which should be similar to most builds on the repo:
     # https://travis-ci.org/pybee/rubicon-objc/builds/348992202
```
dgelessus: Could you change this build URL to the one from where you got the new run times?

.travis.yml
```diff
@@ -48,14 +48,13 @@ install:
...
   - export PATH="$(pyenv root)/shims:${PATH}"
   # These are the Python versions that we want to test on.
-  - pyenv install --skip-existing ${PY34}
   - pyenv install --skip-existing ${PY35}
   - pyenv install --skip-existing ${PY36}
   - pyenv install --skip-existing ${PY37}
-  - pyenv global ${PY37} ${PY36} ${PY35} ${PY34} system
+  - pyenv global ${PY37} ${PY36} ${PY35} system
   # tox is not installed by default in some older Travis environments.
   - python3.6 -m pip install tox
```
dgelessus: I think the manual installation of `tox` is no longer necessary now that we don't use the `xcode6.4` image anymore. On `xcode7.3` (now the oldest image we use), you can see that [`tox` is already installed and the `pip` call does nothing](https://travis-ci.org/pybee/rubicon-objc/jobs/506412490#L2202).

.travis.yml
```diff
@@ -48,14 +48,13 @@ install:
...
   - make -f Makefile
 script:
   - export DYLD_LIBRARY_PATH=`pwd`/tests/objc
-  - python3.6 -m tox -e "py{34,35,36,37}-default,flake8"
+  - python3.6 -m tox -e "py{35,36,37}-default,flake8"
```
dgelessus: Now that Python 3.7 is the newest installed version, should we change this call to use `python3.7`? (It probably doesn't make much of a difference in practice, since that's only the Python version used to run `tox` itself. The tests are run in separate virtualenvs with the proper Python versions no matter what.)

Alternatively we could remove the explicit Python call here altogether and let `tox` get called via the PATH (which *should* work because we call `pyenv global ...` earlier).

setup.cfg
cclauss: W504 fixes made and this change reverted.

.travis.yml
cclauss: xcode8 seems buggy for our purposes because it does not contain pipenv.

.travis.yml
cclauss: I did __pip install --upgrade tox__ to make sure we are running the latest and greatest.

.travis.yml
cclauss: Explicit is better than implicit.

PR #134, opened 2019-05-07
Fix image to use the new domain

jacebrowning:


danyeaw:
@jacebrowning Thanks! :heart:

PR #142, opened 2020-01-02
Fixes #141 - Add initial Github Actions CI config.

### LINKED ISSUE #141: Travis CI build is broken for most macOS versions
Example: https://travis-ci.org/beeware/rubicon-objc/builds/630497088

For all Travis environments except `xcode10.2` (the most recent one that our CI configuration includes),  the installation of Python 3.5 (using `pyenv`) fails with this message:

```
Last 10 log lines:
    import pip
  File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmprxyor7dg/pip-9.0.1-py2.py3-none-any.whl/pip/__init__.py", line 26, in <module>
  File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmprxyor7dg/pip-9.0.1-py2.py3-none-any.whl/pip/utils/__init__.py", line 27, in <module>
  File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmprxyor7dg/pip-9.0.1-py2.py3-none-any.whl/pip/_vendor/pkg_resources/__init__.py", line 35, in <module>
  File "/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/python-build.20191231013638.23409/Python-3.5.6/Lib/plistlib.py", line 65, in <module>
    from xml.parsers.expat import ParserCreate
  File "/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/python-build.20191231013638.23409/Python-3.5.6/Lib/xml/parsers/expat.py", line 4, in <module>
    from pyexpat import *
ImportError: No module named 'pyexpat'
make: *** [install] Error 1
```

I haven't looked into this in detail yet. Clearly the `pyexpat` module is missing, which likely means that it was not built during the CPython build. This usually happens when a required native library is not installed or cannot be found during build. I don't know what needs to be done to fix this - best case we only need to install the missing library with `brew`, but it might be more complicated than that (especially because all of these build environments were previously working).

I also don't know if this problem is specific to Python 3.5 or if it also affects later versions. Right now it's impossible to tell, because Python 3.5 is installed first, and the installation failure stops the entire build.

freakboy3742:
Travis is proving troublesome for some Python versions, and the mechanism for configuring Python has never been ideal. The rest of the BeeWare toolchain is switching to Github Actions as well, so it seems an opportune time to switch Rubicon as well.


freakboy3742:
I tried getting [other XCode versions in the build](https://github.com/beeware/rubicon-objc/runs/370613442); however, it appears that Github is fairly agressively pinning the macOS version, and while they do provide multiple Xcode versions, only Xcode 11+ is available.

freakboy3742:
@dgelessus Ready for your review. The build matrix is split into a 3.5 smoke build, followed by a matrix build for all other Python versions. This ensures that we have a basic expectation that the tests should pass before we expend the build resources on multiple platforms that we know will fail. The smoke build is on 3.5 because it's the one most likely to fail (usually due to accidentally introducing new syntactic elements, like f-strings, into the codebase). Beyond that, the build config should be relatively straightforward.

As noted in an earlier comment, it looks like cross-macOS builds won't be possible; Github is only exposing a single macOS version, and they aren't preserving old Xcode versions (it looks like they had 9.4 until a month or two ago, but they've since disappeared).

There's a separate configuration for "build status" - this does a repeat of the smoke build on what has actually been merged to master. This is what populates the "current build: passing" badge on the homepage.

There's also a publish and release workflow; this means we can cut a release and publish to PyPI by tagging the repository (at least, we can once this PR is merged and I add a PyPI access token to the Github config). I've added documentation on the full release process.

dgelessus:
Thank you for looking into this! I agree that it would be good to migrate to GitHub Actions for consistency with the rest of BeeWare, and the automatic PyPI release would be good to have as well.

However I think the lack of support for older macOSes on GH Actions makes it unsuitable as a proper CI solution for rubicon-objc. While I agree that there needs to be some limit to what macOSes we test on, I think testing on *only* the absolute latest macOS is not enough. The ObjC runtime has been fairly stable across macOSes so far, but I still think we should test on at least a couple of older versions to be safe - if only to check that we don't accidentally rely on APIs that were newly introduced in the newest macOS.

It's unfortunate that GH Actions doesn't even keep old Xcode versions installed, but for the purposes of testing rubicon-objc the Xcode version is not really relevant anyway. rubicon-objc itself operates purely at runtime, so the only thing that matters is the OS version; switching the Xcode version only changes compiler and SDK versions, which have no effect on rubicon-objc.

I'm wondering if it would be possible to use Travis in addition to GH Actions to cover the older macOSes? GH Actions could run style checks and the macOS 10.15 test build, and Travis could run the test builds for older versions. Of course this is only possible if we can get Travis working again, which we were trying to avoid...

I think I'll do some more testing with Travis to see if I can realistically get it working again. If that fails, I think we should still keep Travis, but with only the `xcode10.2` environment enabled (which is the only one that is currently working). That way we at least test on two macOS versions instead of just one.

freakboy3742:
Fair call - I agree dropping old macOS versions is less than ideal; I thought we could maybe live with it, but I can definitely see the other side of that coin.

A hybrid config is definitely possible. If we have both the TravisCI and Actions config, both workflows will trigger. We'll also be able to simplify the Travis config, as we will only need to test a single Python version cross OS versions.

/me dives in...

freakboy3742:
Closing in preference to #145.

PR #143, opened 2020-01-03
Fix Travis Python installation issues

### LINKED ISSUE #141: Travis CI build is broken for most macOS versions

dgelessus:
This removes the manual setting of the `CFLAGS` envvar, which is no longer needed and causes problems with installing Python using `pyenv` on older macOSes.

Fixes #141.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
     - osx_image: xcode7.3
     # macOS 10.13 High Sierra
     - osx_image: xcode10.1
-    # macOS 10.13 Mojave
+    # macOS 10.14 Mojave
     - osx_image: xcode10.2
     # macOS 10.12 Sierra
     - osx_image: xcode9.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,11 +50,9 @@ install:
   # pyenv's shims directory is not in the PATH by default.
   - export PATH="$(pyenv root)/shims:${PATH}"
   # These are the Python versions that we want to test on.
-  # The CFLAGS prefix is to set up the include path so that zlib is available to the Python build
-  # This is needed on Mojave, but doesn't hurt other versions.
-  - CFLAGS="-I$(xcrun --show-sdk-path)/usr/include" pyenv install --skip-existing ${PY35}
-  - CFLAGS="-I$(xcrun --show-sdk-path)/usr/include" pyenv install --skip-existing ${PY36}
-  - CFLAGS="-I$(xcrun --show-sdk-path)/usr/include" pyenv install --skip-existing ${PY37}
+  - pyenv install --skip-existing ${PY35}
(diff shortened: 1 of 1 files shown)


freakboy3742:
@dgelessus ... I take it you use an IDE to manage your git branch names? :-P

dgelessus:
How could you tell. :P Although with bash-completion and/or copy-paste, long branch names work fine in the terminal too.

dgelessus:
Wow, I think this is actually working?

dgelessus:
Personally I would prefer merging this PR first and then using that as a starting point for integrating GitHub Actions. That way we have our existing CI configuration working again in the meantime.

This would mean that you'd need to merge/rebase this change into #142 and #145. That should hopefully be easy to do, as this PR only changes four lines.

freakboy3742:
Sounds like a reasonable approach. Merge incoming!

PR #147, opened 2020-02-13
Add how-to guide for calling C functions and other non-ObjC APIs

dgelessus:
Based on [this discussion on Gitter](https://gitter.im/beeware/general?at=5b54e95357f4f664b794cde2). Originally I wrote this as part of #118, but I split it out into a separate PR because it's really a separate doc page, and because #118 was big enough already.

### DIFF
--- a/docs/how-to/c-functions.rst
+++ b/docs/how-to/c-functions.rst
@@ -89,8 +89,8 @@ Most real examples of C functions are more complicated than ``puts``, but the ba
 
 Each C library only needs to be imported/loaded once, and the ``restype`` and ``argtypes`` only need to be set once per function. This is usually done at module level near the beginning of the module, similar to Python imports.
 
-Inline functions: ``NSLocationInRange``
----------------------------------------
+Inline functions (e. g. ``NSLocationInRange``)
+----------------------------------------------
 
 Regular C functions can be called as explained above, but there is also a second kind of C function that needs to be handled differently: inline functions. Unlike regular C functions, inline functions cannot be called through a library object at runtime. Instead, their implementation is only provided as source code in a header file.
 
@@ -128,8 +128,8 @@ You can then put this translated function into your Python code and call it in p
         def NSLocationInRange(loc, range):
             return loc >= range.location and loc - range.location < range.length
 
-Global variables and constants: ``NSFoundationVersionNumber``
--------------------------------------------------------------
+Global variables and constants (e. g. ``NSFoundationVersionNumber``)
+--------------------------------------------------------------------
 
 Some C libraries expose not just functions, but also global variables. An example of this is the Foundation framework, which defines the global variable ``NSFoundationVersionNumber`` in ``<Foundation/NSObjCRuntime.h>``:
 
--- a/docs/how-to/c-functions.rst
+++ b/docs/how-to/c-functions.rst
@@ -63,7 +63,7 @@ Once you have opened the relevant header file in a text editor, you need to sear
 
     int puts(const char *);
 
-This means that ``puts`` returns an ``int`` and takes a single argument of type ``const char *`` (a C string). This translates to the following Python ``ctypes`` code:
+This means that ``puts`` returns an ``int`` and takes a single argument of type ``const char *`` (a pointer to one or more characters, i. e. a C string). This translates to the following Python ``ctypes`` code:
 
 .. code-block:: python
 
@@ -71,7 +71,9 @@ This means that ``puts`` returns an ``int`` and takes a single argument of type 
     >>> libc.puts.restype = c_int
     >>> libc.puts.argtypes = [c_char_p]
 
-Now that we have provided all of the necessary type information, we can call ``libc.puts``:
+Now that we have provided all of the necessary type information, we can call ``libc.puts``.
+
+For the ``c_char_p`` argument, we pass a byte string with the message we want to print out. ``ctypes`` automatically converts the byte string object to a ``c_char_p`` (``char *``) as the C funcion expects it. The string specifically needs to be a byte string (``bytes``), because C's ``char *`` strings are byte-based, unlike normal Python strings (``str``), which are Unicode-based.
 
 .. code-block:: python
 
--- a/docs/how-to/c-functions.rst
+++ b/docs/how-to/c-functions.rst
@@ -1,0 +1,375 @@
+=====================================
+Calling plain C functions from Python
+=====================================
+
+Most Objective-C APIs are exposed through Objective-C classes and methods, but some parts are implemented as plain C functions. You might also want to want to use a pure C library that provides no Objective-C interface at all. Calling C functions is quite different from calling Objective-C methods and requires some additional work, which will be explained in this how-to.
+
+.. seealso::
+
+    The `ctypes tutorial <https://docs.python.org/3/library/ctypes.html#ctypes-tutorial>`_ in the Python documentation, which explains how to call C functions in general (without a specific focus on Apple platforms and Objective-C).
+
+A simple example: ``puts``
+--------------------------
+
+We'll start with a simple example: calling the ``puts`` function from the C standard library. ``puts`` takes a C string and outputs it to standard output --- it's the C equivalent of a simple ``print`` call.
+
+Before we can call the function, we need to look it up first. To do this, we need to find and load the library in which the function is defined. In the case of standard C functions, this is the standard C library, ``libc``. Because this library is commonly used, Rubicon already loads it by default and exposes it in Python as :attr:`rubicon.objc.runtime.libc`.
+
+.. code-block:: python
+
+    >>> from rubicon.objc.runtime import libc
+    >>> libc
+    <CDLL '/usr/lib/libc.dylib', handle 7fff60d0cb90 at 0x105850b38>
+
+.. note::
+
+    For a list of all C libraries that Rubicon loads and exposes by default, see the :ref:`predefined-c-libraries` section of the :mod:`rubicon.objc.runtime` reference documentation.
+    
+    To access a library that is not predefined by Rubicon, you can use the :func:`~rubicon.objc.runtime.load_library` function:
+    
+    .. code-block:: python
+    
+        >>> from rubicon.objc.runtime import load_library
+        >>> libm = load_library("m")
+        >>> libm
+        <CDLL '/usr/lib/libm.dylib', handle 7fff60d0cb90 at 0x10596be10>
+
+C functions are accessed as attributes on their library:
+
+.. code-block:: python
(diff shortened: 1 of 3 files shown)


docs/how-to/c-functions.rst
```diff
@@ -0,0 +1,375 @@
...
+
+    >>> from ctypes import c_char_p, c_int
+    >>> libc.puts.restype = c_int
+    >>> libc.puts.argtypes = [c_char_p]
+
+Now that we have provided all of the necessary type information, we can call ``libc.puts``:
+
+.. code-block:: python
+
+    >>> libc.puts(b"Hello!")
```
freakboy3742: The fact that a C string is a Python byte string is worth a brief explanation. We probably don't want to get into the weeds of explaining how to do Unicode in C, other than "it's hard, turn back now before you go mad...". :-)

docs/how-to/c-functions.rst
```diff
@@ -0,0 +1,375 @@
...
+    Hello!
+    10
+
+.. note::
+
+    The ``10`` is the return value of ``puts``, which is printed out by the Python REPL; you can ignore it. (A non-negative return value from ``puts`` indicates that the call was successful.)
```
freakboy3742: Would this be clearer if you used `retval = libc.puts(b'Hello!')` as the example? That way you can clearly differentiate between the return value and the console output.

docs/how-to/c-functions.rst
```diff
@@ -0,0 +1,375 @@
...
+
+    If you're running this code from an editor or IDE and you see only the ``10`` and not the ``Hello!``, try running the code from a Python REPL in a terminal window instead. Some editors/IDEs, such as Python's IDLE, can only capture and display output produced by high-level Python functions (such as ``print``), but not output from low-level C functions (such as ``puts``).
+
+Most real examples of C functions are more complicated than ``puts``, but the basic procedure for calling them is the same: import or load the function's C library, set the function's return type and argument types based on the relevant header, and then call the function as needed.
+
+Each C library only needs to be imported/loaded once, and the ``restype`` and ``argtypes`` only need to be set once per function. This is usually done at module level near the beginning of the module, similar to Python imports.
+
+Inline functions: ``NSLocationInRange``
+---------------------------------------
```
freakboy3742: Might be worth rephrasing this as "An inline function example"; as written, this reads as if NSLocationInRange is the only inline function that is available.

docs/how-to/c-functions.rst
```diff
@@ -0,0 +1,375 @@
...
+
+    Python code translated from C like this is sometimes more complicated than necessary and can be simplified. In this case for example, ``True if cond else False`` can be simplified to just ``cond``, ``not (x < y)`` can be simplified to ``x >= y``, and a few redundant parentheses can be removed. A cleaner version of the translated code might look like this:
+
+    .. code-block:: python
+
+        def NSLocationInRange(loc, range):
+            return loc >= range.location and loc - range.location < range.length
+
+Global variables and constants: ``NSFoundationVersionNumber``
+-------------------------------------------------------------
```
freakboy3742: As with inline functions; would be worth reframing this as an example.

docs/how-to/c-functions.rst
dgelessus: Good idea - didn't think of that solution. The return value is basically irrelevant here, so hiding it is not a problem, and it makes the output much clearer.

dgelessus:
Thanks for the review! Those are all good points, they should all be fixed/improved now.

> (And the final example is 🤯)

Totally agree 😄 It was actually a bit challenging to explain all of the steps in an understandable way, because of all the layers of preprocessor magic that you have to work through by hand. Some parts also require previous knowledge about how (Objective-)C(++) headers work, mainly which function/type attributes can and cannot be ignored when calling from Python - I explained those parts a bit vaguely without going into much detail, because otherwise the example would probably be even less readable than it is now.

docs/how-to/c-functions.rst
```diff
@@ -63,34 +63,35 @@ Once you have opened the relevant header file in a text editor, you need to sear
 
     int puts(const char *);
 
-This means that ``puts`` returns an ``int`` and takes a single argument of type ``const char *`` (a C string). This translates to the following Python ``ctypes`` code:
+This means that ``puts`` returns an ``int`` and takes a single argument of type ``const char *`` (a pointer to one or more characters, i. e. a C string). This translates to the following Python ``ctypes`` code:
```
freakboy3742: Typo: space between in `i. e.`

docs/how-to/c-functions.rst
```diff
@@ -63,34 +63,35 @@ Once you have opened the relevant header file in a text editor, you need to sear
...
 .. code-block:: python
 
     >>> from ctypes import c_char_p, c_int
     >>> libc.puts.restype = c_int
     >>> libc.puts.argtypes = [c_char_p]
 
-Now that we have provided all of the necessary type information, we can call ``libc.puts``:
+Now that we have provided all of the necessary type information, we can call ``libc.puts``.
+
+For the ``c_char_p`` argument, we pass a byte string with the message we want to print out. ``ctypes`` automatically converts the byte string object to a ``c_char_p`` (``char *``) as the C funcion expects it. The string specifically needs to be a byte string (``bytes``), because C's ``char *`` strings are byte-based, unlike normal Python strings (``str``), which are Unicode-based.
```
freakboy3742: Typo: funcion

docs/how-to/c-functions.rst
```diff
@@ -63,34 +63,35 @@ Once you have opened the relevant header file in a text editor, you need to sear
...
-    If you're running this code from an editor or IDE and you see only the ``10`` and not the ``Hello!``, try running the code from a Python REPL in a terminal window instead. Some editors/IDEs, such as Python's IDLE, can only capture and display output produced by high-level Python functions (such as ``print``), but not output from low-level C functions (such as ``puts``).
+    The return value of ``puts`` is ignored in this example, because it isn't significant here.
 
 Most real examples of C functions are more complicated than ``puts``, but the basic procedure for calling them is the same: import or load the function's C library, set the function's return type and argument types based on the relevant header, and then call the function as needed.
 
 Each C library only needs to be imported/loaded once, and the ``restype`` and ``argtypes`` only need to be set once per function. This is usually done at module level near the beginning of the module, similar to Python imports.
 
-Inline functions: ``NSLocationInRange``
----------------------------------------
+Inline functions (e. g. ``NSLocationInRange``)
```
freakboy3742: typo: space between `e. g.`

docs/how-to/c-functions.rst
```diff
@@ -128,8 +129,8 @@ You can then put this translated function into your Python code and call it in p
         def NSLocationInRange(loc, range):
             return loc >= range.location and loc - range.location < range.length
 
-Global variables and constants: ``NSFoundationVersionNumber``
--------------------------------------------------------------
+Global variables and constants (e. g. ``NSFoundationVersionNumber``)
```
freakboy3742: typo: space between `e. g.`

docs/how-to/c-functions.rst
```diff
@@ -63,34 +63,35 @@ Once you have opened the relevant header file in a text editor, you need to sear
...
     Hello!
-    10
 
 .. note::
 
-    The ``10`` is the return value of ``puts``, which is printed out by the Python REPL; you can ignore it. (A non-negative return value from ``puts`` indicates that the call was successful.)
+    If you're running this code from an editor or IDE and don't see ``Hello!`` printed out, try running the code from a Python REPL in a terminal window instead. Some editors/IDEs, such as Python's IDLE, can only capture and display output produced by high-level Python functions (such as ``print``), but not output from low-level C functions (such as ``puts``).
 
-    If you're running this code from an editor or IDE and you see only the ``10`` and not the ``Hello!``, try running the code from a Python REPL in a terminal window instead. Some editors/IDEs, such as Python's IDLE, can only capture and display output produced by high-level Python functions (such as ``print``), but not output from low-level C functions (such as ``puts``).
+    The return value of ``puts`` is ignored in this example, because it isn't significant here.
```
freakboy3742: Failure to check success/failure return values is a common source of bugs in C code, so I don't think it's a good idea to brush over the return value as "not significant". This is doubly important because code that returns successfully with an error value *isn't* common Python practice. If we're providing a guide that will potentially empower people without C experience to invoke C methods, it seems appropriate to warn them about the land mines waiting for them. Granted, if `puts` returns a failure value, you've got *much* bigger problems to deal with, but it's a good a time as any to point out the design pattern that exists in C APIs.

docs/how-to/c-functions.rst
dgelessus: Yeah - in German you *have* to put a space there ([we literally have a standard for this](https://de.wiktionary.org/wiki/z._B.) - [DIN 5008](https://en.wikipedia.org/wiki/DIN_5008) 😛), so I tend to do it in English too by habit.

freakboy3742: Huh.. TIL. Sadly, English has no such central authority, so we have to fall back on documents like the [Chicago manual of style](https://www.amazon.com/gp/product/0226104206/ref=as_li_tl?ie=UTF8&camp=1789&creative=9325&creativeASIN=0226104206&linkCode=as2&tag=dwt0d9-20&linkId=798f352a8de37926fe606c37b89b4611), [AP Style Guide](https://www.amazon.com/gp/product/0465062946/ref=as_li_tl?ie=UTF8&camp=1789&creative=9325&creativeASIN=0465062946&linkCode=as2&tag=dwt0d9-20&linkId=0e9d37192ce4793e003e21a7d66f4167), all of which are proprietary documents you can't link to.

docs/how-to/c-functions.rst
dgelessus: On that topic, I'm actually not quite sure how much prior knowledge of C we should assume in this guide? I think for the most part I assumed that the reader is somewhat familiar with C, but not how to interact with C from Python - so they would know about C strings, error code return values, etc., but not how to best interact with those concepts from Python.

In any case, I think it's okay to go into a bit more detail here, even for readers who already know how error handling works in C - it's quite easy to forget these checks in the context of Python where you normally have exceptions.

freakboy3742: Agreed that we don't need this guide to become a tutorial on writing C, but based on the people I'm seeing come to BeeWare and Rubicon, there's a fair amount of "I have grand ideas but no experience with any language other than Python (and even my Python knowledge is limited)". 

To that end, I think it's worth pointing out the things that are likely to be issues, even if we don't provide a *deep* explanation of the details. That will provide a reminder for those who have C experience, and a starting point for further research for those that don't.

PR #154, opened 2020-04-04
Mock ctypes while building docs to fix autodoc in RTD build

### LINKED ISSUE #150: Sphinx autodoc fails on non-Mac systems because rubicon.objc does not import
This can be seen for example in [a recent Read the Docs build](https://readthedocs.org/api/v2/build/10449102.txt), which shows a lot of warnings like this one:

```
WARNING: autodoc: failed to import class 'objc.api.ObjCInstance' from module 'rubicon'; the following exception was raised:
  File "/home/docs/checkouts/readthedocs.org/user_builds/rubicon-objc/envs/latest/lib/python3.7/site-packages/sphinx/ext/autodoc/importer.py", line 154, in import_module
    __import__(modname)
  File "/home/docs/checkouts/readthedocs.org/user_builds/rubicon-objc/checkouts/latest/rubicon/objc/__init__.py", line 15, in <module>
    from . import runtime  # noqa: F401
  File "/home/docs/checkouts/readthedocs.org/user_builds/rubicon-objc/checkouts/latest/rubicon/objc/runtime.py", line 80, in <module>
    libobjc = load_library('objc')
  File "/home/docs/checkouts/readthedocs.org/user_builds/rubicon-objc/checkouts/latest/rubicon/objc/runtime.py", line 76, in load_library
    raise ValueError("Library {!r} not found".format(name))
ValueError: Library 'objc' not found
```

This is because Sphinx' autodoc feature (which I've used in many places in the new reference documentation) reads docstrings and other information from modules by importing them. Most parts of Rubicon can only be imported if a working Objective-C runtime is present. In the average documentation building environment, this isn't the case, so all autodoc imports fail and those parts of the documentation are omitted.

The usual solution to this problem seems to be mocking any dependencies that may be unavailable when building the docs. Sphinx even provides an [`autodoc_mock_imports` config option](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_mock_imports) to do this automatically for specific Python modules. This option won't be enough for Rubicon though - we are missing C libraries, not Python modules, so we probably need to mock them manually.

Some relevant links on the topic:

* [python - How to mock so that `from x import *` works - Stack Overflow](https://stackoverflow.com/q/11877571)
* [How to: Make your Sphinx documentation compile with ReadTheDocs when you’re using Numpy and Scipy « Robin's Blog](http://blog.rtwilson.com/how-to-make-your-sphinx-documentation-compile-with-readthedocs-when-youre-using-numpy-and-scipy/)
* [importerror - Sphinx: how to exclude imports in automodule? - Stack Overflow](https://stackoverflow.com/q/15889621)

dgelessus:
Fixes #150.

TLDR: This replaces ctypes with a mock version during the documentation build, which emulates enough of ctypes to make rubicon.objc import successfully and make autodoc work properly, without depending on the real ctypes module or an Objective-C runtime. For more details, see the commit message and comments in the code.

I don't think we have a way to properly test this before merging the PR, because Read the Docs only builds from the master branch and not PRs (unlike our CI). I did some local tests though to make sure that a documentation build using the mocked ctypes produces reasonable output (see commit message).


dgelessus:
> Would it be plausible to use _actual_ mocks (i.e., unittest.mock) rather than the complex partial implementation here?

Unfortunately not - or at least I haven't been able to figure out how. I tried both the built-in mocking feature of Sphinx autodoc ([`autodoc_mock_imports`](http://www.sphinx-doc.org/en/master/ext/autodoc.html#confval-autodoc_mock_imports)) and manually replacing the `ctypes` module with a [`unittest.mock.MagicMock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock), and both of them weren't able to mock all necessary ctypes features. In particular, they don't support subclassing mock objects, which happens when you try to define a `ctypes.Structure` type when `ctypes` is mocked. I was able to find a workaround for that which worked with `unittest.mock`, but then I ran into issues with `isinstance` checks which I wasn't able to fix.

> Does sphinx actually rely on a _working_ ctypes implementation?

Not Sphinx itself, but Sphinx autodoc needs to be able to import rubicon.objc to read its docstrings, and at import time rubicon.objc already needs to interact with ctypes and the Objective-C runtime a lot. For example it needs to detect whether the system is 32/64-bit and load core classes like `NSObject`. This is complex enough that you can't just replace `ctypes` with a generic mock object that always successfully does nothing. I tried that, and got errors because Rubicon wasn't able to figure out how big a pointer is, couldn't find `NSObject`, thought that every object was a block, etc.

I might be able to shorten the mock ctypes implementation by using `unittest.mock` with explicit configuration (instead of the default "every operation succeeds and returns another mock"). I don't think that's going to help much though - some of the logic needs to be implemented manually so that it behaves in a reasonable way (e. g. C functions need to return values of the correct type, and in some cases returning a null pointer won't work).

The alternative would be to drop autodoc entirely and move the contents of the docstrings into the RST files, so that the documentation build doesn't need to import any Rubicon code. The downside of this is that the docstrings are no longer in sync with the reference documentation, so we'd need to either manually keep them up-to-date (which is easy to forget), or replace them with links to the online reference documentation (which doesn't make for a good experience in the REPL and IDEs).

It would be really nice if Sphinx autodoc could extract docstrings by statically parsing the code instead of importing/executing it, then we wouldn't have this issue...

freakboy3742:
Ok; that's annoying, but it makes sense.

My follow up comments would be:
1. I agree that a recognition of provenance is needed, but is a full reproduction of the Python license required? There's a lot of.... noise in that file. I'd suggest we only need the PSF's license, along with a preface that describes what we've done (which is required by clause 3 of the PSF license). The older licenses, and the history of arguments with the FSF, are definitely necessary in *Python*'s context, but I'm not convinced *we* need to reproduce them.
2. Are you sure this actually fixes the problem? I added a docs-building phase to #159 (so we can catch docs markup errors as part of the build), but I got the same problem loading Foundation. I merged this branch into #159 as a fix, and it still fails loading Foundation.

dgelessus:
> 1. I agree that a recognition of provenance is needed, but is a full reproduction of the Python license required? There's a lot of.... noise in that file. I'd suggest we only need the PSF's license, along with a preface that describes what we've done (which is required by clause 3 of the PSF license). The older licenses, and the history of arguments with the FSF, are definitely necessary in _Python_'s context, but I'm not convinced _we_ need to reproduce them.

You're right, it's probably enough to include just the actual PSF license - especially considering that ctypes was only added in Python 2.5, which is long after the historical license changes, so the ctypes code is clearly under the PSF license only and not under any of the older licenses. python/cpython@d4c9320412177895f598a93d73a0e654db27c351

> 2. Are you sure this actually fixes the problem? I added a docs-building phase to #159 (so we can catch docs markup errors as part of the build), but I got the same problem loading Foundation. I merged this branch into #159 as a fix, and it still fails loading Foundation.

That looks like a working directory issue to me. I think my code assumed that you would always run the doc build from the docs directory - if you run it from somewhere else, the path to _mocked_modules will be resolved relative to the wrong directory, and Python won't find the replacement `ctypes` module. I'll see how I can fix that.

dgelessus:
> That looks like a working directory issue to me. I think my code assumed that you would always run the doc build from the docs directory - if you run it from somewhere else, the path to _mocked_modules will be resolved relative to the wrong directory, and Python won't find the replacement `ctypes` module. I'll see how I can fix that.

Actually, that can't be the issue. According to [the Sphinx documentation for `conf.py`](https://www.sphinx-doc.org/en/master/usage/configuration.html) (emphasis mine):

> The configuration file is executed as Python code at build time (using `execfile()`, and **with the current directory set to its containing directory**)

I'm now fairly sure it's because of some difference between `make -C docs html` and `python3 setup.py build_sphinx`. On my machine, the first command correctly used the mocked version of ctypes, but the second one didn't (most likely because with the second command ctypes was imported at some point before the conf.py is executed). This should be fixed now.

PR #158, opened 2020-04-07
Remove uses of DYLD_LIBRARY_PATH when running tests

dgelessus:
This is mainly because of [bpo-40198](https://bugs.python.org/issue40198) (I don't want to wait for the next Python bugfix releases or have to build Python by hand), but I think we're better off not using `DYLD_LIBRARY_PATH` no matter what. It seems that Apple doesn't really want anyone to use `DYLD_LIBRARY_PATH` outside of local development, and I wouldn't be surprised if they restrict its usage even further in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
         make
     - name: Test
       run: |
-        DYLD_LIBRARY_PATH=tests/objc python setup.py test
+        python setup.py test
 
   python-versions:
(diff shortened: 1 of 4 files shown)


dgelessus:
Hm, for some reason GitHub isn't showing the Travis CI build status for the PR? Travis is running the build for this PR, but you can only see it via the Travis website: https://travis-ci.org/github/beeware/rubicon-objc/builds/672151723

PR #159, opened 2020-04-10
Move to tox as a common entrypoint for all testing

freakboy3742:
After reflecting on the conversation on #145, I've changed my mind about Tox. I now agree that there is value in having a single point of configuration for "how to test", which is distinct from the CI configuration of "how to set up an environment in which you can test.". 

This PR:
 * Adds a tox configuration for packaging metadata checks
 * Adds a tox configuration for documentation formatting
 * Replaces Beefore with a call on the tox `flake8` entrypoint
 * Moves the separate packaging check to a prerelease check, executed via tox.
 * Uses tox as a way to configure how all forms of testing should be performed
 * Replaces the use of `python setup.py test` with `pytest`
 * Uses tox entry points in CI.


freakboy3742:
This includes a merge of #154 as at 636fdb4, in an attempt to get over the docs pre-test problem; however, that doesn't appear to resolve the issue.

tox.ini
```diff
@@ -4,20 +4,42 @@
...
 commands = flake8 {posargs}
+
+[testenv:packaging]
+basepython = python3.8
+deps =
+    wheel
+    twine
+commands =
+    {envpython} setup.py sdist bdist_wheel
+    {envpython} -m twine check dist/*
```
dgelessus, line 38: In my experience, using `{envpython}` to call Python seems to do more harm than good. `{envpython}` is a textual substitution that is performed by tox when it reads the config, which causes problems with argument splitting when the path to the project contains spaces. ([I have been bitten by this in the past.](https://github.com/dgelessus/python-rsrcfork/commit/007d15eb3d369ad8d0006e853e75b567e347998b))

As far as I can tell, it's much better to use plain `python` instead (and that's what every other project I've looked at does, [including tox itself](https://github.com/tox-dev/tox/blob/9d02cd4b7df111a7a69366ea5a498da3025ad234/tox.ini#L78)). All tox commands are run in a virtual environment, so `python` is guaranteed to refer to the correct Python executable, and there is no chance of any argument splitting problems, because `python` is a simple command and not a variable substitution.

tox.ini
```diff
@@ -4,20 +4,42 @@
...
+
+[testenv:docs]
+basepython = python3.8
+deps =
+    -r{toxinidir}/docs/requirements_docs.txt
+commands =
+    {envpython} setup.py build_sphinx -W
```
dgelessus, line 45: As above, it would be better to use `python` instead of `{envpython}`.

dgelessus: Is there a particular reason for using `python setup.py build_sphinx` instead of `cd docs; make html`, which is what our documentation recommends?

On that topic - perhaps we should adjust that documentation to recommend `tox -e docs` as the main way for building the docs. That way you don't have to worry about manually installing the documentation build dependencies or switching to the correct directory, because tox takes care of all of that.

.github/workflows/publish.yml
```diff
@@ -16,7 +16,8 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install --upgrade setuptools wheel twine
+        python -m pip install --upgrade setuptools
+        python -m pip install wheel twine
```
dgelessus, line 20: I'm wondering if we should move this (and the commands below) into a tox environment as well, for consistency? (It would also allow manually uploading a release more easily, although we really shouldn't have to do that anymore now that we have GitHub Actions set up to do that...)

.github/workflows/ci.yml
```diff
@@ -46,14 +40,15 @@ jobs:
...
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip setuptools
-        python -m pip install .
+        python -m pip install --upgrade pip
+        python -m pip install --upgrade setuptools
+        python -m pip install tox
     - name: Build test library
       run: |
         make
```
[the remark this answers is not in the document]
freakboy3742: 👍

.github/workflows/ci.yml
```diff
@@ -71,29 +66,13 @@ jobs:
...
     - name: Test
       run: |
         python setup.py test
```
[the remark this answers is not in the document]
freakboy3742: Wow... I *really* wrote this patch before the coffee kicked in, didn't I... :-)

tox.ini
```diff
@@ -4,20 +4,42 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py35,py36,py37,py38,pypy3,flake8
+envlist = flake8,packaging,docs,py3.5,py3.6,py3.7,py3.8,pypy3
```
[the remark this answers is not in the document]
freakboy3742: Good to know. I'll revert the introduction of the period, and adjust the build invocations in CI.

tox.ini
freakboy3742: I opted for `build_sphinx` because `cd docs && make` requires whitelisting `cd` and `make`. Admittedly, make is already whitelisted, but given there's a convenient entry point, I figured I might as well use it.

As for updating the docs about building docs - +1 to that.

.github/workflows/publish.yml
freakboy3742: Hrm... I completely agree that we *could*; my only hesitation would be making it easy to do something we shouldn't be doing. However, I agree there's value in having all config in one place, so I'll add this in.

freakboy3742: All this radical simplification does make me wonder if there is (and if not, why there isn't) a Github Action to do the "update pip, update setuptools, install tox, run tox" dance...

tox.ini
dgelessus: tox has a built-in [`changedir`](https://tox.readthedocs.io/en/latest/config.html#conf-changedir) option to run a specific environment in a different directory. `cd` probably wouldn't work as expected anyway - I think tox doesn't run all commands in the same shell.

Personally I always automatically avoid using `setup.py whatever` when there is an alternative, because the `setup.py` command always seems to call some legacy `distutils` code that does something you don't want. (For example `setup.py install` not playing nicely with pip so you should use `pip install`, and `setup.py upload` sending your password in plain text so you should use `twine upload`.) Then there's also the general move towards supporting build tools other than setuptools (with PEP 517/518, etc.), so it seems logical to avoid coupling to setuptools for no good reason, in case we ever want to switch to a different build/packaging tool.

Anyway, it probably doesn't matter much in this case - `setup.py build_sphinx` seems to be a fully supported way of calling Sphinx with no indication that it's deprecated or has odd behavior or anything. As long as we choose one way of calling Sphinx and stick to it consistently, I think either `setup.py build_sphinx` or `make` in the docs folder is okay.

docs/how-to/contribute-code.rst
```diff
@@ -34,16 +34,26 @@ to compile this library:
 
 This will produce ``tests/objc/librubiconharness.dylib``.
 
-You can then run the test suite:
+You can then run the full test suite:
```
dgelessus, line 37: Again, GitHub doesn't let me comment on the correct location. The paragraphs above this explain how to manually call `make` to build librubiconharness.dylib. I think we should remove that part entiely - now that `make` is called automatically by our tox config, there should be no need for the user to call it manually anymore.

docs/how-to/contribute-code.rst
```diff
@@ -34,16 +34,26 @@ to compile this library:
...
 .. code-block:: sh
 
     (venv) $ tox -e py37
 
+Or, to run using the version of Python that is in your current virtual
+environment:
+
+.. code-block:: sh
+
+    (venv) $ tox -e py
```
dgelessus, line 57: I think we should recommend `tox -e py` first and `tox -e py37` after that. Most of the time you don't care about the exact Python version used to run the tests, you just want them to be run only once instead of on all Python versions.

docs/how-to/contribute-docs.rst
```diff
@@ -20,18 +20,9 @@ the other index files for clues.
...
-    $ pip install -r requirements_rtd.txt
-
-Create the static files: ::
-
-    $ make html
-
-Check for any errors and,if possible, fix them.
-The output of the file should be in the ``_build/html`` folder.
-Open the file you changed in the browser.
+The output of the file should be in the ``_build/html`` folder. If there are
```
dgelessus, line 27: This is no longer correct when using `python setup.py build_sphinx` - the default build location in that case is build/html in the project root directory, rather than docs/_build/html.

tox.ini
```diff
@@ -4,20 +4,44 @@
...
+[testenv:package]
+basepython = python3.8
+deps =
+    wheel
+    twine
+commands =
+    python setup.py sdist bdist_wheel
+    python -m twine check dist/*
+
+[testenv:docs]
```
dgelessus, line 34: Minor thing: I would reorder the section here so that the `docs` environment isn't in between the `package` and `publish` environments - the last two environments are closely related, so it makes sense to have them defined together.

docs/how-to/contribute-docs.rst
freakboy3742: Looks like it's `build/sphinx/html` - but updated.

PR #162, opened 2020-04-12
Mark Python 3.8 as supported in rubicon.objc.ctypes_patch

dgelessus:


freakboy3742:
I guess this also means we're due for a new point release...

Issue #160, opened 2020-04-11
Travis CI build status not showing up for pushes or pull requests


dgelessus:
The Travis builds still run, and you can see their status if you manually go to https://travis-ci.org/github/beeware/rubicon-objc, but their status is not automatically shown on the relevant commits and PRs anymore.

This started happening some time between February and April. It doesn't seem to be caused by the additions of GitHub Actions, as the earlier commits in #145 display the status of both GitHub Actions and Travis CI.

Here is a commit that correctly displays both GitHub Actions and Travis CI statuses: https://github.com/beeware/rubicon-objc/pull/145/commits/c3c9407dd94975f7bbcf469061b42318fbfff122

This is the latest commit I can find (dated 2020-02-16) that displays the Travis CI build status: https://github.com/beeware/rubicon-objc/commit/18dbdf878ddfff65b9cbe4d189a1ac02c7bee7b8

Commits and PRs after that one no longer display a Travis CI status, for example: #155

After we merged #145, the GitHub Actions status started showing up on commits and PRs as expected, but the Travis CI status is still missing.

freakboy3742:
Looks like this might have been related to some changes at Travis' end. I turned it off and then on again in the TravisCI dashboard, and it looks like that has kicked whatever needed to be kicked; #171 and #172 are now showing Travis statuses.

PR #167, opened 2020-04-13
Fix tox on Travis being run using Python 2 instead of 3

dgelessus:
I think this was introduced by #159, and because of #160 we didn't notice that anything broke on Travis.


dgelessus:
Travis build for this PR: https://travis-ci.org/github/beeware/rubicon-objc/builds/674504555

PR #173, opened 2020-04-26
Added CI configuration checking manifest and newsfile.

freakboy3742:
It should not be possible to submit a PR without a changelog entry or any required MANIFEST.in updates. This introduces CI checks for those two properties.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ distribute-*
 env
 local
 .tox
-pip-wheel-metadata
 
 # OS X
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ distribute-*
 env
 local
 .tox
+pip-wheel-metadata
 
 # OS X
 .DS_Store
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = flake8,check-manifest,towncrier-check,docs,package,py{35,36,37,38},pypy3
+envlist = flake8,towncrier-check,docs,package,py{35,36,37,38},pypy3
 skip_missing_interpreters = true
 
 [testenv]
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -17,22 +18,38 @@ commands =
     pytest -vv
 
 [testenv:flake8]
+skip_install = True
```
dgelessus, line 21: Good idea!

tox.ini
```diff
@@ -17,22 +18,38 @@ commands =
...
 
-[testenv:docs]
+[testenv:check-manifest]
+skip_install = True
 deps =
-    -r{toxinidir}/docs/requirements_docs.txt
+   check_manifest
 commands =
-    python setup.py build_sphinx -W
+   check-manifest -v
```
dgelessus, line 31: I wonder if it would make sense to include this in `package`, because checking the MANIFEST.in is related to packaging. Or do we want to keep each tool in a separate tox environment, regardless of what "topic" they are related to?

tox.ini
```diff
@@ -17,22 +18,38 @@ commands =
...
+
+[testenv:towncrier-check]
+skip_install = True
+deps =
+    towncrier >= 18.5.0
```
dgelessus, line 36: Might be a good idea to use

```suggestion
    {[testenv:towncrier]deps}
```

to ensure that both envs use the same Towncrier dependency.

tox.ini
dgelessus: btw, there seems to be a space missing in the indentation here (it should be 4 spaces but is only 3 spaces).

freakboy3742: Including in package is a reasonable suggestion. I can't think of an obvious benefit to keeping the two separate, and there will be a speed benefit to merging them (one less VM to spin up), so I'll move it over.

tox.ini
freakboy3742: Oh - nice. Will do.

.gitignore
```diff
@@ -10,6 +10,7 @@ distribute-*
 env
 local
 .tox
+pip-wheel-metadata
```
dgelessus, line 13: This entry *should* not be necessary anymore - pip has been changed in pypa/pip#6325 to use a temporary directory instead of one in the project directory, and this change is included in pip 19.3 and later (see the [changelog](https://pip.pypa.io/en/stable/news/), under 19.3, under Bug Fixes).

freakboy3742: I added it because it suddenly appeared in my Rubicon folder after some testing; I'm guessing the environment I was in had an older version of pip. If it's not required on a modern pip, then I agree it can be removed.

PR #174, opened 2020-04-26
Tighten API of send_message

dgelessus:
This change is a little opinionated - it removes a couple of special features from `send_message` that, in my opinion, are either too error-prone or too rarely used to be worth the extra complexity. See the individual commits for explanations on why I think it would be best to remove each of these features.

I'm open to feedback on these changes. If you think any of these features are legitimately useful and removing them would be a bad idea, please let me know.

Also, this is obviously a breaking change - it removes features that were previously documented and supported. As we're still before version 1.0, this is technically fine. I don't think any of these features are widely used enough to require a proper deprecation process, but I'm open to feedback here as well.

### DIFF
--- a/changes/174.removal.rst
+++ b/changes/174.removal.rst
@@ -1,0 +1,4 @@
+Disallowed passing class names as :class:`str`/:class:`bytes` as the ``receiver`` argument of :func:`~rubicon.objc.runtime.send_message`.
+If you need to send a message to a class object (i. e. call a class method),
+use :class:`~rubicon.objc.api.ObjCClass` or :func:`~rubicon.objc.runtime.get_class` to look up the class,
+and pass the resulting :class:`~rubicon.objc.api.ObjCClass` or :class:`~rubicon.objc.runtime.Class` object as the receiver.
--- a/changes/174.removal.rst
+++ b/changes/174.removal.rst
@@ -1,4 +1,19 @@
-Disallowed passing class names as :class:`str`/:class:`bytes` as the ``receiver`` argument of :func:`~rubicon.objc.runtime.send_message`.
-If you need to send a message to a class object (i. e. call a class method),
-use :class:`~rubicon.objc.api.ObjCClass` or :func:`~rubicon.objc.runtime.get_class` to look up the class,
-and pass the resulting :class:`~rubicon.objc.api.ObjCClass` or :class:`~rubicon.objc.runtime.Class` object as the receiver.
+Tightened the API of :func:`~rubicon.objc.runtime.send_message`,
+removing some previously allowed shortcuts that were rarely used,
+or likely to be used by accident in an unsafe way.
+
+.. note::
+
+    In most cases,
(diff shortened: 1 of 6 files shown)


changes/174.removal.rst
```diff
@@ -0,0 +1,39 @@
...
+  However,
+  this feature was easy to misuse by accident,
+  as it allowed passing extra arguments to *any* method,
+  even though most Objective-C methods are not variadic.
+  Extra arguments passed this way were silently ignored without causing an error or a crash.
+
+  To prevent accidentally passing too many arguments like this,
+  the number of arguments now has to exactly match the number of ``argtypes``.
+  Variadic methods can still be called,
+  but the varargs now need to be passed as a list into the separate ``varargs`` keyword arugment.
```
freakboy3742, line 38: The ragged line-length formatting in this file won't affect rendered output, but it's a bit hard to consumer as text.

rubicon/objc/runtime.py
```diff
@@ -619,61 +619,69 @@ def should_use_fpret(restype):
...
-        or an Objective-C class name (as a :class:`str` or :class:`bytes`).
+    .. note::
+
+        Some Objective-C methods take variadic arguments (varargs), for example `+[NSString stringWithFormat:]
+        <https://developer.apple.com/documentation/foundation/nsstring/1497275-stringwithformat?language=objc>`_.
+        When using :func:`send_message`, variadic arguments are not passed as regular arguments in the ``args`` list,
+        but as a separate ``varargs`` parameter.
+
+        This protects against accidentally passing too many arguments into a method and having the extra arguments
+        treated as varargs, even though the method doesn't actually take varargs.
```
freakboy3742, line 635: This doesn't describe how types are handled for the variadic arguments. Reading between the lines, I'm guessing it's a bit of "throw it over the fence and hope it works" - but that's a function of Objective C, not something we're in any position to fix.

changes/174.removal.rst
dgelessus: Yeah - with documentation text I've gotten into the habit of using "LaTeX line wrapping", i. e. putting a line break after periods, commas, etc. instead of wrapping to a fixed column. This way you can add a few words to the start of the paragraph without having to rewrap the whole rest. It helps avoid merge conflicts when you have multiple people working on the same text, and keeps `git blame` cleaner. (Obviously this is more relevant for regular documentation than for changelog fragments.)

I can rewrap this to a fixed column width if you like.

rubicon/objc/runtime.py
dgelessus: Right, it would be good to explain that in more detail. `varargs` are converted by `ctypes` using its default type mappings (i. e. Python `int`s become C `int`s, Python `bytes` become C `char *`s, etc.), but I would recommend using explicit casts/constructors if at all possible, so that `ctypes` doesn't accidentally do a different conversion than you expected.

PR #177, opened 2020-05-02
Change ObjCMethod to use send_message instead of calling IMPs directly

dgelessus:
This makes all method calls go through `send_message`/`objc_msgSend`, rather than having some of them bypass that mechanism and call method `IMP`s directly. Aside from eliminating a redundant code path, this should improve compatibility with swizzling and other runtime method manipulation. It also seems to give slightly better performance (which makes sense - all compiler-generated Objective-C method calls go through `objc_msgSend`, so Apple will have optimized that function a lot).

### DIFF
--- a/docs/reference/rubicon-objc-api.rst
+++ b/docs/reference/rubicon-objc-api.rst
@@ -63,9 +63,14 @@ Objective-C classes
 
         The method's return type, as a :mod:`ctypes` type, or ``None`` for ``void`` methods.
 
-    .. attribute:: argtypes
+    .. attribute:: imp_argtypes
 
         The method's argument types, as a sequence of :mod:`ctypes` types. The types of the implicit ``self`` and ``_cmd`` parameters, :class:`~rubicon.objc.runtime.objc_id` and :class:`~rubicon.objc.runtime.SEL`, are included here.
+
+    .. attribute:: method_argtypes
+
+        The method's argument types, as a sequence of :mod:`ctypes` types. This is identical to :attr:`imp_argtypes`,
+        except that the types of the implicit ``self`` and ``_cmd`` parameters are not included.
 
 Standard Objective-C and Foundation classes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/reference/rubicon-objc-api.rst
+++ b/docs/reference/rubicon-objc-api.rst
@@ -47,31 +47,6 @@ Objective-C classes
 
 .. autoclass:: ObjCMetaClass(name_or_ptr)
 
-.. autoclass:: ObjCMethod
-
-    .. automethod:: __call__
-
-    .. attribute:: selector
-
-        The method's selector, as a :class:`~rubicon.objc.runtime.SEL`.
-
-    .. attribute:: encoding
-
-        The method's signature type encoding, as :class:`bytes`.
-
-    .. attribute:: restype
-
-        The method's return type, as a :mod:`ctypes` type, or ``None`` for ``void`` methods.
-
-    .. attribute:: imp_argtypes
-
-        The method's argument types, as a sequence of :mod:`ctypes` types. The types of the implicit ``self`` and ``_cmd`` parameters, :class:`~rubicon.objc.runtime.objc_id` and :class:`~rubicon.objc.runtime.SEL`, are included here.
-
-    .. attribute:: method_argtypes
-
-        The method's argument types, as a sequence of :mod:`ctypes` types. This is identical to :attr:`imp_argtypes`,
-        except that the types of the implicit ``self`` and ``_cmd`` parameters are not included.
-
 Standard Objective-C and Foundation classes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/docs/reference/rubicon-objc-api.rst
+++ b/docs/reference/rubicon-objc-api.rst
@@ -67,14 +67,6 @@ Objective-C classes
 
         The method's argument types, as a sequence of :mod:`ctypes` types. The types of the implicit ``self`` and ``_cmd`` parameters, :class:`~rubicon.objc.runtime.objc_id` and :class:`~rubicon.objc.runtime.SEL`, are included here.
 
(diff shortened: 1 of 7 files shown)


docs/reference/rubicon-objc-api.rst
```diff
@@ -63,17 +63,14 @@ Objective-C classes
 
         The method's return type, as a :mod:`ctypes` type, or ``None`` for ``void`` methods.
 
-    .. attribute:: argtypes
+    .. attribute:: imp_argtypes
```
freakboy3742, line 66: No argument that these changes are needed; however, we should be wary about changing interfaces, especially when they're documented. 

We're definitely "pre-1.0", so from that perspective I'm not *overly* concerned - getting the API *right* is more important than preserving strict backwards compatibility. However, we should probably start thinking about which of these attributes we consider to be part of the public API that we're committing to maintaining, and which are useful internals that we're documenting because they're useful, but we reserve the right to change if they get in the way.

rubicon/objc/runtime.py
```diff
@@ -671,7 +671,8 @@ def send_message(receiver, selector, *args, restype, argtypes, varargs=None):
             .format(tp=type(receiver))
         )
 
-    selector = SEL(selector)
+    if not isinstance(selector, SEL):
+        selector = SEL(selector)
```
freakboy3742, line 675: This sort of "universal coercion" makes me a little nervous; What's the use case for this change? Since we're doubling down on using `send_message`, is there any reason that we shouldn't go the other way here and enforce `if not isinstance(selector, SEL): raise TypeError()` (or similar)?

rubicon/objc/api.py
```diff
@@ -86,27 +86,14 @@ def __init__(self, method):
...
         self.encoding = libobjc.method_getTypeEncoding(method)
-        self.restype, *self.argtypes = ctypes_for_method_encoding(self.encoding)
-        self.imp = libobjc.method_getImplementation(method)
-        self.func = None
-
-    def get_prototype(self):
-        """Returns a ctypes CFUNCTYPE for the method."""
-        return CFUNCTYPE(self.restype, *self.argtypes)
+        self.restype, *self.imp_argtypes = ctypes_for_method_encoding(self.encoding)
+        assert self.imp_argtypes[:2] == [objc_id, SEL]
```
freakboy3742, line 91: assert statements are optimized to no-op if you turn on Python VM optimization, so they're generally only advisable as a "loose" check you can forget about in high-performance conditions. Should this be a hard check with a `TypeError()` on failure?

docs/reference/rubicon-objc-api.rst
dgelessus: Right, I thought about that too. In the case of `ObjCMethod` I don't think it's a big issue, because (quoting the docs):

> Currently, Rubicon doesn't provide any documented way to get an unbound `ObjCMethod` object for an instance method of an `ObjCClass`.

That is, the only documented/supported way to get an `ObjCMethod` object is by creating one directly from an Objective-C `Method` pointer. This isn't something that many people are likely to do, and [a quick GitHub search](https://github.com/search?q=rubicon+objcmethod&type=Code) shows no external uses of `ObjCMethod`.

If this was after the release of 1.0, I would add a property with the old name for compatibility, and have it raise a `DeprecationWarning` and redirect to the new property. But as we're still before 1.0 and this feature probably has no external users, I think it's okay to just rename the property with no compatibility handling.

dgelessus: On second thought, perhaps it would be a good idea to "un-document" `ObjCMethod` completely for now, as there's already almost no good way to use it. That would give us more freedom to work on its API, until we add proper introspection features to allow users to access a class's methods and such.

rubicon/objc/runtime.py
dgelessus: To clarify: the previous code meant that `send_message` only accepted `str` and `bytes` objects as the `selector` parameter, and passing a real `SEL` object would cause a `TypeError`, even though the documentation said that this was allowed (see 18b2ccdf9c6d9cf35bf9d8622237b1ea983eea21).

In general I agree that argument type mismatches should be an error and we shouldn't try to automatically "fix" the type like this. But in this case there is a good use cases for passing `str`/`bytes` as the `selector`, namely calling `send_message` manually (`send_message(NSString, "string", restype=objc_id, argtypes=[])`). Also, passing `str`/`bytes` as `selector` has been supported for a long time (and until now has been the *only* way to call `send_message`) and is done often in practice, so I don't think we should forbid that.

rubicon/objc/api.py
dgelessus: Hm, I'm not quite sure. I'm aware that `assert` is only meant for internal invariant checks and not for input validation. The question is which category this check falls into. There is generally no reason why `imp_argtypes[:2]` would ever be anything other than `[objc_id, SEL]` unless the Objective-C compiler or runtime has seriously messed up, so I would consider this more of an invariant check than input validation. On the other hand, we're interfacing with external code, so strictly speaking this is external input and needs to be validated properly.

I'm not strongly attached to using `assert` here - I'd be okay with replacing it with a proper check.

dgelessus:
I went ahead and removed `ObjCMethod` from the API documentation, as mentioned in https://github.com/beeware/rubicon-objc/pull/177#discussion_r419085577. I don't think there's much use in having a class in the public API when there's no good way to use it outside of internal Rubicon code. We can make it public again once it is actually useful to external users (and by that point we hopefully have its API figured out).

rubicon/objc/runtime.py
freakboy3742: Ah - I was missing that this was really about the historical str/bytes usage. Agreed that is something that is both (a) needed for backwards compatibility, and (b) makes sense to preserve as an API (since it's an obvious usage)

Following the logic down the chain, it's going to break because it won't be castable to `bytes` as part of constructing the SEL, so I guess there's not a whole lot of risk here.

rubicon/objc/api.py
freakboy3742: I tend to agree that this is more of an invariant check - I'm having difficulty thinking of a way that this would fail in practice. We can probably leave it as is.

PR #186, opened 2020-09-12
Fix detection of return/argument types by Block in a few cases

### LINKED ISSUE #153: ObjCBlock For bounded methods
I'm having trouble using a bounded method as a block, for example, running this

```
class A(object):

	def __init__(self):
		handler = objc.Block(self.handler)

	def handler(self) -> None:
		pass

A()
```
gives the following exception,
`ValueError: Block callables must be fully annotated or explicit argument types must be specified.`

Is there a way to annotate the "self" argument?

dgelessus:
Fixes #153 (both bugs described there).

## PR Checklist:


freakboy3742 (approved):
The change makes sense (and seems to include some nice conceptual cleanups around the signature discovery); plus it's got good test coverage. 👍

Issue #185, opened 2020-09-12
Replace/remove the DeallocationObserver mechanism


dgelessus:
Currently, every Objective-C object that is wrapped in `ObjCInstance` by Rubicon has a `DeallocationObserver` object associated with it. `DeallocationObserver` is a custom Objective-C class defined by Rubicon, which is used to notify Rubicon when the Objective-C object it's attached to is deallocated. When that happens, the object's `ObjCInstance` is removed from the global `ObjCInstance._cached_objects` cache.

This mechanism has two main disadvantages:

* It requires creating a new Objective-C object from Python whenever an Objective-C object is wrapped in `ObjCInstance` for the first time. This has a noticeable performance impact when creating many Objective-C objects from Python - see #183.
* It uses the Objective-C runtime's "associated objects" mechanism. There's nothing directly wrong with this - it's a public, documented, mature, non-deprecated API - but it is a relatively obscure feature of the runtime, and general advice is that [it should be considered a last-resort solution](https://nshipster.com/associated-objects/).
  * Also, the GNUStep Objective-C runtime doesn't support associated objects. GNUStep support isn't a very high priority and would also require some other changes, but the fact that Rubicon's core makes heavy use of associated objects is actually one of the major blockers for GNUStep compatibility - see #176.

The question is how we could replace this mechanism. I don't know of any better way to implement the *exact* behavior that `DeallocationObserver` currently provides, i. e. notifying Rubicon when an Objective-C object is deallocated. But there are other ways to implement what we use this mechanism for, namely removing no longer needed entries from the `ObjCInstance` cache. For example, we could replace the cache dictionary with a `weakref.WeakValueDictionary`, so that entries would automatically disappear as soon as the `ObjCInstance` in question is deallocated on the Python side.

A consequence of using a weak dictionary as the cache would be that it's no longer guaranteed that wrapping the same Objective-C object will *always* return the same `ObjCInstance`. It's only guaranteed that if an `ObjCInstance` exists for an Objective-C object, wrapping the same object again will return the same `ObjCInstance`. But once the last reference to that `ObjCInstance` (outside the cache) is deleted, the `ObjCInstance` is probably deallocated, so wrapping the same Objective-C object again in the future will probably return a new `ObjCInstance`.

[...]

Issue #183, opened 2020-08-26
Speed up instantiating new class instances


SamSchott:
**Is your feature request related to a problem? Please describe.**

Instantiating an ObjC class with rubicon-objc takes about 600 - 700 μs which is relatively slow compared to PyObjc with 10 - 15 μs. This can become problematic for instance when showing a table with 30 rows and 5 columns which are visible (see https://github.com/beeware/toga/issues/1030): each of the 150 cells is a `NSTableCellView` which has a `NSTextField` and a `NSImageView` and 6 layout constraints. Creating these 1350 class instances then takes almost once second which is a very noticeable lag for a "simple" table. A similar task in PyObjc takes about 16 ms.

**Describe the solution you'd like**

Speed up the creation of ObjCClass instances. I am not sure about how rubicon-objc handles creating class instances but can try to look through the code to see if there are obvious slowdowns.

dgelessus:
Hi, sorry for not replying earlier. Thanks for the report - in the past we haven't focused that much on performance in rubicon-objc, so it's quite likely that some parts of the code just aren't very well optimized. PyObjC will always have a slight performance advantage because it uses native code for many parts of the Objective-C interface (whereas rubicon-objc uses only `ctypes`), but the overhead of `ctypes` shouldn't be *that* big, and certainly doesn't explain why rubicon-objc is 60 times as slow as PyObjC.

For testing, I ran `for _ in range(1000): NSObject.new().autorelease()` with cProfile, to see if there were any obvious slow calls that happen often. Here are the most significant things I found:

* Whenever an Objective-C method returns an Objective-C object (rather than a primitive C value), the returned object is passed through `py_from_ns`, to convert the object into a more Python-friendly form. To determine if and how the object should be converted, `py_from_ns` checks if the object is an instance of certain classes that can be converted (like `NSString`, `NSArray`, etc.) using the Objective-C `isKindOfClass` method.
  
[...]

PR #187, opened 2020-09-19
Make ObjCInstance cache weak and avoid using DeallocationObserver

dgelessus:
This implements the change I described in #185.

This improves the performance of wrapping Objective-C objects in most cases, because most ObjCInstances never have custom Python attributes added and now no longer need to have a ``DeallocationObserver`` associated. Profiling some test code that creates objects in a loop, the execution time of the test code went down from about 180 - 200 ms to about 120 - 140 ms.

```python
import cProfile

from rubicon import objc

def to_profile():
    for _ in range(1000):
        objc.NSObject.new().autorelease()

cProfile.run("to_profile()", "output.pstat")
```

Linking #183 because this is a performance improvement.

## PR Checklist:

### DIFF
--- a/changes/185.misc.rst
+++ b/changes/185.misc.rst
@@ -1,0 +1,14 @@
+Changed the way :class:`~rubicon.objc.api.ObjCInstance`\s are cached.
+The custom ``DeallocationObserver`` class is no longer used to manage the cache
+for most objects - instead the cache has been made a weak value dictionary,
+meaning that entries are automatically removed if they are no longer referenced
+from Python.
+
+:class:`~rubicon.objc.api.ObjCInstance`\s with custom Python attributes are
+still managed using strong references and ``DeallocationObserver``, because
+these attributes are only stored on the Python side and not on the Objective-C
+object, so they could otherwise be lost if an object has only Objective-C
+references and no Python references. The ``DeallocationObserver`` for an
+:class:`~rubicon.objc.api.ObjCInstance` is now only created once it has a
+custom Python attribute added to it, which improves the performance of wrapping
+most other objects that don't use custom Python attributes.
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -2,6 +2,7 @@ import collections.abc
 import decimal
 import enum
 import inspect
+import weakref
 from ctypes import (
     CFUNCTYPE, POINTER, Array, Structure, Union, addressof, byref, c_bool, c_char_p, c_int, c_uint,
     c_uint8, c_ulong, c_void_p, cast, sizeof, string_at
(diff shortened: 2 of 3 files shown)


freakboy3742 (approved):
This is... disturbingly straightforward :-)

Can't find any fault with the implementation. The only tradeoff I can see is the one that you flagged on #185 - and that's easy to resolve by keeping a Python-side reference to the object (AFAICT, using `self.delegate` rather than `delegate` in your example would avoid the problem).

The only question I've got is whether we might want to expose a public API for this (effectively, a Python-side `retain()` analog). However, I think we can live without that for an initial release, and if it turns out that there is a real-world requirement, it's a relatively minor (and backwards compatible) change to add a public interface around `_ensure_keep_alive()`.

dgelessus:
Thanks for the review!

> The only tradeoff I can see is the one that you flagged on #185 - and that's easy to resolve by keeping a Python-side reference to the object (AFAICT, using `self.delegate` rather than `delegate` in your example would avoid the problem).

If you're referring to the Toga code snippet from https://github.com/beeware/rubicon-objc/issues/185#issuecomment-691481580, that should actually still work without modification. For now I tried to make this change as compatible as possible and avoid breaking existing code, which is why I didn't remove `DeallocationObserver` entirely and instead changed it to only be created when it's needed to make Python attributes work.

> The only question I've got is whether we might want to expose a public API for this (effectively, a Python-side `retain()` analog).

I don't think that's really necessary, because it's not very difficult to "retain" a Python object - you can just put it in a global variable, or if you need to "retain" many objects, make your own global list/dict of objects. And if this change is working as intended, it shouldn't be necessary to manually "retain" anything on the Python side, because Rubicon should do so automatically when it's really needed.

freakboy3742:
That was indeed the code snippet I was referring to. On my first inspection, I thought assigning `delegate` as an attribute of `self.native` wouldn't cause a reference to be created; but I now see that because `self.native.delegate` is a property on an `ObjCInstance`, that will cause the allocationobjec

freakboy3742:
That was indeed the code snippet I was referring to. On my first inspection, I wasn't sure if assigning `delegate` as an attribute of `self.native` would cause a reference to be created. `self.native.delegate` is a native property on an `ObjCInstance`, rather than a Python property, so it won't trigger keepalive. However, I'm guessing assigning `delegate.interface` will cause a keepalive to be created, which will keep the chain alive? Or have I misunderstood the lifecycle here?

Agreed on the Python-side `retain()` though.

dgelessus:
> `self.native.delegate` is a native property on an `ObjCInstance`, rather than a Python property, so it won't trigger keepalive. However, I'm guessing assigning `delegate.interface` will cause a keepalive to be created, which will keep the chain alive?

Exactly - the `delegate` is kept alive when `delegate.interface` is assigned, because starting at that point `delegate` has data that exists only on the Python object and not the Objective-C object. The assignment to `self.native.delegate` doesn't cause anything to be kept alive, because it's an Objective-C property and not a Python attribute.

Assigning an Objective-C property never needs to keep anything alive. If the value in question has custom Python attributes, then it has already been kept alive when the first Python attribute was added. If it doesn't have Python attributes, the `ObjCInstance` can be safely destroyed and recreated later if needed, because all important information is also stored on the Objective-C object.

freakboy3742:
That all makes sense. In which case, confirming my 🚢 approval from before :-)

PR #189, opened 2020-09-29
Refactor/simplify some parts of Python-defined ObjC class creation

dgelessus:
This is just an internal refactoring and shouldn't have any visible effect to users.

Instances of `objc_method`, `objc_property`, etc. have callback methods that are called by `ObjCClass` and `ObjCProtocol` when a new Objective-C class/protocol is created from Python. These callback methods contain the code that adds the method/property/etc. in question to the Objective-C class/property.

For class creation there were two callback methods, `pre_register` and `register`, which were called at different stages of the class creation process. This PR changes it so that only one callback method (`class_register`) is needed, which does the job of both `pre_register` and `register` at a single point during class creation. This makes the logic a bit easier to follow, and as a side effect some checks in `ObjCClass.__new__` are also no longer needed.

## PR Checklist:

### DIFF
--- a/changes/189.misc.rst
+++ b/changes/189.misc.rst
@@ -1,0 +1,1 @@
+Refactored some parts of the internal logic for defining Objective-C classes from Python.
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -273,7 +273,7 @@ class objc_method(object):
 
     def register(self, cls, attr):
         name = attr.replace("_", ":")
-        cls.imp_keep_alive_table[name] = add_method(cls, name, self, self.encoding)
+        add_method(cls, name, self, self.encoding)
 
     def protocol_register(self, proto, attr):
         name = attr.replace('_', ':')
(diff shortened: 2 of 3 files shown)


freakboy3742 (approved):
Looks like a solid cleanup to me.

The only thing that grates against me a *little* is the fact that the cache of IMPs is a global. To my mind, there is a certain elegance to having the keep alive list for methods on a class tied to that class. I can't spot an *obvious* reason why you would have made that change - but it wouldn't surprise me if there was a bootstrapping problem somewhere. 

I'm happy to give this patch an approval as is; but if it's possible to keep the state local to the class rather than global, I wouldn't complain if the patch was updated :-)

dgelessus:
The reason why I made the IMP keepalive table global is so that `class_register` can take just a low-level `Class` pointer and doesn't need a full `ObjCClass` object, which allows all registration methods to be called before the `ObjCClass` is created. It also simplifies all of the `add_method` calls, because the IMPs are automatically kept alive by `add_method`, so the caller doesn't need to take care of this anymore.

I agree that it isn't very nice to have a single global keepalive table for all IMPs, but realistically there's no point in keeping them stored on the class, because the table is never used for lookups and nothing is ever removed from it.

If anything, storing the keepalive table on the class could lead to problems with the IMPs being destroyed early: if the Python `ObjCClass` is no longer referenced, its keepalive table is destroyed, but the class and its methods could still be used from Objective-C.

freakboy3742:
That all sounds completely reasonable - so 🚢 !

PR #194, opened 2020-10-31
Unpin sphinx_tabs dependency in requirements_docs.txt

dgelessus:
This fixes the deprecation warning when building the docs.

Not sure why this was pinned in the first place when none of the other dependencies in the same file are. Possibly just a copy-paste artifact.

(I also did some other minor cleanups in requirements_docs.txt while I'm here)

## PR Checklist:

### DIFF
--- a/changes/194.misc.rst
+++ b/changes/194.misc.rst
@@ -1,0 +1,1 @@
+Updated ``sphinx_tabs`` to fix a deprecation warning when building the docs.
--- a/docs/requirements_docs.txt
+++ b/docs/requirements_docs.txt
@@ -3,4 +3,4 @@ sphinxcontrib-spelling
 pyenchant
 sphinx-autobuild
 sphinx_rtd_theme
-sphinx_tabs==1.1.8
+sphinx_tabs
--- a/docs/requirements_docs.txt
+++ b/docs/requirements_docs.txt
@@ -1,6 +1,6 @@
-sphinx
-sphinxcontrib-spelling
 pyenchant
+sphinx
 sphinx-autobuild
 sphinx-rtd-theme
 sphinx-tabs
+sphinxcontrib-spelling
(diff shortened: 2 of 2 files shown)


freakboy3742 (approved):
👍 

The sphinx-tabs version pin was almost certainly a copy-paste from another project.

freakboy3742:
Hrm... this [failed on the RTD build](https://readthedocs.org/projects/rubicon-objc/builds/12234291/)... some investigation required.

dgelessus:
> ```
> Running Sphinx v1.8.5
> ```

That's probably why - [sphinx-tabs 1.2.0](https://github.com/executablebooks/sphinx-tabs/blob/master/CHANGELOG.md#v120---2020-08-18) dropped support for Sphinx versions older than 2.

Does the RTD build *have* to use such an old Sphinx version, or can that be configured? The current version is Sphinx 3.2.1, and in my local testing, the rubicon-objc docs build fine with that version.

freakboy3742:
I don't think I've actively set anything to use an old version of sphinx. I'm wondering if perhaps being an "old" configuration that isn't using an in-repo configuration file is pinning sphinx to an old version by default...

dgelessus:
> I'm wondering if perhaps being an "old" configuration that isn't using an in-repo configuration file is pinning sphinx to an old version by default...

That's definitely possible. The RTD docs don't explicitly say anything about this (at least not anywhere I could find...), but [the configuration file docs](https://docs.readthedocs.io/en/latest/config-file/index.html) say that "Using a configuration file is the recommended way of using Read the Docs.". I wouldn't be surprised if the presence and version of the RTD config file would somehow affect which Sphinx version is used for the RTD build.

freakboy3742:
Ok - since RTD builds only trigger off master, I'm going to fiddle around with some direct-to-master commits to create a readthedocs.yml file and see if that fixes anything.

freakboy3742:
Ok; so the `readthedocs.yml` didn't change anything (other than silence the "you should be using readthedocs.yml" warning); but putting an explicit `sphinx > 2` version requirement in `requirements_docs.txt` looks like it's done the job.

dgelessus:
Nice, thank you for debugging that!

PR #195, opened 2020-10-31
Add macOS 11 (Big Sur) to GitHub Actions CI

dgelessus:
According to [the docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on), GitHub Actions already supports macOS Big Sur. Let's see how well rubicon-objc works there. (This will be the first time I test rubicon-objc on macOS 11 - I don't have it installed on any of my machines yet.)

For the moment I'm keeping our main Python compatibility test matrix on macOS 10.15 (as it is currently) and test only with Python 3.5 on macOS 11 (like our Travis-based CI does it for the older macOS versions).

## PR Checklist:
- [ ] All new features have been tested

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
     - name: Set up Python 3.5
       uses: actions/setup-python@v1
       with:
-        python-version: 3.5
+        python-version: 3.7
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,26 @@ jobs:
   smoke:
     name: Smoke test (3.5)
     needs: beefore
-    runs-on: macOS-latest
+    runs-on: macos-10.15
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.5
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.5
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


dgelessus:
Surprisingly, everything seems to be working fine on macOS 11!

.github/workflows/ci.yml
```diff
@@ -51,10 +51,29 @@ jobs:
...
+  macos-versions:
+    name: macOS compatibility test
+    needs: smoke
+    runs-on: macos-11.0
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.5
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
```
freakboy3742, line 63: There's a discrepancy between the commend on line 60, and the version actually in use here; is there a particular reason you've used Python 3.7 for the Big Sur tests?

freakboy3742 (changes requested):
Looks good (and I share your surprise that it works out of the box!).

One comment about the Python version in use.

.github/workflows/ci.yml
dgelessus: Oops, I forgot to update the version in the `name`. (It would be really nice if GH Actions could auto-generate useful names for actions based on their settings so this wouldn't happen...)

Python 3.7 is the earliest available version on macOS 11 on GitHub Actions - see d37e35d82c9322564242cf4ff63ff69e0dc8bc5b.

freakboy3742 (approved):
👍 

I've done a merge with master to make sure the RTD change hasn't broken anything in test; assuming that passes, I'll merge this in.

PR #196, opened 2020-10-31
Replace manual version extraction with setup.cfg attr directive

dgelessus:
This now works, because setuptools 46.4.0 and later extract attribute values statically where possible, rather than having to import the module in question at runtime. The latter fails in the case of rubicon-objc when building on a system that doesn't have a working Objective-C runtime.

Followup to https://github.com/beeware/rubicon-objc/pull/156#discussion_r405227677.

## PR Checklist:


freakboy3742 (approved):
🎉

PR #198, opened 2020-11-07
Reduce Travis CI matrix to minimum (closes #197)

### LINKED ISSUE #197: Travis CI no longer provides unlimited free builds
Travis CI has [announced a new pricing model](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) today, and among other things they are no longer providing unlimited free builds for open-source projects. Existing open-source users are switched to a free trial plan, which means that you get 10,000 credits *once* - which is equivalent to 200 minutes of build time on macOS. That isn't all that much when you consider that one of our Travis builds normally takes about 5-6 minutes (for all different macOS versions combined) and you usually have multiple builds for each PR. If you do the math, that gives us maybe 10 PRs worth of Travis builds? The blog post also says:

> * We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment.

I suppose we could try asking, but the way it's worded doesn't exactly sound like they have a lot of extra free capacity that they want to give away... And realistically, considering everything that's been happening with Travis CI recently, it's probably safer to move away completely to another CI service if possible. We already use GitHub Actions where possible, except for testing on old macOS versions (10.11 through 10.14) that only Travis CI provides - see #141 and #145.

Unfortunately it seems there are still no other CI services that provide a large number of older macOS versions like Travis does/did. GitHub Actions still only provides 10.15 (and recently 11.0), and both [AppVeyor](https://www.appveyor.com/docs/build-environment/#build-worker-images) and [Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software) provide 10.14 and 10.15. This means that we probably have to do away with CI for all older macOS versions. We could use AppVeyor or Azure Pipelines to replace the macOS 10.14 Travis build, though I don't know if that's worth the effort for one macOS version... We could also keep Travis around, but only running the oldest available macOS (10.11), to make the most of the remaining free credits.

dgelessus:
Instead of removing our Travis setup entirely, this reduces the matrix to a single build on the oldest available Xcode/macOS version (Xcode 7.3 on Mac OS X 10.11). This way we still have *some* CI testing on old macOS versions until the remainig free Travis CI credits run out (and at that point we can completely remove Travis CI).

As far as I know, Travis CI counts credits per owner (i. e. user or organization), so this will use up Travis CI credits that could theoretically also be used by other BeeWare projects. But as I understand it, almost all other BeeWare projects have moved off of Travis by now, so this shouldn't be a big issue.

Closes #197.

## PR Checklist:
- [ ] All new features have been tested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,7 @@ branches:
   only:
     - master
 os: osx
-matrix:
-  include:
-    # OSX 10.11 El Capitan
-    - osx_image: xcode7.3
-    # macOS 10.12 Sierra
-    - osx_image: xcode8.3
-    - osx_image: xcode9.2
-    # macOS 10.13 High Sierra
-    - osx_image: xcode10.1
-    # macOS 10.14 Mojave
-    - osx_image: xcode10.2
+osx_image: xcode7.3 # OSX 10.11 El Capitan
 language: generic
 cache:
   directories:
--- a/changes/197.removal.rst
+++ b/changes/197.removal.rst
@@ -1,0 +1,11 @@
+Removed macOS 10.12 through 10.14 from our automatic test matrix,
+due to pricing changes in one of our CI services (Travis CI).
+OS X 10.11 is still included in the test matrix for now,
+but will probably be removed relatively soon.
+Automatic tests on macOS 10.15 and 11.0 are unaffected
+as they run on a different CI service (GitHub Actions).
+
+Rubicon will continue to support macOS 10.14 and earlier on a best-effort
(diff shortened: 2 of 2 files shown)


freakboy3742 (approved):
👍 Seems like the best-effort approach we can use for now... at least until Travis decides to change the deal again...

Issue #48, opened 2017-07-08
Improve memory management documentation


dimaqq:
Looking at old SO discussion (e.g. https://stackoverflow.com/a/6610/705086 ) it looks like when `SomeOBJCClass.alloc().init()` is called from Python code, it becomes Python's responsibility either to `.release()` the resulting object on the spot, or to mark it for GC at the bottom of event loop using `.autorelease()`.

Is that correct?

If that's correct, it must be documented...

dimaqq:
Actually reading up on ARC, I'm completely confused... are calls to `.alloc()` recorded to current autorelease pool by the compiler or runtime?

Depending on that, Python must or must not call `.[auto]release()`...

dgelessus:
ARC is a compiler feature, it makes the compiler insert `retain`/`release` calls so the programmer doesn't have to. Because rubicon-objc operates at the runtime level, there is no way to use ARC. That means you have to insert `retain`/`release`/`autorelease` calls in the correct places, like you would without ARC.

Alternatively, it would be possible to make rubicon-objc call `retain` and `release` automatically. This is how Pythonista's `objc_util` module does it - it calls `retain` when an `ObjCInstance` is created, and `release` when it is garbage-collected (in `__del__`). The only issue with this is that it needs special cases for `alloc`/`copy`/`mutableCopy`/`new` methods, which return an already retained object ([docs](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmRules.html#//apple_ref/doc/uid/20000994-SW1)).

mhsmith:
> it would be possible to make rubicon-objc call `retain` and `release` automatically. This is how Pythonista's `objc_util` module does it - it calls `retain` when an `ObjCInstance` is created, and `release` when it is garbage-collected (in `__del__`

This was implemented in #201, but the documentation is still lacking:

* [rubicon-objc-api](https://rubicon-objc.readthedocs.io/en/latest/reference/rubicon-objc-api.html) says "Rubicon currently does not perform any automatic memory management on the Objective-C object wrapped in an [ObjCInstance](https://rubicon-objc.readthedocs.io/en/latest/reference/rubicon-objc-api.html#rubicon.objc.api.ObjCInstance)", which is no longer true.
[...]

PR #201, opened 2021-01-10
Autorelease objects which we own

### LINKED ISSUE #200: Releasing memory from deleted Objc classes
**Describe the bug**
I am not sure if this actually is a bug or intended behaviour. When deleting memory heavy objects on the Python side, the actual memory is not released / returned to the OS. I know that Python sometimes does not give back freed memory, especially in case of small blocks which may be reused later. But this seems different.

Is rubicon-objc hanging on to a reference of those objects somewhere? Or does ObjC just not release the memory when all Python references have disappeared?

**To Reproduce**
This can be reproduced by loading a 14 MB image into memory and deleting the Python reference to the object:

```Python
from rubicon.objc.runtime import load_library

appkit = load_library("AppKit")

NSImage = ObjCClass("NSImage")

# load a 14 MB image into memory
image = NSImage.alloc().initWithContentsOfFile("/System/Library/Desktop Pictures/Catalina Coast.heic")

del image  # memory is not released
```

**Expected behavior**
I would have expected at least part the 14 MB to be returned to the OS.

**Environment:**
 - Rubicon version: 0.4.1.dev1
 - Operating System: macOS 11
 - Python version: 3.9

SamSchott:
This PR aims to fix #200 and #48 by autoreleasing any obj-c instances which we own (created with a method which starts with "alloc", "copy", "mutableCopy", or "new").

The implementation goes as follows:

1. When an object is returned from a objc method call which qualifies, the object is marked with `_needs_release = True`.
2. Just before the Python instance is deleted, autorelease is called on the objc instance.

I've opted to use `autorelease` instead of `release` because the latter would result in occasional segfaults when the objc-instance is still needed while the Python instance has been destroyed. This is the case for some attributes of the cell views in a `toga.Table`.

Similarly, `autorelease` is not called directly when the instance is created because this would result in some Python variables pointing to a released objc instance.

@dgelessus, is this in line with what you envisioned? The workaround with the `_needs_release` is a bit of a hack but it seems necessary to me because `ObjCInstance` itself does not know which method created it. Any better ideas are welcome!

## PR Checklist:

### DIFF
--- a/docs/how-to/memory-management.rst
+++ b/docs/how-to/memory-management.rst
@@ -7,12 +7,11 @@ will automatically track where variables are referenced and free memory when
 the reference count drops to zero whereas Objective-C uses explicit reference
 counting to manage memory. The methods ``retain``, ``release`` and
 ``autorelease`` are used to increase and decrease the reference counts as
-described in the `Apple developer documentation`_. When enabling automatic
-reference counting (ARC), the appropriate calls for memory management will be
-inserted for you at compile-time. However, since Rubicon Objective-C operates
-at runtime, it cannot make use of ARC.
-
-.. _Apple developer documentation: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/MemoryMgmt.html
+described in the `Apple developer documentation
+<https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/MemoryMgmt.html>`__.
+When enabling automatic reference counting (ARC), the appropriate calls for
+memory management will be inserted for you at compile-time. However, since
+Rubicon Objective-C operates at runtime, it cannot make use of ARC.
 
 You won’t have to manage reference counts in Python, Rubicon Objective-C will
 do that work for you. It does so by tracking when you gain ownership of an
--- a/docs/how-to/memory-management.rst
+++ b/docs/how-to/memory-management.rst
@@ -1,5 +1,5 @@
 ===========================================
-Memory-management for Objective-C instances
+Memory management for Objective-C instances
 ===========================================
 
 Reference counting works differently in Objective-C compared to Python. Python
@@ -13,12 +13,12 @@ When enabling automatic reference counting (ARC), the appropriate calls for
 memory management will be inserted for you at compile-time. However, since
 Rubicon Objective-C operates at runtime, it cannot make use of ARC.
 
-You won’t have to manage reference counts in Python, Rubicon Objective-C will
+You won't have to manage reference counts in Python, Rubicon Objective-C will
 do that work for you. It does so by tracking when you gain ownership of an
 object. This is the case when you create an Objective-C instance using a method
-whose name begins with “alloc”, “new”, “copy”, or “mutableCopy” or explicitly
-call ``retain``. Rubicon Objet-C will then insert a ``release`` call when the
-Python variable that corresponds to the Objective-C instance is deallocated.
+whose name begins with "alloc", "new", "copy", or "mutableCopy". Rubicon
+Objective-C will then insert a ``release`` call when the Python variable that
+corresponds to the Objective-C instance is deallocated.
 
 You will only need to pay attention to reference counting in case of **weak
 references**. In Objective-C, creating a **weak reference** means that the
--- a/docs/how-to/memory-management.rst
+++ b/docs/how-to/memory-management.rst
@@ -21,7 +21,7 @@ whose name begins with “alloc”, “new”, “copy”, or “mutableCopy” 
 call ``retain``. Rubicon Objet-C will then insert a ``release`` call when the
 Python variable that corresponds to the Objective-C instance is deallocated.
 
-You will only need to pay attention to reference counting is in case of **weak
+You will only need to pay attention to reference counting in case of **weak
 references**. In Objective-C, creating a **weak reference** means that the
 reference count of the object is not incremented and the object will still be
 deallocated when no strong references remain. Any weak references to the object
--- a/docs/how-to/memory-management.rst
+++ b/docs/how-to/memory-management.rst
@@ -20,7 +20,11 @@ whose name begins with "alloc", "new", "copy", or "mutableCopy". Rubicon
 Objective-C will then insert a ``release`` call when the Python variable that
 corresponds to the Objective-C instance is deallocated.
 
-You will only need to pay attention to reference counting in case of **weak
+An exception to this is when you manually ``retain`` an object. Rubicon
+Objective-C will not keep track of such retain calls and you will be
+responsible to insert appropriate ``release`` calls yourself.
+
+You will also need to pay attention to reference counting in case of **weak
 references**. In Objective-C, creating a **weak reference** means that the
 reference count of the object is not incremented and the object will still be
 deallocated when no strong references remain. Any weak references to the object
--- a/docs/how-to/memory-management.rst
+++ b/docs/how-to/memory-management.rst
@@ -1,0 +1,50 @@
+===========================================
+Memory-management for Objective-C instances
+===========================================
+
+Reference counting works differently in Objective-C compared to Python. Python
+will automatically track where variables are referenced and free memory when
+the reference count drops to zero whereas Objective-C uses explicit reference
+counting to manage memory. The methods ``retain``, ``release`` and
+``autorelease`` are used to increase and decrease the reference counts as
+described in the `Apple developer documentation`_. When enabling automatic
+reference counting (ARC), the appropriate calls for memory management will be
+inserted for you at compile-time. However, since Rubicon Objective-C operates
+at runtime, it cannot make use of ARC.
+
+.. _Apple developer documentation: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/MemoryMgmt.html
+
+You won’t have to manage reference counts in Python, Rubicon Objective-C will
+do that work for you. It does so by tracking when you gain ownership of an
+object. This is the case when you create an Objective-C instance using a method
+whose name begins with “alloc”, “new”, “copy”, or “mutableCopy” or explicitly
+call ``retain``. Rubicon Objet-C will then insert a ``release`` call when the
+Python variable that corresponds to the Objective-C instance is deallocated.
+
+You will only need to pay attention to reference counting is in case of **weak
+references**. In Objective-C, creating a **weak reference** means that the
+reference count of the object is not incremented and the object will still be
+deallocated when no strong references remain. Any weak references to the object
+are then set to ``nil``.
+
+Some objects will store references to other objects as a weak reference. Such
+properties will be declared in the Apple developer documentation as
+"@property(weak)" or "@property(assign)". This is commonly the case for
+delegates. For example, in the code below, the ``NSOutlineView`` only stores a
+weak reference to the object which is assigned to its delegate property:
+
+.. code-block:: python
+
+    from rubicon.objc import NSObject, ObjCClass
+    from rubicon.objc.runtime import load_library
+
+    app_kit = load_library("AppKit")
+    NSOutlineView = ObjCClass("NSOutlineView")
+
+    outline_view = NSOutlineView.alloc().init()
+    delegate = NSObject.alloc().init()
+
+    outline_view.delegate = delegate
+
+You will need to keep a reference to the Python variable ``delegate`` so that
+the corresponding Objective-C instance does not get deallocated.
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -657,8 +657,8 @@ class ObjCInstance(object):
         return self
 
     def __del__(self):
-        # Release all objects which we own, except for autorelease pools.
-        if self._needs_release and self.objc_class.name != 'NSAutoreleasePool':
+
+        if self._needs_release:
             send_message(self, "release", restype=objc_id, argtypes=[])
 
     def __str__(self):
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -657,11 +657,9 @@ class ObjCInstance(object):
         return self
 
     def __del__(self):
-        # Autorelease all objects which we own, except for autorelease pools.
-        # We use autorelease instead of release here to prevent instances without a Python
-        # reference but with an Obj-C reference from being released too early.
+        # Release all objects which we own, except for autorelease pools.
         if self._needs_release and self.objc_class.name != 'NSAutoreleasePool':
-            send_message(self, "autorelease", restype=objc_id, argtypes=[])
+            send_message(self, "release", restype=objc_id, argtypes=[])
 
     def __str__(self):
         """Get a human-readable representation of ``self``.
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -656,6 +656,17 @@ class ObjCInstance(object):
 
         return self
 
+    def retain(self):
+        """
+        Manually increment the reference count of the corresponding objc object
+
+        The objc object is sent a dealloc message when its reference count reaches 0.
+        Use this method to retain objects which we don't own.
+        """
+        self._needs_release = True
+        result = send_message(self, "retain", restype=objc_id, argtypes=[])
+        return ObjCInstance(result)
+
     def release(self):
         """
         Manually decrement the reference count of the corresponding objc object
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -164,7 +164,7 @@ class ObjCMethod(object):
         # Convert result to python type if it is a instance or class pointer.
         if self.restype is not None and issubclass(self.restype, objc_id):
             result = ObjCInstance(result)
-            
+
         # Mark for release if we acquire ownership of an object. Do not autorelease here because
         # we might retain a Python reference while the Obj-C reference goes out of scope.
         if any(self.name.startswith(prefix) for prefix in (b'alloc', b'new', b'copy', b'mutableCopy')):
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -656,6 +656,17 @@ class ObjCInstance(object):
 
         return self
 
+    def release(self):
+        """
+        Manually decrement the reference count of the corresponding objc object
+
+        The objc object is sent a dealloc message when its reference count reaches 0. Calling
+        this method manually should not be necessary, the object will be automatically
+        released when the Python object is deallocated.
+        """
+        self._needs_release = False
+        send_message(self, "release", restype=objc_id, argtypes=[])
+
     def __del__(self):
         """
         Release the corresponding objc instance if we own it, i.e., if it was returned by
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -661,8 +661,9 @@ class ObjCInstance(object):
         Manually decrement the reference count of the corresponding objc object
 
         The objc object is sent a dealloc message when its reference count reaches 0. Calling
-        this method manually should not be necessary, the object will be automatically
-        released when the Python object is deallocated.
+        this method manually should not be necessary, unless the object was explicitly
+        ``retain``\\ed before. Objects returned from ``.alloc().init...(...)`` and similar calls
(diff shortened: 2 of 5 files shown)


SamSchott:
Oh, and it looks like the test suite is hitting a segfault somewhere, but I am note sure where. The toga examples all seemed to run just fine...

SamSchott:
A further update on the segfault: It always occurs in `test_core` but the actual method which triggers it is not reproducible. The failure is also specific to Python 3.5 and does not occur on Python 3.7 to 3.9 (all tested on macOS 11.1).

rubicon/objc/api.py
```diff
@@ -649,6 +656,13 @@ def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
 
         return self
 
+    def __del__(self):
+        # Autorelease all objects which we own, except for autorelease pools.
+        # We use autorelease instead of release here to prevent instances without a Python
+        # reference but with an Obj-C reference from being released too early.
+        if self._needs_release and self.objc_class.name != 'NSAutoreleasePool':
+            self.autorelease()
```
dgelessus, line 664: It might be a good idea to use `send_message` for this call instead of the normal high-level syntax, to avoid the overhead of Rubicon looking up the method first, and also saving an unnecessary wrapping of the returned object (which is just `self`). With any luck it could also change something about the segfault.

dgelessus (commented):
Thank you for working on this! The implementation looks logical and straightforward to me.

> The workaround with the `_needs_release` is a bit of a hack but it seems necessary to me because `ObjCInstance` itself does not know which method created it.

I wouldn't worry about that - I think an extra marker like this is unavoidable here, because as you say there's otherwise no way to tell if an object came from an `alloc`/etc. call and still has an extra reference that needs to be freed.

> I've opted to use `autorelease` instead of `release` because the latter would result in occasional segfaults when the objc-instance is still needed while the Python instance has been destroyed. This is the case for some attributes of the cell views in a `toga.Table`.

Hm, I'm not quite sure about this one. `autorelease` is definitely the safer option, but in most cases a normal `release` should work fine too. `autorelease` is (as far as I know) only *necessary* when writing a factory method that returns a newly allocated object, but *without* an extra reference like `alloc`/etc. For this to work, the returned object needs to be `autorelease`d and not `release`d, because otherwise it would already be deallocated by the time it's returned from the method.

An advantage of using `release` is that it makes reference management problems easier to detect and debug. If you `release` an object too early, it will be deallocated right away, and any code that still uses it will fail relatively quickly. However, if you `autorelease` too early, the object remains usable at first, and you'll notice only after the end of the current autorelease pool if the object is still incorrectly referenced anywhere.

Rubicon needs to properly support `autorelease` of course, but I think it would be better if *by default* `__del__` uses `release`, with the option to make it use `autorelease` instead for certain objects. Not sure how this would be best implemented API-wise.

> it looks like the test suite is hitting a segfault somewhere

> A further update on the segfault: It always occurs in `test_core` but the actual method which triggers it is not reproducible. The failure is also specific to Python 3.5 and does not occur on Python 3.7 to 3.9 (all tested on macOS 11.1).

Interesting, especially that it only happens on Python 3.5. I don't have any solid ideas here and haven't had the time to properly test this myself, so I'm mostly just guessing what *could* go wrong.

For the record, it can't have anything to do with processor architecture differences, because the crash also happens on GH Actions under macOS 10.15, and that macOS version only supports x86_64 (too new for i386 and too old for arm64).

A weird thing I ran across is how `__del__` interacts with weak references: when CPython deallocates a Python object with a `__del__` method, any weak references pointing to that object are only cleared *after* the `__del__` call has finished. This could cause issues when multiple threads use rubicon-objc at the same time. If `ObjCInstance.__del__` is called on one thread (which then calls `release`/`autorelease`) and another thread at the same time looks up that same object by its pointer (which retrieves the existing instance from `ObjCInstance._cached_objects`), the second thread would receive an `ObjCInstance` which was already released (or is currently being released or has been marked for autorelease) on the Objective-C side.

Another possibility is that there's a reference management bug somewhere in the rubicon-objc code that we simply didn't notice before, because newly `alloc`ed objects never had their original reference released. I don't have any concrete ideas where that might happen though - probably somewhere in the unit test code itself, because I'm not sure if the main rubicon.objc code ever `alloc`s anything.

This is one case where using `release` instead of `autorelease` in `__del__` could help with debugging 🙂

I also noticed that apparently these segfaults don't trigger Python's `faulthandler`, which we [explicitly enable in our unit tests](https://github.com/beeware/rubicon-objc/blob/8f72b2e01cbc4ac7f68543b5207882a3516443d0/tests/__init__.py#L17). I think this never happened to me before - `faulthandler` is usually quite reliable and helpful with crashes in native code. Maybe it would help to move the `faulthandler.enable()` call further up in the __init__.py, so that it gets enabled before `rubicon.objc` is imported?

Also, if you get a problem report popup from macOS when you reproduce the crash locally, can you copy the crash report and post it here (or as an attachment or Gist, if it's long)? Even though the native macOS crash report usually isn't as helpful as the Python `faulthandler` stack trace, it's better than having no information about the crash at all.

rubicon/objc/api.py
```diff
@@ -164,6 +164,12 @@ def __call__(self, receiver, *args, convert_args=True, convert_result=True):
         # Convert result to python type if it is a instance or class pointer.
         if self.restype is not None and issubclass(self.restype, objc_id):
             result = ObjCInstance(result)
+
+        # Mark for release if we acquire ownership of an object. Do not autorelease here because
+        # we might retain a Python reference while the Obj-C reference goes out of scope.
+        if any(self.name.startswith(prefix) for prefix in (b'alloc', b'new', b'copy', b'mutableCopy')):
```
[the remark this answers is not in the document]
SamSchott: Thanks! You live and learn :)

SamSchott:
Here is the macOS crash report:  [rubicon-objc-py35-crash.txt](https://github.com/beeware/rubicon-objc/files/5801835/rubicon-objc-py35-crash.txt) I try switching to `release` and enabling the fault handler earlier and report back if this gives any more information.

Regarding the use of `autorelease` instead of `release`: I was trying to avoid a very particular error which may be indeed be bad memory management on toga's side. Admittedly, this is code which I have written [here](https://github.com/beeware/toga/blob/0e3bfae0f049cf63b636c6b158a8fbac44938db3/src/cocoa/toga_cocoa/widgets/internal/cells.py#L32-L38). It creates a `NSImageView` instance and assigns it to a property of `NSTableCellView`. The image view never actually gets assigned to a Python variable. This therefore triggers `__del__` and `release` immediately after creating it and later results in errors similar to:

```
  File "/usr/local/lib/python3.9/site-packages/rubicon/objc/api.py", line 270, in __call__
    result = self.py_method(py_self, *args)
  File "/usr/local/lib/python3.9/site-packages/toga_cocoa/widgets/internal/cells.py", line 112, in setImage_
    self.imageView.image = None
AttributeError: 'NoneType' object has no attribute 'image'
```

The solution to this would be to always assign objc instances to Python variables and make sure they do not garbage collected. This could be a bit tedious.

SamSchott:
Alas, neither enabling the faulthander earlier, switching to `release`, or using `send_message` helped with identifying the cause of the segfault. The behaviour remains the same. I'll have to look at the actual test code more closely to see what might be going wrong. Hopefully some time this week.

dgelessus:
> Here is the macOS crash report

Welp, the stack trace in that crash report is not very helpful - I'm guessing the stack got destroyed before/during the crash.

> Regarding the use of `autorelease` instead of `release`: I was trying to avoid a very particular error which may indeed be due to bad memory management on toga's side. Admittedly, this is code which I have written myself [here](https://github.com/beeware/toga/blob/0e3bfae0f049cf63b636c6b158a8fbac44938db3/src/cocoa/toga_cocoa/widgets/internal/cells.py#L32-L39). It creates a `NSTextField` instance and assigns it to a property of `NSTableCellView`. The text field never actually gets assigned to a Python variable. This therefore triggers `__del__` and `release` immediately after creating it

The problem here is that `NSTableCellView`'s `imageView` and `textField` properties are declared as `assign`, which basically means that the property getters/setters don't do any reference management. This is actually quite unusual for object properties in (relatively) modern Objective-C code. Usually you would declare such properties as `weak` instead, to make it explicit that the object in the property is not retained, and to ensure safe behavior when the object gets deallocated while the property still points to it. From the user's point of view, `assign` and `weak` behave basically the same though - the important part is that such properties don't retain the objects they store, so the code that sets the properties needs to ensure that the objects are retained some other way.

I searched a bit, and according to [this SO question](https://stackoverflow.com/q/19464136), this problem with `NSTableCellView` happens even in pure Objective-C if you don't use Interface Builder (which seems to be the recommended way to set these properties) and aren't careful with your reference management. One comment there suggests:

> The key to make it work is adding text-view and image view as subviews of the `NSTableCellView`.

Toga does this already, but only *after* setting the properties, meaning that the objects are already deallocated by that point. This can probably be fixed as you say by storing the views in local variables, so they are kept retained during the entire method. Though to be safe, I would also add them as subviews as quickly as possible, to ensure that they are also retained that way.

SamSchott:
@dgelessus, thanks for the analyses of the issue with `NSTableCellView`. I should have read Apple's developer docs more carefully! In this case, there really is nothing stopping us from using `release` instead of `autorelease`.

SamSchott:
I think I have found the problem with the failing tests: the tests which would trigger a segfault are `test_objcclass_requires_class`, `test_objcprotocol_requires_protocol` and `test_objcmetaclass_requires_metaclass`. All three explicitly release an object, sometimes resulting in a segfault when that object is garbage collected.

Don't ask me why this only is an issue in Python 3.5, or why some of the methods still passed...

SamSchott:
Hm, it looks like the macOS compatibility test failed to start.

dgelessus:
> I think I have found the problem with the failing tests: the tests which would trigger a segfault [...] explicitly release an object

Ah, that explains it, nice catch! Lucky that this caused noticeable segfaults on Python 3.5 at least so we didn't miss this completely. Not sure why it doesn't crash on later Python versions - probably as you say because of some change to when/how CPython deallocates objects and runs the garbage collector.

On that note though - I wonder if it would be good to implement some compatibility handling so that old code like this (that correctly releases extra references) doesn't break because of this change. Maybe `ObjCInstance` should implement a custom `release` method (only in Python, not in Objective-C with `@objc_method`) that calls the Objective-C `release` method, but also sets `self._needs_release = False`, so that `__del__` won't release the object a second time. If I'm thinking correctly, this way the extra reference will be released correctly whether or not you explicitly call `release`.

> Hm, it looks like the macOS compatibility test failed to start.

Strange, not sure what's going on there. I see two runs for "CI / macOS compatibility test (pull_request)", one that failed (where  GH Actions just shows "This check failed") and one that ran normally and succeeded. I'm guessing it restarted the job automatically, but remembered the failed job and for some reason considers the entire CI failed as a result. I'm guessing it's just a temporary failure - I'll restart the CI jobs.

SamSchott:
> Maybe ObjCInstance should implement a custom release method (only in Python, not in Objective-C with @objc_method) that calls the Objective-C release method, but also sets self._needs_release = False, so that __del__ won't release the object a second time.

Yes, I think that's a good idea. It's also nice because it gives the user the option to explicitly release an object instead of implicitly by deleting all Python references (dangerous as this may be).

Another thing which is missing IMO is documentation. I'll try to familiarise myself with the rubicon-objc docs and see where a paragraph about memory management would fit it.

SamSchott:
Looking a the docs, the `rubicon.objc.api` reference would probably be the appropriate place for notes on memory management. However, this might be a bit buried. What do you think?

freakboy3742:
@SamSchott Nice work - I'll leave it to @dgelessus to give final approval, but this looks like a reasonable and easily explained solution to a non-trivial problem.

As for docs: the API docs are the right place to literally document that the `release` API exists; however, I'd be inclined to suggest that what we need is a new HowTo on memory management. It's a sufficiently important topic that it warrants a good explanation - especially given that someone from a Python background may not have a good handle on memory management in general, let alone the Objective C 'retain and release' model. It doesn't need to be an extensive essay on the topic - something that lays out the basic rules and common patterns of API usage should be sufficient.

dgelessus:
What @freakboy3742 said 🙂 At the moment we unfortunately don't have any proper docs about how to do memory management for Objective-C objects with Rubicon, so there's no obvious place in the docs where this new behavior could be documented. Agreed that we probably need a new short howto for Objective-C memory management, to explain when you do/don't have to call `retain` and `release` when using Rubicon.

re. the changes to the tests: I think it would be better to not explicitly do `del random_obj` and instead let the object be deleted normally at the end of the method. Almost all code that uses Rubicon will not explicitly `del` objects when it's done using them, so we should write the tests the same way.

Actually, the new custom `release` method should also have some dedicated tests, to ensure that code that explicitly calls `release` still works. Probably at least a test for releasing an object that came from `.alloc().init()` and one for retaining and then releasing an object that *didn't* come from `.alloc().init()`.

SamSchott:
Agreed on all points.

When designing the tests, I was wondering how we should handle `retain` calls. Should we also provide a custom `retain` method which sets `_needs_release = True` so that it will be automatically released for the user later? Or should we assume that when a user manually `retain`s an object, they'll also manually `release` it later? This would give them the option to retain an object beyond the lifecycle of the Python variable.

rubicon/objc/api.py
```diff
@@ -656,6 +656,17 @@ def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
...
 
+    def retain(self):
+        """
+        Manually increment the reference count of the corresponding objc object
+
+        The objc object is sent a dealloc message when its reference count reaches 0.
+        Use this method to retain objects which we don't own.
+        """
+        self._needs_release = True
+        result = send_message(self, "retain", restype=objc_id, argtypes=[])
```
dgelessus, line 667: This unconditionally calls `retain` on the Objective-C side every time `retain` is called from Python, but the boolean `_needs_release` attribute can only track one of those `retain` calls (or none at all, if the object came from `alloc`/etc. and so already has `_needs_release` set to `True`). This means that if you retain an object multiple times from Python, `__del__` will not release all of those references, so you instead need to manually call `release` for every `retain` call.

This could be solved by replacing `_needs_release` with an integer reference counter, so that multiple `retain` calls can be remembered and `__del__` can call `release` the correct number of times.

Though I'm wondering if it's necessary at all to have custom handling for when `retain` is called from Python. The custom handling in `release` and `autorelease` is needed so that code that calls these methods continues to work, even though we added an automatic `release` call in `__del__` for Python code that never does Objective-C reference management. However, existing code that calls `retain` already works correctly with just the changes to `release`/`autorelease`, because this PR doesn't add any automatic calls to `retain` that would need to be suppressed if the user code already calls `retain` manually.

rubicon/objc/api.py
```diff
@@ -667,6 +678,17 @@ def release(self):
...
 
+    def autorelease(self):
+        """
+        Decrements the receiver’s reference count at the end of the current autorelease pool block
+
+        The objc object is sent a dealloc message when its reference count reaches 0. If called,
+        the object will not be released when the Python object is deallocated.
+        """
+        self._needs_release = False
+        result = send_message(self, "autorelease", restype=objc_id, argtypes=[])
```
dgelessus, line 689: On the other hand, this is correct and necessary to avoid a second `release` if user code calls `autorelease`. Good catch - I forgot about that before.

docs/how-to/memory-management.rst
```diff
@@ -0,0 +1,50 @@
+===========================================
+Memory-management for Objective-C instances
```
dgelessus, line 2: Minor thing: "memory management" isn't hyphenated, as far as I know.

```suggestion
Memory management for Objective-C instances
```

dgelessus (commented):
The new how-to and the tests look good - my comments in the docs are just about minor formatting things. On the new Python-side `retain` method I've left a longer comment, because it currently doesn't handle some cases correctly and I'm wondering how that should be handled best.

docs/how-to/memory-management.rst
```diff
@@ -0,0 +1,50 @@
...
+will automatically track where variables are referenced and free memory when
+the reference count drops to zero whereas Objective-C uses explicit reference
+counting to manage memory. The methods ``retain``, ``release`` and
+``autorelease`` are used to increase and decrease the reference counts as
+described in the `Apple developer documentation`_. When enabling automatic
+reference counting (ARC), the appropriate calls for memory management will be
+inserted for you at compile-time. However, since Rubicon Objective-C operates
+at runtime, it cannot make use of ARC.
+
+.. _Apple developer documentation: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/MemoryMgmt.html
```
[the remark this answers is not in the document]
SamSchott: Sure- There are a few split links in the docs, such as in `contribute-code.rst`, but you are right: inline links are easier to maintain.

rubicon/objc/api.py
SamSchott: True, that's a fair point. Rather than manually tracking the reference count in Python I am also inclined to just drop the `release` method override.

docs/how-to/memory-management.rst
```diff
@@ -0,0 +1,50 @@
...
+
+You won’t have to manage reference counts in Python, Rubicon Objective-C will
```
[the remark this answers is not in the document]
SamSchott: Yes, that must have been the editor...

docs/how-to/memory-management.rst
```diff
@@ -0,0 +1,53 @@
...
+described in the `Apple developer documentation
+<https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/MemoryMgmt.html>`__.
+When enabling automatic reference counting (ARC), the appropriate calls for
+memory management will be inserted for you at compile-time. However, since
+Rubicon Objective-C operates at runtime, it cannot make use of ARC.
+
+You won't have to manage reference counts in Python, Rubicon Objective-C will
+do that work for you. It does so by tracking when you gain ownership of an
+object. This is the case when you create an Objective-C instance using a method
+whose name begins with “alloc”, “new”, “copy”, or “mutableCopy”. Rubicon
```
dgelessus, line 19: These quotes have remained curled:

```suggestion
whose name begins with "alloc", "new", "copy", or "mutableCopy". Rubicon
```

docs/how-to/memory-management.rst
```diff
@@ -0,0 +1,53 @@
...
+Objet-C will then insert a ``release`` call when the Python variable that
```
dgelessus, line 20: Small typo that I missed the first time:

```suggestion
Objective-C will then insert a ``release`` call when the Python variable that
```

rubicon/objc/api.py
```diff
@@ -649,6 +656,38 @@ def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
 
         return self
 
+    def release(self):
+        """
+        Manually decrement the reference count of the corresponding objc object
+
+        The objc object is sent a dealloc message when its reference count reaches 0. Calling
+        this method manually should not be necessary, the object will be automatically
+        released when the Python object is deallocated.
```
dgelessus, line 665: This should be updated now that an explicit `release` is necessary again when the object was explicitly `retain`ed before.

```suggestion
        The objc object is sent a dealloc message when its reference count reaches 0. Calling
        this method manually should not be necessary, unless the object was explicitly
        ``retain``\ed before. Objects returned from ``.alloc().init...(...)`` and similar calls
        are released automatically by Rubicon when the corresponding Python object is deallocated.
```

rubicon/objc/api.py
```diff
@@ -649,6 +656,38 @@ def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
...
+        """
+        self._needs_release = False
+        result = send_message(self, "autorelease", restype=objc_id, argtypes=[])
+        return ObjCInstance(result)
+
+    def __del__(self):
+        """
+        Release the corresponding objc instance if we own it, i.e., if it was returned by
+        by a method starting with 'alloc', 'new', 'copy', or 'mutableCopy' and it wasn't
+        already explicitly released by calling :meth:`release`.
```
dgelessus, line 685:
```suggestion
        already explicitly released by calling :meth:`release` or :meth:`autorelease`.
```

SamSchott:
Thank you @dgelessus for very carefully reading the documentation!

PR #203, opened 2021-01-27
Schedule callback for signal handler in CFRunLoop

### LINKED ISSUE #202: Event loop does not call registered signal handlers
**Describe the bug**
When using rubicon-objc's CFRunLoop integration, handlers registered with `loop.add_signal_handler()` are never invoked.

**To Reproduce**
Run the following Python code and send SIGINT to the process (Ctrl-C). The process keeps going.

```Python
import asyncio
import signal

asyncio.set_event_loop_policy(EventLoopPolicy())
loop = asyncio.get_event_loop()

future = loop.create_future()

def stop_loop():
    print("Shutting down")
    future.set_result(True)

loop.add_signal_handler(signal.SIGINT, stop_loop)
loop.run_until_complete(future)
```

When commenting out `asyncio.set_event_loop_policy(EventLoopPolicy())`, the process terminates as expected.

**Expected behavior**
Invoke the registered handler on SIGINT.

**Environment:**
 - Operating System: macOS 11.1
 - Python version: 3.8

SamSchott:
This PR fixes #202 by scheduling any callbacks on signals to run in the CFRunLoop. 

## PR Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
   # minutes on macOS 10.11.
   - pyenv install --skip-existing 3.5.2
   - pyenv global 3.5.2
-  - python -m pip install --upgrade pip
+  - python -m pip install --upgrade "pip<21"
   - python -m pip install --upgrade setuptools
   - python -m pip install tox
 script:
--- a/changes/202.bugfix.rst
+++ b/changes/202.bugfix.rst
@@ -1,0 +1,1 @@
+Fix calling of signal handlers added to the asyncio loop with CFRunLoop integration.
--- a/rubicon/objc/eventloop.py
+++ b/rubicon/objc/eventloop.py
@@ -540,6 +540,17 @@ class CFEventLoop(unix_events.SelectorEventLoop):
             args=args
         )
 
+    def _handle_signal(self, sig):
+        """Internal helper that is the actual signal handler."""
+
+        handle = self._signal_handlers.get(sig)
+        if handle is None:
+            return  # Assume it's some race condition.
+        if handle._cancelled:
+            self.remove_signal_handler(sig)  # Remove it properly.
+        else:
+            self.call_soon(handle._callback)
+
     def time(self):
         """Return the time according to the event loop's clock.
 
--- a/rubicon/objc/eventloop.py
+++ b/rubicon/objc/eventloop.py
@@ -540,17 +540,6 @@ class CFEventLoop(unix_events.SelectorEventLoop):
             args=args
         )
 
-    def _handle_signal(self, sig):
-        """Internal helper that is the actual signal handler."""
-
-        handle = self._signal_handlers.get(sig)
-        if handle is None:
-            return  # Assume it's some race condition.
-        if handle._cancelled:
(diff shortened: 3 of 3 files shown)


SamSchott:
And again we have the unexplained failure of the macOS compatibility test...

SamSchott:
@freakboy3742, do you think scheduling an immediately firing timer is the right approach?

If we'd want to mirror Python's `BaseEventLoop`, we'd probably want to add a new `CFHandle` class, inheriting from `events.Handle` instead of `events.TimerHandle` and using `CFRunLoopObserver` instead of `CFRunLoopTimer`.

I thought however that it may not be worth the effort to achieve essentially the same effect as an immediately firing and non-recurring timer.

SamSchott:
@freakboy3742, I know that you are currently swamped with PRs for toga but could you maybe having a look at this? It really only is a small change...

freakboy3742:
@SamSchott My apologies - this completely fell off my radar. I'll take a look right now...

freakboy3742 (approved):
Apologies for the slow response on this. 

The fix you've proposed definitely works; however, I think there's a better fix slightly higher up the stack using the same general approach. 

Your replica of `_handle_signal` is identical to the base class, except for replacing the call to `_add_callback_signalsafe` with the `call_soon` call. However, this leaves `_add_callback_signalsafe` - and `_add_callback` for that matter - in a broken state. 

Part of the problem is that we've inherited a lot of behavior from asyncio base classes (such as the `_write_to_self` handler that assumes it will cause `run_once()` will be invoked, clearing the handlers). `_add_callback_signalsafe` is only invoked (AFAICT) from `_handle_signal`; but since it's a method on the base event loop class, my assumption is that it *could* be invoked elsewhere in future. If that ever happens, fixing the problem at `_add_callback_signalsafe` means we'll be protected. Since the only difference between `_add_callback` and `_add_callaback_threadsafe` is the self-pipe waking call, that seems like a better place to address this problem.  I've made that change.

That leaves the decision of whether to use a 0-interval timer, or try to address this with an Observer. I had a quick poke to see how much work it would be - and while it might be nice, it's going to introduce a whole lot of complexity, and the only real improvement is that `call_soon` might become a little more efficient. Unless there's a particularly compelling performance reason, I'm inclined to leave it as timer-based for now.

SamSchott:
> My apologies - this completely fell off my radar.

No worries, I suspected as much.

>  I think there's a better fix slightly higher up the stack using the same general approach.

Good point. I'm not very familiar with asyncio and projects that make such heavy use of subclassing can be a bit confusing when first reading the code.

I've noticed that you require pip < 21 in the CI. Is there any particular reason for this?

freakboy3742:
The pip requirement was the source of the Travis build error. Pip dropped support for Python 3.5 in v21.0, and we still have Python 3.5 in our build matrix. 

Strangely, you can *install* pip 21 - but when you try to use it, it breaks. 

So - the version pin means we can install the most recent version of pip that is supported by the version of Python we're testing with.

We should possibly think about formally dropping Py3.5 support - which would mean we could remove the version pin, too.

PR #205, opened 2021-02-11
Allow restarting a stopped asyncio event loop.

### LINKED ISSUE #204: Using asyncio.run() fails with CFRunLoop integration
**Describe the bug**
When using rubicon-objc's custom `EventLoopPolicy` together with `asyncio.run()`, a runtime error is raised when the loop is shut down, specifically during the cleanup of tasks and async generators. This appears to be because `run_forever` is called a second time, triggering the lifecycle to be set again.

**To Reproduce**
The following code reproduces the error:

```Python
import asyncio

asyncio.set_event_loop_policy(EventLoopPolicy())

async def main():
    await asyncio.sleep(5)

asyncio.run(main())
``` 

This raises the following error:

```
ValueError                                Traceback (most recent call last)
<ipython-input-7-eaadc83a82ea> in <module>
----> 1 asyncio.run(main())

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py in run(main, debug)
     46         try:
     47             _cancel_all_tasks(loop)
---> 48             loop.run_until_complete(loop.shutdown_asyncgens())
     49             loop.run_until_complete(loop.shutdown_default_executor())
     50         finally:

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/rubicon/objc/eventloop.py in run_until_complete(self, future, **kw)
    434         future.add_done_callback(stop)
    435         try:
--> 436             self.run_forever(**kw)
    437         finally:
    438             future.remove_done_callback(stop)

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/rubicon/objc/eventloop.py in run_forever(self, lifecycle)
    445     def run_forever(self, lifecycle=None):
    446         """Run until stop() is called."""
--> 447         self._set_lifecycle(lifecycle if lifecycle else CFLifecycle(self._cfrunloop))
    448 
    449         if self.is_running():

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/rubicon/objc/eventloop.py in _set_lifecycle(self, lifecycle)
    591         """
    592         if self._lifecycle is not None:
--> 593             raise ValueError("Lifecycle is already set")
    594         if self.is_running():
    595             raise RuntimeError("You can't set a lifecycle on a loop that's already running.")

ValueError: Lifecycle is already set
```

**Additional context**
I did note that the example given the in the documentation explicitly uses `loop.run_forever()` instead of `asyncio.run()` to avoid those cleanup tasks. However, this requires a lot more manual cleanup which isn't always trivial.

On a side note: the asyncio integration could generally use some love. It's basics are functional but there probably remain a few methods and corner cases which don't work (see for instance #203). Maybe this something to work on before a big 1.0 release? Otherwise, the project seems quite close to a stable release.

SamSchott:
This PR fixes #204 by not setting the asyncio loop lifecycle again when it has already been set.

## PR Checklist:

### DIFF
--- a/changes/205.bugfix.rst
+++ b/changes/205.bugfix.rst
@@ -1,0 +1,1 @@
+Allow restarting a stopped event loop.
--- a/rubicon/objc/eventloop.py
+++ b/rubicon/objc/eventloop.py
@@ -444,7 +444,8 @@ class CFEventLoop(unix_events.SelectorEventLoop):
 
     def run_forever(self, lifecycle=None):
         """Run until stop() is called."""
-        self._set_lifecycle(lifecycle if lifecycle else CFLifecycle(self._cfrunloop))
+        if not self._lifecycle:
+            self._set_lifecycle(lifecycle if lifecycle else CFLifecycle(self._cfrunloop))
 
         if self.is_running():
             raise RuntimeError(
@@ -467,7 +468,8 @@ class CFEventLoop(unix_events.SelectorEventLoop):
         The implementation is effectively all the parts of a call to
         :meth:`run_forever()`, but without any of the shutdown/cleanup logic.
         """
-        self._set_lifecycle(lifecycle if lifecycle else CFLifecycle(self._cfrunloop))
+        if not self._lifecycle:
+            self._set_lifecycle(lifecycle if lifecycle else CFLifecycle(self._cfrunloop))
 
         if self.is_running():
             raise RuntimeError(


dgelessus:
Thank you for the PR! I'll have to ask @freakboy3742 to review this - I don't really have much experience with `asyncio`, and he's the one who wrote basically all of `rubicon.objc.eventloop` 🙂

freakboy3742 (approved):
👍 Looks good to me. Thanks for the fix (and apologies again for the delayed review)

PR #209, opened 2021-05-11
Replace deallocation observer

### LINKED ISSUE #185: Replace/remove the DeallocationObserver mechanism

SamSchott:
This PR closes #185 by replacing the deallocation observer mechanism. Its still work in progress but I am posting here because I'd like some feedback from @dgelessus on a couple of questions.

The mechanism for handling new attributes with Python instances as values is as follows:

1. When the Python instance `pyo` is assigned through `__setattr__`, a new `WrappedPyObject` is created to "wrap" that object. Practically, the `WrappedPyObject` manually increments the refcount of `pyo` and stores the address to `pyo` in an ivar.
2. This `WrappedPyObject` is associated with the original ObjCInstance through `objc_setAssociatedObject` which holds the only reference to it. The lifecycles of the ObjCInstance and the `WrappedPyObject` are therefore tied together.
3. When the ObjCInstance is deallocated, so is the `WrappedPyObject`. This will cause the refcount of `pyo` to be decremented in the overridden `WrappedPyObject.dealloc` method.

I've also updated the tests to ensure that all Python objects are indeed destroyed when they should be. This has been working quite well in my manual testing but I am still uncertain about a couple of implementation details. The questions are inline.

A larger question is regarding performance. Creating a new `WrappedPyObject` for each instance attribute can potentially be costly. But I don't really see an easy way around that. What do you think?

Edit: @freakboy3742 is of course also very welcome to comment!

## PR Checklist:

### DIFF
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -5,7 +5,7 @@ import inspect
 import weakref
 from ctypes import (
     CFUNCTYPE, POINTER, Array, Structure, Union, addressof, byref, c_bool, c_char_p, c_int, c_uint,
-    c_uint8, c_ulong, c_void_p, cast, sizeof, string_at
+    c_uint8, c_ulong, c_void_p, cast, sizeof, string_at, py_object,
 )
 
 from .types import (
@@ -570,14 +570,6 @@ class ObjCInstance(object):
     # and a new ObjCInstance might be created for it if it is wrapped again later.)
     _cached_objects = weakref.WeakValueDictionary()
 
-    # Some ObjCInstance objects must be kept alive for as long as their corresponding Objective-C object exists
-    # (see the _ensure_keep_alive method).
-    # Such objects are inserted into this regular (non-weak) dictionary,
-    # in addition to the weak _cached_objects dictionary.
-    # Entries are removed from this dictionary only once the corresponding Objective-C object is deallocated.
-    # This is done using the DeallocationObserver class.
-    _keep_alive_objects = {}
-
     @property
     def objc_class(self):
         """The Objective-C object's class, as an :class:`ObjCClass`."""
@@ -596,6 +588,10 @@ class ObjCInstance(object):
             # after it's been wrapped in an ObjCInstance.
             super(ObjCInstance, type(self)).__setattr__(self, "_objc_class", ObjCClass(libobjc.object_getClass(self)))
             return super(ObjCInstance, type(self)).__getattribute__(self, "_objc_class")
+
+    @staticmethod
+    def _associated_attr_key_for_name(name):
+        return SEL("rubicon.objc.py_attr.{}".format(name))
 
     def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
         """The constructor accepts an :class:`~rubicon.objc.runtime.objc_id` or anything that can be cast to one,
@@ -803,36 +799,20 @@ class ObjCInstance(object):
 
         if method:
             return ObjCBoundMethod(method, self)
-        else:
+
+        # Check if the attribute name corresponds to an instance attribute defined at
+        # runtime from Python. Return it if yes, raise an AttributeError otherwise.
+        key = self._associated_attr_key_for_name(name)
+        pyo_wrapper = libobjc.objc_getAssociatedObject(self, key)
+
+        if pyo_wrapper.value is None:
             raise AttributeError('%s.%s %s has no attribute %s' % (
                 type(self).__module__, type(self).__qualname__, self.objc_class.name, name)
             )
+        address = get_ivar(pyo_wrapper, "wrapped_pointer")
+        pyo = cast(address.value, py_object)
 
-    def _ensure_keep_alive(self) -> None:
-        """Ensure that this :class:`ObjCInstance` is kept alive as long as its Objective-C object exists."""
-
-        # Check first if this ObjCInstance is already kept alive,
-        # in which case we don't need to do anything.
-        if self.ptr.value not in ObjCInstance._keep_alive_objects:
-            ObjCInstance._keep_alive_objects[self.ptr.value] = self
-
-            # Create a DeallocationObserver and associate it with the Objective-C object.
-            # When the Objective-C object is deallocated, the observer will remove
-            # this ObjCInstance from the _keep_alive_objects dictionary.
-            # This will cause the ObjCInstance to be destroyed,
-            # assuming that there are no other references to it from Python
-            # (which shouldn't happen - if Python code has a reference to an ObjCInstance,
-            # it's expected to also retain a reference to the Objective-C object).
-            observer = send_message(
-                send_message(get_class('DeallocationObserver'), 'alloc', restype=objc_id, argtypes=[]),
-                'initWithObject:', self, restype=objc_id, argtypes=[objc_id]
-            )
-            libobjc.objc_setAssociatedObject(self, observer, observer, 0x301)
-            # The observer is now retained by the object we associate it to.
-            # We now release our own reference to the observer,
-            # so that the only remaining reference is released
-            # when the associated object is deallocated.
-            send_message(observer, 'release', restype=None, argtypes=[])
+        return pyo.value
 
     def __setattr__(self, name, value):
         """Allows modifying Objective-C properties using Python syntax.
@@ -852,16 +832,38 @@ class ObjCInstance(object):
                     value = value.value
                 ObjCBoundMethod(method, self)(value)
             else:
-                # Add a new Python attribute.
-                # These attributes are currently only stored on the Python ObjCInstance object
-                # and not on the actual Objective-C object.
-                # To ensure that the attributes aren't lost when the ObjCInstance has no more Python references
-                # (but the Objective-C object still has Objective-C references
-                # and may be wrapped in an ObjCInstance again in the future),
-                # every ObjCInstance with additional Python attributes is "kept alive" on the Python side.
-                # This prevents the ObjCInstance from being destroyed while the Objective-C object still exists.
-                self._ensure_keep_alive()
-                super(ObjCInstance, type(self)).__setattr__(self, name, value)
+
+                # Wrap the Python object in a WrappedPyObject instance.
+                # A reference will be retained as long as the WrappedPyObject is alive.
+
+                wrapper = send_message(
+                    send_message(get_class('WrappedPyObject'), 'alloc',
+                                 restype=objc_id, argtypes=[]),
+                    'initWithObjectId:', id(value), restype=objc_id, argtypes=[objc_id]
+                )
+
+                # Set the Python value as an associated object. This will release
+                # any previous wrapper object with the same key.
+                key = self._associated_attr_key_for_name(name)
+                libobjc.objc_setAssociatedObject(self, key, wrapper, 0x301)
+
+                # Release the wrapper object, it will be retained by the association.
+                send_message(wrapper, 'release', restype=objc_id, argtypes=[])
+
+    def __delattr__(self, name):
+        if name in self.__dict__:
+            # For attributes already in __dict__, use the default __delattr__.
+            super(ObjCInstance, type(self)).__delattr__(self, name)
+        else:
+            key = self._associated_attr_key_for_name(name)
+            # Check for instance attributes defined at runtime.
+            pyo_wrapper = libobjc.objc_getAssociatedObject(self, key)
+            if pyo_wrapper.value is None:
+                raise AttributeError('%s.%s %s has no attribute %s' % (
+                    type(self).__module__, type(self).__qualname__, self.objc_class.name, name)
+                )
+            # If set, clear the instance attribute / associated object.
+            libobjc.objc_setAssociatedObject(self, key, None, 0x301)
 
 
 # The inheritance order is important here.
@@ -1554,42 +1556,52 @@ class ObjCProtocol(ObjCInstance):
 NSObjectProtocol = ObjCProtocol('NSObject')
 
 
-# Instances of DeallocationObserver are associated with every
-# ObjCInstance that needs to be kept alive, which is the case
-# when the ObjCInstance has had additional Python attributes added to it.
-# Their sole purpose is to watch for when the Objective-C object
-# is deallocated, and then remove the object from the
-# ObjCInstance._keep_alive_objects dictionary.
+# When a Python object is assigned to a new ObjCInstance attribute, the Python
+# object should be kept alive for the lifetime of the ObjCInstance. This is
+# done by wrapping the Python object as a WrappedPyObject that increments
+# the reference count during assignment and decrements it when the WrappedPyObject
+# and the owning ObjCInstance are deallocated.
 #
 # The methods of the class defined below are decorated with
-# rawmethod() instead of method() because DeallocationObservers
+# rawmethod() instead of method() because WrappedPyObject
 # are created inside of ObjCInstance's __new__ method and we have
 # to be careful to not create another ObjCInstance here (which
 # happens when the usual method decorator turns the self argument
 # into an ObjCInstance), or else get trapped in an infinite recursion.
 
-# Try to reuse an existing DeallocationObserver class.
+# Try to reuse an existing WrappedPyObject class.
 # This allows reloading the module without having to restart
-# the interpreter, although any changes to DeallocationObserver
+# the interpreter, although any changes to WrappedPyObject
 # itself are only applied after a restart of course.
-try:
-    DeallocationObserver = ObjCClass("DeallocationObserver")
-except NameError:
-    class DeallocationObserver(NSObject):
 
-        observed_object = objc_ivar(objc_id)
+
+# Dictionary to keep references to wrapped Python objects. This prevents them
+# from being collected if there are otherwise only Objective-C references left
+# to the object.
+_keep_alive_objects = {}
+
+
+try:
+    WrappedPyObject = ObjCClass("WrappedPyObject")
+except NameError:
+    class WrappedPyObject(NSObject):
+
+        wrapped_pointer = objc_ivar(c_void_p)
 
         @objc_rawmethod
-        def initWithObject_(self, cmd, anObject):
+        def initWithObjectId_(self, cmd, address):
             self = send_message(self, 'init', restype=objc_id, argtypes=[])
             if self is not None:
-                set_ivar(self, 'observed_object', anObject)
+                pyo = cast(address, py_object)
+                _keep_alive_objects[(self.value, address.value)] = pyo.value
+                set_ivar(self, 'wrapped_pointer', address)
             return self.value
 
         @objc_rawmethod
         def dealloc(self, cmd) -> None:
-            anObject = get_ivar(self, 'observed_object')
-            ObjCInstance._keep_alive_objects.pop(anObject.value, None)
+            address = get_ivar(self, 'wrapped_pointer')
+            if address.value:
+                del _keep_alive_objects[(self.value, address.value)]
             send_super(__class__, self, 'dealloc', restype=None, argtypes=[])
 
         @objc_rawmethod
@@ -1598,8 +1610,9 @@ except NameError:
             # (which would have to be explicitly started with
             # objc_startCollectorThread(), so probably not too much reason
             # to have this here, but I guess it can't hurt.)
-            anObject = get_ivar(self, 'observed_object')
-            ObjCInstance._keep_alive_objects.pop(anObject.value, None)
+            address = get_ivar(self, 'wrapped_pointer')
+            if address.value:
+                del _keep_alive_objects[(self.value, address.value)]
             send_super(__class__, self, 'finalize', restype=None, argtypes=[])
 
 
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -5,6 +5,7 @@ import unittest
 from ctypes import Structure, byref, c_char, c_double, c_float, c_int, c_void_p, cast, create_string_buffer
 from decimal import Decimal
 from enum import Enum
+import weakref
 
 from rubicon.objc import (
     SEL, NSEdgeInsets, NSEdgeInsetsMake, NSMakeRect, NSObject, NSObjectProtocol, NSRange, NSRect, NSSize, NSUInteger,
@@ -14,6 +15,9 @@ from rubicon.objc import (
 from rubicon.objc.runtime import get_ivar, libobjc, objc_id, set_ivar
 
 from . import OSX_VERSION, rubiconharness
+
+
+NSAutoreleasePool = ObjCClass("NSAutoreleasePool")
 
 
 class struct_int_sized(Structure):
@@ -1163,6 +1167,13 @@ class RubiconTest(unittest.TestCase):
         self.assertEqual(thing.python_object_1, range(2, 8))
         self.assertEqual(thing.python_object_2, type)
 
+        # Test deleting Python attribute.
+
+        del thing.python_object_1
+
+        with self.assertRaises(AttributeError):
+            thing.python_object_1
+
     def test_objcinstance_python_attribute_keep_alive(self):
         """Python attributes on an ObjCInstance are kept even if the object temporarily has no Python references."""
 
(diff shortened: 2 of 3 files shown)


rubicon/objc/api.py
```diff
@@ -851,17 +828,31 @@ def __setattr__(self, name, value):
...
-                # every ObjCInstance with additional Python attributes is "kept alive" on the Python side.
-                # This prevents the ObjCInstance from being destroyed while the Objective-C object still exists.
-                self._ensure_keep_alive()
+            elif name.startswith("_"):
                 super(ObjCInstance, type(self)).__setattr__(self, name, value)
+            else:
+
+                # Wrap the Python object as an WrappedPyObject instance.
+                # A reference will be retained as long as the WrappedPyObject is alive.
+                wrapper = WrappedPyObject.alloc().initWithObjectId_(id(value))
```
SamSchott, line 838: This is more legible for the time being but should probably be replaced by the appropriate `send_message` calls.

SamSchott:
There is one more issue which I have come across with toga: all example apps segfault when executing a callback, for instance on a button press. The stacktrace is not very helpful and I am not sure if this is an issue with the changes here or with toga itself. Toga may have implicitly relied on (most) ObjCInstances to live forever.

rubicon/objc/api.py
```diff
@@ -803,36 +795,21 @@ def __getattr__(self, name):
...
-                send_message(get_class('DeallocationObserver'), 'alloc', restype=objc_id, argtypes=[]),
-                'initWithObject:', self, restype=objc_id, argtypes=[objc_id]
-            )
-            libobjc.objc_setAssociatedObject(self, observer, observer, 0x301)
-            # The observer is now retained by the object we associate it to.
-            # We now release our own reference to the observer,
-            # so that the only remaining reference is released
-            # when the associated object is deallocated.
-            send_message(observer, 'release', restype=None, argtypes=[])
+        send_message(pyo_wrapper, 'release', restype=None, argtypes=[])
```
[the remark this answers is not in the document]
SamSchott: Never mind, the explicit call to release is not necessary after all. It appears that the object returned by `libobjc.objc_getAssociatedObject()` is autoreleased. Since the tests don't use an `NSAutoreleasePool`, those references where never cleaned up.

SamSchott:
Turns out that the explicit `release` call in `__getattr__()` was not necessary after all and was causing the crashes in toga.

rubicon/objc/api.py
SamSchott: I've replaced this with the appropriate `send_message` calls.

SamSchott:
Alright, this should be ready review :)

rubicon/objc/api.py
```diff
@@ -1554,42 +1552,47 @@ def __subclasscheck__(self, subclass):
...
-# Instances of DeallocationObserver are associated with every
-# ObjCInstance that needs to be kept alive, which is the case
-# when the ObjCInstance has had additional Python attributes added to it.
-# Their sole purpose is to watch for when the Objective-C object
-# is deallocated, and then remove the object from the
-# ObjCInstance._keep_alive_objects dictionary.
+# When a Python object is assigned to a new ObjCInstance attribute, the Python
+# object should be kept alive for the lifetime of the ObjCInstance. This is
+# done by wrapping the Python object as a WrappedPyObject that manages increments
+# the reference count during assignment and decrements it the the WrappedPyObject
```
dgelessus, line 1558:
```suggestion
# done by wrapping the Python object as a WrappedPyObject that increments
# the reference count during assignment and decrements it when the WrappedPyObject
```

rubicon/objc/api.py
```diff
@@ -1554,42 +1552,47 @@ def __subclasscheck__(self, subclass):
...
 # itself are only applied after a restart of course.
 try:
-    DeallocationObserver = ObjCClass("DeallocationObserver")
+    WrappedPyObject = ObjCClass("WrappedPyObject")
 except NameError:
-    class DeallocationObserver(NSObject):
+    class WrappedPyObject(NSObject):
 
-        observed_object = objc_ivar(objc_id)
+        wrapped_pointer = objc_ivar(c_long)
```
dgelessus, line 1577: Would `c_void_p` work here instead, to make it clear that it's a pointer?

rubicon/objc/api.py
```diff
@@ -1554,42 +1552,47 @@ def __subclasscheck__(self, subclass):
...
 
         @objc_rawmethod
-        def initWithObject_(self, cmd, anObject):
+        def initWithObjectId_(self, cmd, address: c_long):
             self = send_message(self, 'init', restype=objc_id, argtypes=[])
             if self is not None:
-                set_ivar(self, 'observed_object', anObject)
+                pyo = cast(address, py_object)
+                pyo_ptr = cast(id(pyo.value), POINTER(PyObject))
```
dgelessus, line 1584: `id(pyo.value)` should be the same as `address`, right? Would it be possible to remove the `pyo` variable and use `address` directly here instead? Or is the cast to `py_object` important for reference counting reasons (or something else)?

```suggestion
                pyo_ptr = cast(address, POINTER(PyObject))
```

rubicon/objc/api.py
```diff
@@ -1554,42 +1552,47 @@ def __subclasscheck__(self, subclass):
...
             return self.value
 
         @objc_rawmethod
         def dealloc(self, cmd) -> None:
-            anObject = get_ivar(self, 'observed_object')
-            ObjCInstance._keep_alive_objects.pop(anObject.value, None)
+            address = get_ivar(self, 'wrapped_pointer')
+            if address.value:
+                pyo = cast(address.value, py_object)
+                pyo_ptr = cast(id(pyo.value), POINTER(PyObject))
```
dgelessus, line 1594: Like above - is the `pyo` variable needed here?

```suggestion
                pyo_ptr = cast(address.value, POINTER(PyObject))
```

rubicon/objc/api.py
```diff
@@ -1554,42 +1552,47 @@ def __subclasscheck__(self, subclass):
...
+                pythonapi.Py_DecRef(pyo_ptr)
```
dgelessus, line 1595: I think `Py_DecRef`'s `restype`/`argtypes` aren't set anywhere in the current version of the PR. `ctypes_patch` declares `Py_IncRef`, but not `Py_DecRef` - and as explained above, I would like to avoid depending on `ctypes_patch` for these definitions.

rubicon/objc/api.py
```diff
@@ -1598,8 +1601,11 @@ def finalize(self, cmd) -> None:
             # (which would have to be explicitly started with
             # objc_startCollectorThread(), so probably not too much reason
             # to have this here, but I guess it can't hurt.)
-            anObject = get_ivar(self, 'observed_object')
-            ObjCInstance._keep_alive_objects.pop(anObject.value, None)
+            address = get_ivar(self, 'wrapped_pointer')
+            if address.value:
+                pyo = cast(address.value, py_object)
+                pyo_ptr = cast(id(pyo.value), POINTER(PyObject))
```
dgelessus, line 1607:
```suggestion
                pyo_ptr = cast(address.value, POINTER(PyObject))
```

tests/test_core.py
```diff
@@ -1163,21 +1167,36 @@ def test_objcinstance_python_attribute(self):
...
 
-        Example = ObjCClass("Example")
-        example = Example.alloc().init()
-        Thing = ObjCClass("Thing")
-        thing = Thing.alloc().init()
+        example = NSObject.alloc().init()
+        thing = NSObject.alloc().init()
+
+        # create NSAutoreleasePool to clean up any leftover Obj-C instances
+        pool = NSAutoreleasePool.alloc().init()
```
dgelessus, line 1184: This isn't something that needs to be done in this PR, but ideally Rubicon should provide a proper API for autorelease pools, probably as a context manager (`with objc.autoreleasepool(): ...`).

In normal Objective-C code, `NSAutoreleasePool` is more or less superseded by `@autoreleasepool` blocks, and when using ARC, `@autoreleasepool` is the only option - see the warning in [the `NSAutoreleasePool` docs](https://developer.apple.com/documentation/foundation/nsautoreleasepool?language=objc). Internally, `@autoreleasepool` blocks are translated to calls to the Objective-C runtime functions `objc_autoreleasePoolPush` and `objc_autoreleasePoolPop`, which are undocumented and not declared in Apple's headers, but [seem to be widely known](https://stackoverflow.com/q/15014152). It's *probably* fine if we keep using `NSAutoreleasePool` for now - it's not officially deprecated, and we aren't affected by ARC restrictions. But for future-proofing it would be better to define our own high-level API for autorelease pools that uses the appropriate mechanism under the hood. (Again, not something that needs to be done in this PR.)

tests/test_core.py
```diff
@@ -1163,21 +1167,36 @@ def test_objcinstance_python_attribute(self):
...
+
     def test_objcinstance_python_attribute_keep_alive(self):
         """Python attributes on an ObjCInstance are kept even if the object temporarily has no Python references."""
 
-        Example = ObjCClass("Example")
-        example = Example.alloc().init()
-        Thing = ObjCClass("Thing")
-        thing = Thing.alloc().init()
+        example = NSObject.alloc().init()
+        thing = NSObject.alloc().init()
```
dgelessus, line 1181: I think this changes the behavior of the test quite significantly, along with the change below from `example.setThing(thing)` to `example.thing = thing`. Previously, it tested that:

* If the Objective-C object `thing` has Python attributes
* and it is stored inside an Objective-C property on another object (`example`)
* and `thing` temporarily has *no* references from Python
* once `thing` is retrieved from the Objective-C property, it still has its Python attributes

The new test code is similar, but it stores `thing` in a *Python property* on an Objective-C object, and not in a real Objective-C property. The difference is relevant because when `thing` is stored in a Python property, its Python `ObjCInstance` is kept alive. This test is supposed to test the case where the `ObjCInstance` is temporarily destroyed, but the underlying Objective-C object continues to exist, and is then later used from Python again.

The test as it's written now is probably useful to have as well, but it should be a separate test in addition to the old one, not a replacement.

dgelessus (commented):
Thank you for the PR, and sorry for the late review!

Interesting approach, I hadn't thought about storing the Python objects inside Objective-C associated objects. My thought would have been adding a variant of `objc_property` that stores a Python object, and to require users to declare all Python object properties that they want to use. But this implementation using associated objects is much nicer - it no longer requires keeping `ObjCInstance`s alive, but still allows setting arbitrary Python attributes on any Objective-C object. That should mean that no existing user code needs to be changed, which is always a good thing.

rubicon/objc/api.py
```diff
@@ -16,6 +16,7 @@
     Class, SEL, add_ivar, add_method, ensure_bytes, get_class, get_ivar, libc, libobjc, objc_block, objc_id,
     objc_property_attribute_t, object_isClass, set_ivar, send_message, send_super,
 )
+from .ctypes_patch import PyObject
```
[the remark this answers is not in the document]
SamSchott: That's a fair point.

rubicon/objc/api.py
```diff
@@ -851,17 +827,39 @@ def __setattr__(self, name, value):
...
+
+                wrapper = send_message(
+                    send_message(get_class('WrappedPyObject'), 'alloc',
+                                 restype=objc_id, argtypes=[]),
+                    'initWithObjectId:', id(value), restype=objc_id, argtypes=[objc_id]
+                )
+
+                # Set the Python value as an associated object. This will release
+                # any previous wrapper object with the same key.
+                libobjc.objc_setAssociatedObject(self, SEL(name), wrapper, 0x301)
```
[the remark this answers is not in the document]
SamSchott: Agreed, it's better to have a namespace for those automatically defined selectors.

rubicon/objc/api.py
```diff
@@ -851,17 +827,39 @@ def __setattr__(self, name, value):
...
-                # Add a new Python attribute.
-                # These attributes are currently only stored on the Python ObjCInstance object
-                # and not on the actual Objective-C object.
-                # To ensure that the attributes aren't lost when the ObjCInstance has no more Python references
-                # (but the Objective-C object still has Objective-C references
-                # and may be wrapped in an ObjCInstance again in the future),
-                # every ObjCInstance with additional Python attributes is "kept alive" on the Python side.
-                # This prevents the ObjCInstance from being destroyed while the Objective-C object still exists.
-                self._ensure_keep_alive()
+            elif name.startswith("_"):
```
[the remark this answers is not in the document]
SamSchott: I did run into trouble with the `_needs_release` attribute previously but cannot reproduce it now. As you say, it will already exist in the `__dict__` and is therefore caught by the previous check. Pending any upcoming issues, I'll remove the special handling of private attributes.

rubicon/objc/api.py
SamSchott: Good catch, `id(pyo.value)` and `address` are indeed the same. This roundabout way is just a leftover from previous code which I didn't clean up properly.

rubicon/objc/api.py
SamSchott: Sure, I can add those definitions in a new module for consistency.

SamSchott:
Thanks for the diligent review! I've addressed most points but still need to move all CPython API definitions to a separate module.

In the meantime, I came across two questions related to freeing unreferenced objects:

1. I tried to adapt the original `test_objcinstance_python_attribute_keep_alive` test to release all Python objects at the end by deleting `thing` and `example` and using an autorelease pool. Somehow, I could not get it to work. Is there anything about the `Example` and `Thing` classes which would prevent the Python objects or the classes themselves from being released?

2. While this PR *should* fix memory leaks in toga, it currently does not. The reason for this are circular references introduced in the Objective-C side. This is a common pattern throughout toga-cocoa:

    ```Python
    from rubicon.objc import NSObject


    class InterfaceObject()
        pass


    class ImplObject()
        pass


    interface = InterfaceObject()
    interface._impl = ImplObject()
    interface._impl.native = NSObject.alloc().init()

    interface._impl.native.interface = interface
    interface._impl.native.impl = interface._impl

   del interface  # should release the entire spagetti but does not
    ```

    The resulting circular references are obviously not garbage collected because Python does not know who is holding on to the remaining references of `interface` and `impl`. Writing our own garbage collection is a really bad idea. Ideally, one would like to declare weak properties on the `NSObject` to store arbitrary Python objects. Do you think this is feasible?

    Introducing such circular references somehow strikes me as a bad design pattern in the first place and maybe it can be avoided. AFAIK, the `interface` and `impl` attributes are only used when `NSObject` is its own target or delegate and needs access to the interface or implementation layer in callbacks. This could be solved by writing separate delegates instead of subclassing the original objects. Since `setDelegate()` does not take ownership of the object, we avoid the reference cycle. @freakboy3742, what do you think?

rubicon/objc/api.py
SamSchott: Yes.

rubicon/objc/api.py
SamSchott: Done.

dgelessus:
re. the CPython APIs: sorry for going back and forth on this, but it would actually be better to only pull out those parts of `ctypes_patch` that are part of the documented CPython API and thus safe to use (namely `PyTypeObject`, `PyObject`, `PyVarObject`, `Py_IncRef`, and `Py_DecRef`). All other declarations (`PyDict_Type`, `PyDictObject`, `mappingproxyobject`, `ffi_type`, `GETFUNC`, `SETFUNC`, and `StgDictObject`) are undocumented and possibly unstable, so it would be better to leave those in `ctypes_patch`, together with the other possibly dangerous code.

Also, I think it would be better to name the submodule something other than `pythonapi`, because `ctypes` already uses the name `pythonapi` for its predefined library object for the CPython API. The two names aren't actually conflicting, because the code uses the right combination of `from`-`import`s and only imports one of the two `pythonapi`s directly. But IMO it's still a bit confusing and possibly error-prone to have the module and the library named exactly the same. Maybe the module could be named `pythonapi_defs` or something like that? Then it would be clear that the two are related, but not the same.

> I tried to adapt the original `test_objcinstance_python_attribute_keep_alive` test to release all Python objects at the end by deleting `thing` and `example` and using an autorelease pool. Somehow, I could not get it to work. Is there anything about the `Example` and `Thing` classes which would prevent the Python objects or the classes themselves from being released?

Hmm, not sure what could be going on there... There's always the problem that *officially* Python doesn't guarantee when a Python object is destroyed once it's no longer referenced - but in practice the way that you're doing it (`del` all references to the object, then `gc.collect()`) should be reliable enough. It's especially strange that (if I understand you correctly) the object gets deleted if you use `NSObject` but not if you use `Example` and `Thing`.

> The resulting circular references are obviously not garbage collected because Python does not know who is holding on to the remaining references of `interface` and `impl` on the Objective-C side. Writing our own garbage collection is a really bad idea.

Agreed that we shouldn't try to build our own garbage collection. But sadly I think there's also no way we could make Python's garbage collection properly understand references on the Objective-C side, especially when it comes to more complex reference cycles like "Python object -> ObjC object 1 -> ObjC object 2 -> Python object".

> Ideally, one would like to declare weak properties on the `NSObject` to store arbitrary Python objects. Do you think this is feasible?

That's probably the best solution. In the Objective-C -> Python direction it should be easy to do by just storing a Python `weakref` as an attribute on the Objective-C object. The other direction could be implemented using something similar to the existing `DeallocationObserver` - although Objective-C also has native support for weak references, which could probably be used to implement Python -> Objective-C weakrefs more efficiently and without another custom associated object.

dgelessus:
Thinking about the CPython API thing some more - we could completely avoid calling the CPython API if we keep using the existing `_keep_alive_objects` solution for keeping objects alive. Basically, `pythonapi.Py_IncRef(pyo_ptr)` could be replaced with `ObjCInstance._keep_alive_objects[id(pyo)] = pyo`, and `pythonapi.Py_DecRef(pyo_ptr)` with `del ObjCInstance._keep_alive_objects[id(pyo)]`. The difference to before would be that now we only keep the individual attribute values alive, and not the entire `ObjCInstance` object as before.

SamSchott:
No worries about the Python API, I fully understand your points about the module name and about not exposing undocumented declarations.

> Thinking about the CPython API thing some more - we could completely avoid calling the CPython API if we keep using the existing _keep_alive_objects solution for keeping objects alive.

Possibly yes, but we would also have to store the reference count for the Objective-C side and only remove the Python object from the dict once it drops to zero. This would handle cases where a Python object is referred to from multiple Objective-C instances. Basically our own simple reference counting. The advantage of this approach is that conventional methods of finding memory leaks such as `gc. get_referrers()` would at least show that the references were created by rubicon-objc. I'd be happy with either solution.

> It's especially strange that (if I understand you correctly) the object gets deleted if you use NSObject but not if you use Example and Thing.

Yes, this is what confused me as well. I also didn't see any issues with other Cocoa classes such as `NSButton`, etc.

Regarding storing weak references to Python objects, I am not sure I fully understand your suggestions, especially regarding the  Python -> Objective-C direction. Do you mean creating weak references to Objective-C objects?

Broadly, I can see the following APIs:

1. Allow `objc_property` to accept Python objects and provide an additional argument specifying weak vs strong references. Strong references to Python objects can again be implemented by managing the ref count whereas weak references would just leave the ref count alone.
2. Provide some kind of `weakref` wrapper to wrap Python objects when assigning them to Objective-C attributes. This would enable creating attributes with weak references for individual instances instead of declaring them for the class. For example:

    ```Python
    obj = NSObject.alloc().init()
    pyo = object()
    obj.attribute = objc_weakref(pyo)
    assert objc.attribute is pyo
   ```
   
   This `objc_weakref` would just tell the `WrappedPyObject` not to manage the ref counts. This approach might however be confusing.
3. Don't deal with the issue at all in rubicon-objc and require users to manually use Python's `weakref` module.

I tend to prefer the first option, also because it allows storing weak references to both Python and Objective-C objects. What do you think?

dgelessus:
> Possibly yes, but we would also have to store the reference count for the Objective-C side and only remove the Python object from the dict once it drops to zero. This would handle cases where a Python object is referred to from multiple Objective-C instances.

Ah, good point, I forgot about that case. I think this could also be solved by making the key a tuple that contains both the address of the referenced Python object and the address of the referencing `WrappedPyObject`. That would avoid having to write our own reference counting.

IMO a Python-based solution like this would be better than calling `Py_IncRef`/`Py_DecRef` via `ctypes`. It saves us the trouble of having to interact with the C API, it's easier to ensure that the references are added/deleted correctly, and as you've already mentioned it makes it easier to trace the references while debugging.

> Regarding storing weak references to Python objects, I am not sure I fully understand your suggestions, especially regarding the Python -> Objective-C direction. Do you mean creating weak references to Objective-C objects?

When I replied, I wasn't sure which direction you were asking about, so I wrote a short reply about each direction 😄 But re-reading your comment again now I think you were only talking about adding weak references from an Objective-C object to a Python object, not the other way around, so I guess you can ignore the second half of my comment 🙂

I think the simplest solution for now is to have no special support for weak references in Rubicon. In cases where a reference to a Python object needs to be weak, you can manually wrap the object in a `weakref` before storing it as an attribute. The overhead of doing it manually shouldn't be very big - you just need to replace `obj.attr = value` with `obj.attr = weakref.weakref(value)` when setting the attribute, and replace `obj.attr` with `obj.attr()` when using it.

I agree that it would be nice if we could declare Objective-C properties for storing Python objects (i. e. `objc_property(py_object)`). Then we could add native support for weak references in `objc_property` as you're suggesting. But that would require some additional work on `objc_property` that isn't really related to this PR (replacing `DeallocationObserver`), so IMO it would be better to do that as a separate PR in the future.

SamSchott:
> I think this could also be solved by making the key a tuple that contains both the address of the referenced Python object and the address of the referencing WrappedPyObject

Agreed, that solution is easier to manage and debug than using the C API. It might also come in handy when writing tests, though we probably mostly want to test the API (i.e., that Python objects are retained and freed correctly) and not the implementation.

Regarding weak references, yes, I was only talking about weak references in the Objective-C -> Python direction, though the reverse could certainly be useful as well. But in any case, I agree that this is best done in other PR. I just raised it here because I was a bit disappointed when realising that this PR does not solve the memory leaks in toga-cocoa :(

dgelessus (approved):
Alright, I think that's everything addressed now! Thanks for the PR - even if it doesn't directly fix the memory management issues you're running into, it makes the Python attribute storage more robust, and will probably help with implementing `objc_property(py_object)` in the future.

Before merging - I'm thinking if it would make sense to drop the commits be7e48c5a4bf3b63f3008f533a908e235f157017 and dbf540cf8868ea375e3ebbb839717d8fd4d0b526 from the PR. The second commit just reverts the first one, and they are quite close to each other in the history, so it should be possible to safely remove them. This would help with keeping the Git history (and especially `git blame`) a bit cleaner in the long term.

I don't know how familiar/comfortable you are with interactive rebase and rewriting Git history - you can make the changes yourself if you like, but I can also do while merging the PR if you prefer.

SamSchott:
Thanks for the thorough reviews and for bearing with my while I was finding my way through ctypes and the ObjC runtime!

Yeah, there are a lot of commits in the history just from trying different approaches and reverting them. I've squashed them all into a single commit for the changes to the packages itself and left separate commits for the tests and towncrier entry.

tests/test_core.py
SamSchott: Yes, an `autoreleasepool` context manager would be a good idea. A lot of code that uses rubicon-objc will likely also run a Cocoa event loop which will create and drain pools for each cycle but there of course will remain use cases for manually managing an autorelease pool.

I'm not so sure about using undocumented Objective-C runtime functions. Could this potentially lead to issues with app review if they are considered private? I have never gone through the review process myself...

dgelessus: I've copied the discussion into a separate issue and replied to your question there: #210

PR #211, opened 2021-06-27
Allow declaring weak properties

SamSchott:
This PR adds support for declaring weak properties when creating custom Objective-C classes.

## PR Checklist:

### DIFF
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -445,27 +445,18 @@ class objc_property(object):
         attrs = self._get_property_attributes()
         libobjc.class_addProperty(class_ptr, ensure_bytes(attr_name), attrs, len(attrs))
 
-        # Add cleanup routines to dealloc.
+    def dealloc_callback(self, objc_self, attr_name):
 
-        old_dealloc = libobjc.class_getMethodImplementation(class_ptr, SEL("dealloc"))
+        ivar_name = '_' + attr_name
 
-        def _new_delloc(objc_self, _cmd):
-
-            # Clean up ivar.
-            if self.weak:
-                # Clean up weak reference.
-                set_ivar(objc_self, ivar_name, self.vartype(None), weak=True)
-            elif issubclass(self.vartype, objc_id):
-                # If the old value is a non-null object, release it. There is no need to set the actual ivar to nil.
-                old_value = get_ivar(objc_self, ivar_name, weak=self.weak)
-                send_message(old_value, 'release', restype=None, argtypes=[])
-
-            # Invoke original dealloc.
-            cfunctype = CFUNCTYPE(None, objc_id, SEL)
-            old_dealloc_callable = cast(old_dealloc, cfunctype)
-            old_dealloc_callable(objc_self, SEL("dealloc"))
-
-        add_method(class_ptr, 'dealloc', _new_delloc, [None, ObjCInstance, SEL], replace=True)
+        # Clean up ivar.
+        if self.weak:
+            # Clean up weak reference.
+            set_ivar(objc_self, ivar_name, self.vartype(None), weak=True)
+        elif issubclass(self.vartype, objc_id):
+            # If the old value is a non-null object, release it. There is no need to set the actual ivar to nil.
+            old_value = get_ivar(objc_self, ivar_name, weak=self.weak)
+            send_message(old_value, 'release', restype=None, argtypes=[])
 
     def protocol_register(self, proto_ptr, attr_name):
         attrs = self._get_property_attributes()
@@ -1008,6 +999,24 @@ class ObjCClass(ObjCInstance, type):
                 pass
             else:
                 class_register(ptr, attr_name)
+
+        # Add cleanup of ivars / properties to dealloc
+
+        old_dealloc = libobjc.class_getMethodImplementation(ptr, SEL("dealloc"))
+
+        def _new_delloc(objc_self, _cmd):
+
+            # Invoke dealloc callback of each property.
+            for attr_name, obj in attrs.items():
+                if isinstance(obj, objc_property):
+                    obj.dealloc_callback(objc_self, attr_name)
+
+            # Invoke original dealloc.
+            cfunctype = CFUNCTYPE(None, objc_id, SEL)
+            old_dealloc_callable = cast(old_dealloc, cfunctype)
+            old_dealloc_callable(objc_self, SEL("dealloc"))
+
+        add_method(ptr, "dealloc", _new_delloc, [None, ObjCInstance, SEL], replace=True)
 
         # Register the ObjC class
         libobjc.objc_registerClassPair(ptr)
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -1008,8 +1008,12 @@ class ObjCClass(ObjCInstance, type):
 
             # Invoke dealloc callback of each property.
             for attr_name, obj in attrs.items():
-                if isinstance(obj, objc_property):
-                    obj.dealloc_callback(objc_self, attr_name)
+                try:
+                    dealloc_callback = obj.dealloc_callback
+                except AttributeError:
+                    pass
+                else:
+                    dealloc_callback(objc_self, attr_name)
 
             # Invoke original dealloc.
             cfunctype = CFUNCTYPE(None, objc_id, SEL)
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -442,6 +442,28 @@ class objc_property(object):
         attrs = self._get_property_attributes()
         libobjc.class_addProperty(class_ptr, ensure_bytes(attr_name), attrs, len(attrs))
 
+        # Add cleanup routines to dealloc.
+
+        old_dealloc = libobjc.class_getMethodImplementation(class_ptr, SEL("dealloc"))
+
+        def _new_delloc(objc_self, _cmd):
+
+            # Clean up ivar.
+            if self.weak:
+                ivar = libobjc.class_getInstanceVariable(libobjc.object_getClass(objc_self), ('_' + attr_name).encode())
+                libobjc.objc_storeWeak(objc_self.value + libobjc.ivar_getOffset(ivar), None)
+            elif issubclass(self.vartype, objc_id):
+                # If the old value is a non-null object, release it. The is no need to set the actual ivar to nil.
+                old_value = get_ivar(objc_self, '_' + attr_name, weak=self.weak)
+                send_message(old_value, 'release', restype=None, argtypes=[])
+
+            # Invoke original dealloc.
+            cfunctype = CFUNCTYPE(None, objc_id, SEL)
+            old_dealloc_callable = cast(old_dealloc, cfunctype)
+            old_dealloc_callable(objc_self, SEL("dealloc"))
+
+        replace_method(class_ptr, 'dealloc', _new_delloc, [None, ObjCInstance, SEL])
+
     def protocol_register(self, proto_ptr, attr_name):
         attrs = self._get_property_attributes()
         libobjc.protocol_addProperty(proto_ptr, ensure_bytes(attr_name), attrs, len(attrs), True, True)
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -454,8 +454,7 @@ class objc_property(object):
             # Clean up ivar.
             if self.weak:
                 # Clean up weak reference.
-                ivar = libobjc.class_getInstanceVariable(libobjc.object_getClass(objc_self), ivar_name.encode())
-                libobjc.objc_storeWeak(objc_self.value + libobjc.ivar_getOffset(ivar), None)
+                set_ivar(objc_self, ivar_name, self.vartype(None), weak=True)
             elif issubclass(self.vartype, objc_id):
                 # If the old value is a non-null object, release it. There is no need to set the actual ivar to nil.
                 old_value = get_ivar(objc_self, ivar_name, weak=self.weak)
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -1023,7 +1023,7 @@ class ObjCClass(ObjCInstance, type):
                     dealloc_callback(objc_self, attr_name)
 
             # Invoke super dealloc.
-            send_super(ptr, objc_self, "dealloc", restype=None, argtypes=[])
+            send_super(ptr, objc_self, "dealloc", restype=None, argtypes=[], _allow_dealloc=True)
 
         add_method(ptr, "dealloc", _new_delloc, [None, ObjCInstance, SEL])
 
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -368,26 +368,33 @@ class objc_property(object):
     the generated setter keeps the stored object retained, and releases it when it is replaced.
 
     In a custom Objective-C protocol, only the metadata for the property is generated.
+
+    If ``weak`` is ``True``, the property will be created as a weak property. When assigning an object to it,
+    the reference count of the object will not be increased. When the object is deallocated, the property
+    value is set to None.
     """
 
-    def __init__(self, vartype=objc_id):
+    def __init__(self, vartype=objc_id, weak=False):
         super().__init__()
 
         self.vartype = ctype_for_type(vartype)
+        self.weak = weak
 
     def _get_property_attributes(self):
         attrs = [
             objc_property_attribute_t(b'T', encoding_for_ctype(self.vartype)),  # Type: vartype
         ]
         if issubclass(self.vartype, objc_id):
-            attrs.append(objc_property_attribute_t(b'&', b''))  # retain
+            reference = b'W' if self.weak else b'&'
+            attrs.append(objc_property_attribute_t(reference, b''))
         return (objc_property_attribute_t * len(attrs))(*attrs)
 
     def class_register(self, class_ptr, attr_name):
         add_ivar(class_ptr, '_' + attr_name, self.vartype)
 
         def _objc_getter(objc_self, _cmd):
-            value = get_ivar(objc_self, '_' + attr_name)
+            value = get_ivar(objc_self, '_' + attr_name, weak=self.weak)
+
             # ctypes complains when a callback returns a "boxed" primitive type, so we have to manually unbox it.
             # If the data object has a value attribute and is not a structure or union, assume that it is
             # a primitive and unbox it.
@@ -403,12 +410,21 @@ class objc_property(object):
             if not isinstance(new_value, self.vartype):
                 # If vartype is a primitive, then new_value may be unboxed. If that is the case, box it manually.
                 new_value = self.vartype(new_value)
-            old_value = get_ivar(objc_self, '_' + attr_name)
-            if issubclass(self.vartype, objc_id) and new_value:
+
+            if issubclass(self.vartype, objc_id) and not self.weak:
+                old_value = get_ivar(objc_self, '_' + attr_name, weak=self.weak)
+
+                if new_value.value == old_value.value:
+                    # old and new value are the same, nothing to do
+                    return
+
+            if not self.weak and issubclass(self.vartype, objc_id) and new_value:
                 # If the new value is a non-null object, retain it.
                 send_message(new_value, 'retain', restype=objc_id, argtypes=[])
-            set_ivar(objc_self, '_' + attr_name, new_value)
-            if issubclass(self.vartype, objc_id) and old_value:
+
+            set_ivar(objc_self, '_' + attr_name, new_value, weak=self.weak)
+
+            if not self.weak and issubclass(self.vartype, objc_id) and old_value:
                 # If the old value is a non-null object, release it.
                 send_message(old_value, 'release', restype=None, argtypes=[])
 
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -390,10 +390,13 @@ class objc_property(object):
         return (objc_property_attribute_t * len(attrs))(*attrs)
 
     def class_register(self, class_ptr, attr_name):
-        add_ivar(class_ptr, '_' + attr_name, self.vartype)
+
+        ivar_name = '_' + attr_name
+
+        add_ivar(class_ptr, ivar_name, self.vartype)
 
         def _objc_getter(objc_self, _cmd):
-            value = get_ivar(objc_self, '_' + attr_name, weak=self.weak)
+            value = get_ivar(objc_self, ivar_name, weak=self.weak)
 
             # ctypes complains when a callback returns a "boxed" primitive type, so we have to manually unbox it.
             # If the data object has a value attribute and is not a structure or union, assume that it is
@@ -412,7 +415,7 @@ class objc_property(object):
                 new_value = self.vartype(new_value)
 
             if issubclass(self.vartype, objc_id) and not self.weak:
-                old_value = get_ivar(objc_self, '_' + attr_name, weak=self.weak)
+                old_value = get_ivar(objc_self, ivar_name, weak=self.weak)
 
                 if new_value.value == old_value.value:
                     # old and new value are the same, nothing to do
@@ -422,7 +425,7 @@ class objc_property(object):
                 # If the new value is a non-null object, retain it.
                 send_message(new_value, 'retain', restype=objc_id, argtypes=[])
 
-            set_ivar(objc_self, '_' + attr_name, new_value, weak=self.weak)
+            set_ivar(objc_self, ivar_name, new_value, weak=self.weak)
 
             if not self.weak and issubclass(self.vartype, objc_id) and old_value:
                 # If the old value is a non-null object, release it.
@@ -450,11 +453,12 @@ class objc_property(object):
 
             # Clean up ivar.
             if self.weak:
-                ivar = libobjc.class_getInstanceVariable(libobjc.object_getClass(objc_self), ('_' + attr_name).encode())
+                # Clean up weak reference.
+                ivar = libobjc.class_getInstanceVariable(libobjc.object_getClass(objc_self), ivar_name.encode())
                 libobjc.objc_storeWeak(objc_self.value + libobjc.ivar_getOffset(ivar), None)
             elif issubclass(self.vartype, objc_id):
-                # If the old value is a non-null object, release it. The is no need to set the actual ivar to nil.
-                old_value = get_ivar(objc_self, '_' + attr_name, weak=self.weak)
+                # If the old value is a non-null object, release it. There is no need to set the actual ivar to nil.
+                old_value = get_ivar(objc_self, ivar_name, weak=self.weak)
                 send_message(old_value, 'release', restype=None, argtypes=[])
 
             # Invoke original dealloc.
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -13,7 +13,7 @@ from .types import (
     register_ctype_for_type
 )
 from .runtime import (
-    Class, SEL, add_ivar, replace_method, ensure_bytes, get_class, get_ivar, libc, libobjc, objc_block, objc_id,
+    Class, SEL, add_ivar, add_method, ensure_bytes, get_class, get_ivar, libc, libobjc, objc_block, objc_id,
     objc_property_attribute_t, object_isClass, set_ivar, send_message, send_super,
 )
 
@@ -279,7 +279,7 @@ class objc_method(object):
 
     def class_register(self, class_ptr, attr_name):
         name = attr_name.replace("_", ":")
-        replace_method(class_ptr, name, self, self.encoding)
+        add_method(class_ptr, name, self, self.encoding)
 
     def protocol_register(self, proto_ptr, attr_name):
         name = attr_name.replace('_', ':')
@@ -315,7 +315,7 @@ class objc_classmethod(object):
 
     def class_register(self, class_ptr, attr_name):
         name = attr_name.replace("_", ":")
-        replace_method(libobjc.object_getClass(class_ptr), name, self, self.encoding)
+        add_method(libobjc.object_getClass(class_ptr), name, self, self.encoding)
 
     def protocol_register(self, proto_ptr, attr_name):
         name = attr_name.replace('_', ':')
@@ -433,11 +433,11 @@ class objc_property(object):
 
         setter_name = 'set' + attr_name[0].upper() + attr_name[1:] + ':'
 
-        replace_method(
+        add_method(
             class_ptr, attr_name, _objc_getter,
             [self.vartype, ObjCInstance, SEL],
         )
-        replace_method(
+        add_method(
             class_ptr, setter_name, _objc_setter,
             [None, ObjCInstance, SEL, self.vartype],
         )
@@ -466,7 +466,7 @@ class objc_property(object):
             old_dealloc_callable = cast(old_dealloc, cfunctype)
             old_dealloc_callable(objc_self, SEL("dealloc"))
 
-        replace_method(class_ptr, 'dealloc', _new_delloc, [None, ObjCInstance, SEL])
+        add_method(class_ptr, 'dealloc', _new_delloc, [None, ObjCInstance, SEL], replace=True)
 
     def protocol_register(self, proto_ptr, attr_name):
         attrs = self._get_property_attributes()
@@ -501,7 +501,7 @@ class objc_rawmethod(object):
 
     def class_register(self, class_ptr, attr_name):
         name = attr_name.replace("_", ":")
-        replace_method(class_ptr, name, self, self.encoding)
+        add_method(class_ptr, name, self, self.encoding)
 
     def protocol_register(self, proto_ptr, attr_name):
         raise TypeError('Protocols cannot have method implementations, use objc_method instead of objc_rawmethod')
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -993,20 +993,27 @@ class ObjCClass(ObjCInstance, type):
 
         # Register all methods, properties, ivars, etc.
         for attr_name, obj in attrs.items():
-            try:
-                class_register = obj.class_register
-            except AttributeError:
-                pass
-            else:
-                class_register(ptr, attr_name)
+            if attr_name != "dealloc":
+                try:
+                    class_register = obj.class_register
+                except AttributeError:
+                    pass
+                else:
+                    class_register(ptr, attr_name)
 
-        # Add cleanup of ivars / properties to dealloc
+        # Register any user-defined dealloc method. We treat dealloc differently to
+        # inject our own cleanup code for properties, ivars, etc.
 
-        old_dealloc = libobjc.class_getMethodImplementation(ptr, SEL("dealloc"))
+        user_dealloc = attrs.get("dealloc", None)
 
         def _new_delloc(objc_self, _cmd):
 
-            # Invoke dealloc callback of each property.
+            # Invoke user-defined dealloc.
+            if user_dealloc:
+                user_dealloc(objc_self, _cmd)
+
+            # Invoke dealloc callback of each attribute. Currently
+            # defined for properties only.
             for attr_name, obj in attrs.items():
                 try:
                     dealloc_callback = obj.dealloc_callback
@@ -1015,12 +1022,10 @@ class ObjCClass(ObjCInstance, type):
                 else:
                     dealloc_callback(objc_self, attr_name)
 
-            # Invoke original dealloc.
-            cfunctype = CFUNCTYPE(None, objc_id, SEL)
-            old_dealloc_callable = cast(old_dealloc, cfunctype)
-            old_dealloc_callable(objc_self, SEL("dealloc"))
+            # Invoke super dealloc.
+            send_super(ptr, objc_self, "dealloc", restype=None, argtypes=[])
 
-        add_method(ptr, "dealloc", _new_delloc, [None, ObjCInstance, SEL], replace=True)
+        add_method(ptr, "dealloc", _new_delloc, [None, ObjCInstance, SEL])
 
         # Register the ObjC class
         libobjc.objc_registerClassPair(ptr)
@@ -1656,7 +1661,6 @@ except NameError:
             address = get_ivar(self, 'wrapped_pointer')
             if address.value:
                 del _keep_alive_objects[(self.value, address.value)]
-            send_super(__class__, self, 'dealloc', restype=None, argtypes=[])
 
         @objc_rawmethod
         def finalize(self, cmd) -> None:
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -13,7 +13,7 @@ from .types import (
     register_ctype_for_type
 )
 from .runtime import (
-    Class, SEL, add_ivar, add_method, ensure_bytes, get_class, get_ivar, libc, libobjc, objc_block, objc_id,
+    Class, SEL, add_ivar, replace_method, ensure_bytes, get_class, get_ivar, libc, libobjc, objc_block, objc_id,
     objc_property_attribute_t, object_isClass, set_ivar, send_message, send_super,
 )
 
@@ -279,7 +279,7 @@ class objc_method(object):
 
     def class_register(self, class_ptr, attr_name):
         name = attr_name.replace("_", ":")
-        add_method(class_ptr, name, self, self.encoding)
+        replace_method(class_ptr, name, self, self.encoding)
 
     def protocol_register(self, proto_ptr, attr_name):
         name = attr_name.replace('_', ':')
@@ -315,7 +315,7 @@ class objc_classmethod(object):
 
     def class_register(self, class_ptr, attr_name):
         name = attr_name.replace("_", ":")
-        add_method(libobjc.object_getClass(class_ptr), name, self, self.encoding)
+        replace_method(libobjc.object_getClass(class_ptr), name, self, self.encoding)
 
     def protocol_register(self, proto_ptr, attr_name):
         name = attr_name.replace('_', ':')
@@ -430,11 +430,11 @@ class objc_property(object):
 
         setter_name = 'set' + attr_name[0].upper() + attr_name[1:] + ':'
 
-        add_method(
+        replace_method(
             class_ptr, attr_name, _objc_getter,
             [self.vartype, ObjCInstance, SEL],
         )
-        add_method(
+        replace_method(
             class_ptr, setter_name, _objc_setter,
             [None, ObjCInstance, SEL, self.vartype],
         )
@@ -475,7 +475,7 @@ class objc_rawmethod(object):
 
     def class_register(self, class_ptr, attr_name):
         name = attr_name.replace("_", ":")
-        add_method(class_ptr, name, self, self.encoding)
+        replace_method(class_ptr, name, self, self.encoding)
 
     def protocol_register(self, proto_ptr, attr_name):
         raise TypeError('Protocols cannot have method implementations, use objc_method instead of objc_rawmethod')
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -18,7 +18,7 @@ __all__ = [
     'Method',
     'SEL',
     'add_ivar',
-    'replace_method',
+    'add_method',
     'get_class',
     'get_ivar',
     'libc',
@@ -858,8 +858,8 @@ def send_super(cls, receiver, selector, *args, restype=c_void_p, argtypes=None):
 _keep_alive_imps = []
 
 
-def replace_method(cls, selector, method, encoding):
-    """Add a new instance method to the given class or replace an existing instance method.
+def add_method(cls, selector, method, encoding, replace=False):
+    """Add a new instance method to the given class.
 
     To add a class method, add an instance method to the metaclass.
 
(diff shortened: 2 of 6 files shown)


SamSchott:
Hrmp, not sure why this segfaults in Python 3.5 :/ The tests all pass locally on Python 3.9.

SamSchott:
I am having trouble installing Python 3.5 on my own machine with macOS 11 but can confirm that the tests all run and pass on Python 3.6.

freakboy3742:
FWIW, I can confirm I'm seeing the crash on Py3.5, but tests pass fine on 3.6-3.9.

One possibility here would be to drop Python 3.5 support. 

Dropping support because of a bug we can't explain doesn't make me *entirely* comfortable, but if it turns out that Python 3.5 behavior is the underlying problem (which seems plausible, given the changes 3.6 introduced around dict ordering), then I'd be open to dropping 3.5 support. We've dropped Python 3.5 elsewhere in BeeWare, and 3.5 is no longer officially supported by Python itself.

SamSchott:
@freakboy3742, can you check which test exactly is causing the crash? It should (hopefully) be one of the two new ones which I have introduced. And is there a useful stacktrace?

I agree that Python 3.5 support is not so important but as you say, it would still be good to understand why it crashes.

rubicon/objc/runtime.py
```diff
@@ -888,6 +904,50 @@ def add_ivar(cls, name, vartype):
     )
 
 
+def get_weak_ivar(obj, varname):
```
dgelessus, line 907: Would it make sense to merge this into the existing `get_ivar` function (using a `weak` keyword argument that defaults to `False`) instead of defining a separate function? As far as I can tell, the code in `get_weak_ivar` and `get_ivar` is very similar.

The same question applies to `set_weak_ivar`/`set_ivar`.

rubicon/objc/api.py
```diff
@@ -403,12 +412,25 @@ def _objc_setter(objc_self, _cmd, new_value):
             if not isinstance(new_value, self.vartype):
                 # If vartype is a primitive, then new_value may be unboxed. If that is the case, box it manually.
                 new_value = self.vartype(new_value)
-            old_value = get_ivar(objc_self, '_' + attr_name)
-            if issubclass(self.vartype, objc_id) and new_value:
+
+            if self.weak:
+                old_value = get_weak_ivar(objc_self, '_' + attr_name)
+            else:
+                old_value = get_ivar(objc_self, '_' + attr_name)
```
dgelessus, line 419: If `get_weak_ivar` and `get_ivar` were combined like I suggested in my other comment, this call could be simplified to

```suggestion
            old_value = get_ivar(objc_self, '_' + attr_name, weak=self.weak)
```

The same would apply to the `set_weak_ivar`/`set_ivar` calls below.

rubicon/objc/runtime.py
```diff
@@ -383,6 +387,18 @@ class objc_property_attribute_t(Structure):
 libobjc.objc_getProtocol.restype = objc_id
 libobjc.objc_getProtocol.argtypes = [c_char_p]
 
+# id objc_initWeak(id *object, id value)
+libobjc.objc_initWeak.restype = objc_id
+libobjc.objc_initWeak.argtypes = [POINTER(objc_id), objc_id]
```
dgelessus, line 392: Minor thing: this function is never used, so we don't need to declare it.

dgelessus:
re. the crashing tests on Python 3.5: We've already had a similar situation before in #201, where everything ran fine on newer Python versions, and only Python 3.5 sometimes crashed while running the tests. In that case, it turned out that we had incorrect reference management code in the tests (some test objects were released twice), which for some reason only caused visible crashes on Python 3.5.

I would guess that the crash we're seeing here is similar - there's probably some sort of memory management bug in the new code, and Python 3.5 is the only version where we're noticing it. Maybe it's because [we haven't added handling for weak properties in `dealloc` yet](https://github.com/beeware/rubicon-objc/pull/211#discussion_r661046717)?

It's of course also possible that the crash comes from some difference between Python 3.5 and later versions, but I don't know what difference specifically could cause it. It's probably not related to dict ordering, because the code added/changed in this PR doesn't really use dicts.

About Python 3.5 support in general - I'm not strongly attached to 3.5, and 3.6 has some nice features like f-strings, so in general I would be fine with dropping Python 3.5 support. But we should do that *after* we've figured out where the crashes here are coming from. Even if the crashes only happen on 3.5, the underlying issue could apply to later versions too.

freakboy3742:
@SamSchott As best as I can make out, it is `tests/test_core.py::RubiconTest::tests/test_core.py::RubiconTest::test_class_properties_lifecycle_strong` that is failing. No stacktrace, unfortunately; just a segfault. Working through the test, it's the final `properties.object` attribute access on line 1014 that is causing the segfault. 

Digging deeper than that, it's the call to `object_isClass(object_ptr)` (api.py, line 655):
<details>
  <summary>Full pdb trace</summary>

    tests/test_core.py::RubiconTest::test_class_properties_lifecycle_strong 
    >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
    > /Users/rkm/projects/beeware/rubicon/objc/tests/test_core.py(1015)test_class_properties_lifecycle_strong()
    -> properties.object
    (Pdb) s
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(744)__getattr__()
    -> def __getattr__(self, name):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(797)__getattr__()
    -> if not name.endswith('_'):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(798)__getattr__()
    -> method = self.objc_class._cache_property_accessor(name)
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(597)objc_class()
    -> @property
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(603)objc_class()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(604)objc_class()
    -> return super(ObjCInstance, type(self)).__getattribute__(self, "_objc_class")
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(604)objc_class()-><rubicon.objc...x7fa5d157a140>
    -> return super(ObjCInstance, type(self)).__getattribute__(self, "_objc_class")
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(1142)_cache_property_accessor()
    -> def _cache_property_accessor(self, name):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(1147)_cache_property_accessor()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(1148)_cache_property_accessor()
    -> methods = self.instance_properties[name]
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(1152)_cache_property_accessor()
    -> if methods:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(1153)_cache_property_accessor()
    -> return methods[0]
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(1153)_cache_property_accessor()-><ObjCMethod: b'object' b'@@:'>
    -> return methods[0]
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(799)__getattr__()
    -> if method:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(800)__getattr__()
    -> return ObjCBoundMethod(method, self)()
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(217)__init__()
    -> def __init__(self, method, receiver):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(219)__init__()
    -> self.method = method
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(220)__init__()
    -> if type(receiver) == Class:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(223)__init__()
    -> self.receiver = receiver
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(223)__init__()->None
    -> self.receiver = receiver
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(229)__call__()
    -> def __call__(self, *args, **kwargs):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(231)__call__()
    -> return self.method(self.receiver, *args, **kwargs)
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(97)__call__()
    -> def __call__(self, receiver, *args, convert_args=True, convert_result=True):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(120)__call__()
    -> if len(args) != len(self.method_argtypes):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(126)__call__()
    -> if convert_args:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(127)__call__()
    -> converted_args = []
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(128)__call__()
    -> for argtype, arg in zip(self.method_argtypes, args):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(156)__call__()
    -> result = send_message(
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(157)__call__()
    -> receiver, self.selector, *converted_args,
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(158)__call__()
    -> restype=self.restype, argtypes=self.method_argtypes,
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(679)send_message()
    -> def send_message(receiver, selector, *args, restype, argtypes, varargs=None):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(720)send_message()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(721)send_message()
    -> receiver = receiver._as_parameter_
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(725)send_message()
    -> if not isinstance(receiver, objc_id):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(731)send_message()
    -> if not isinstance(selector, SEL):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(734)send_message()
    -> if len(args) != len(argtypes):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(740)send_message()
    -> if varargs is None:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(741)send_message()
    -> varargs = []
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(743)send_message()
    -> send = _msg_send_for_types(restype, argtypes)
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(641)_msg_send_for_types()
    -> def _msg_send_for_types(restype, argtypes):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(651)_msg_send_for_types()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(653)_msg_send_for_types()
    -> return _msg_send_cache[(restype, *argtypes)]
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(653)_msg_send_for_types()-><_FuncPtr obj...t 0x1031b79a8>
    -> return _msg_send_cache[(restype, *argtypes)]
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(745)send_message()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(746)send_message()
    -> result = send(receiver, selector, *args, *varargs)
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(123)__new__()
    -> def __new__(cls, init=None):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(128)__new__()
    -> if isinstance(init, (bytes, str)):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(133)__new__()
    -> self = super().__new__(cls, init)
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(134)__new__()
    -> self._inited = False
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(135)__new__()
    -> return self
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(135)__new__()->rubicon.objc.runtime.SEL(None)
    -> return self
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(137)__init__()
    -> def __init__(self, init=None):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(138)__init__()
    -> if not self._inited:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(139)__init__()
    -> super().__init__(init)
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(139)__init__()->None
    -> super().__init__(init)
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(395)_objc_getter()
    -> def _objc_getter(objc_self, _cmd):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(396)_objc_getter()
    -> if self.weak:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(399)_objc_getter()
    -> value = get_ivar(objc_self, '_' + attr_name)
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(951)get_ivar()
    -> def get_ivar(obj, varname):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(964)get_ivar()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(965)get_ivar()
    -> obj = obj._as_parameter_
    (Pdb) 
    AttributeError: 'objc_id' object has no attribute '_as_parameter_'
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(965)get_ivar()
    -> obj = obj._as_parameter_
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(966)get_ivar()
    -> except AttributeError:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(967)get_ivar()
    -> pass
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(969)get_ivar()
    -> ivar = libobjc.class_getInstanceVariable(libobjc.object_getClass(obj), ensure_bytes(varname))
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(567)ensure_bytes()
    -> def ensure_bytes(x):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(573)ensure_bytes()
    -> if isinstance(x, bytes):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(577)ensure_bytes()
    -> return x.encode('utf-8')
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(577)ensure_bytes()->b'_object'
    -> return x.encode('utf-8')
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(970)get_ivar()
    -> vartype = ctype_for_encoding(libobjc.ivar_getTypeEncoding(ivar))
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/types.py(296)ctype_for_encoding()
    -> def ctype_for_encoding(encoding):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/types.py(314)ctype_for_encoding()
    -> encoding = encoding.lstrip(b"NORVnor")
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/types.py(316)ctype_for_encoding()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/types.py(318)ctype_for_encoding()
    -> return _ctype_for_encoding_map[encoding]
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/types.py(318)ctype_for_encoding()-><class 'rubic...time.objc_id'>
    -> return _ctype_for_encoding_map[encoding]
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(972)get_ivar()
    -> if isinstance(vartype, objc_id):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(975)get_ivar()
    -> return vartype.from_address(obj.value + libobjc.ivar_getOffset(ivar))
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(975)get_ivar()-><objc_id obje...t 0x1030097b8>
    -> return vartype.from_address(obj.value + libobjc.ivar_getOffset(ivar))
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(403)_objc_getter()
    -> if not isinstance(value, (Structure, Union)):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(404)_objc_getter()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(405)_objc_getter()
    -> value = value.value
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(409)_objc_getter()
    -> return value
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(409)_objc_getter()->140350158314784
    -> return value
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(756)send_message()
    -> if restype == c_void_p:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(758)send_message()
    -> return result
    (Pdb) 
    --Return--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/runtime.py(758)send_message()-><objc_id obje...t 0x10300b950>
    -> return result
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(161)__call__()
    -> if not convert_result:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(165)__call__()
    -> if self.restype is not None and issubclass(self.restype, objc_id):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(166)__call__()
    -> result = ObjCInstance(result)
    (Pdb) 
    --Call--
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(620)__new__()
    -> def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(640)__new__()
    -> if not isinstance(object_ptr, objc_id):
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(644)__new__()
    -> if not object_ptr.value:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(649)__new__()
    -> try:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(650)__new__()
    -> return cls._cached_objects[object_ptr.value]
    (Pdb) 
    --Call--
    > /Users/rkm/.pyenv/versions/3.5.9/lib/python3.5/weakref.py(134)__getitem__()
    -> def __getitem__(self, key):
    (Pdb) 
    > /Users/rkm/.pyenv/versions/3.5.9/lib/python3.5/weakref.py(135)__getitem__()
    -> if self._pending_removals:
    (Pdb) 
    > /Users/rkm/.pyenv/versions/3.5.9/lib/python3.5/weakref.py(137)__getitem__()
    -> o = self.data[key]()
    (Pdb) 
    KeyError: 140350158314784
    > /Users/rkm/.pyenv/versions/3.5.9/lib/python3.5/weakref.py(137)__getitem__()
    -> o = self.data[key]()
    (Pdb) 
    --Return--
    > /Users/rkm/.pyenv/versions/3.5.9/lib/python3.5/weakref.py(137)__getitem__()->None
    -> o = self.data[key]()
    (Pdb) 
    KeyError: 140350158314784
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(650)__new__()
    -> return cls._cached_objects[object_ptr.value]
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(651)__new__()
    -> except KeyError:
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(652)__new__()
    -> pass
    (Pdb) 
    > /Users/rkm/projects/beeware/rubicon/objc/rubicon/objc/api.py(655)__new__()
    -> if not issubclass(cls, ObjCClass) and object_isClass(object_ptr):
    (Pdb) 
    zsh: segmentation fault  pytest tests/test_core.py::RubiconTest::test_class_properties_lifecycle_stron

</summary>

rubicon/objc/api.py
```diff
@@ -426,6 +448,8 @@ def _objc_setter(objc_self, _cmd, new_value):
         attrs = self._get_property_attributes()
         libobjc.class_addProperty(class_ptr, ensure_bytes(attr_name), attrs, len(attrs))
 
+        # register dealloc
+
```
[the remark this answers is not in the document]
SamSchott: Oh, yes, but modifying dealloc to `release` property values is probably best left for a separate PR. I'll remove the comment for now.

rubicon/objc/runtime.py
SamSchott: True. It's a leftover when I was trying to figure out LLVMs generated calls.

rubicon/objc/runtime.py
SamSchott: Yes, I think this makes sense.

SamSchott:
> Maybe it's because we haven't added handling for weak properties in dealloc yet?

Hm, I don't think so. Weak properties should not need any handling in dealloc, right? It's the strong properties where we need to explicitly call `release` during dealloc, if I understand things correctly.

@freakboy3742, thanks for the investigation, this is very helpful. I can reproduce the segfault when adding a breakpoint just before accessing the `properties.object` attribute in `test_class_properties_lifecycle_strong`. So it's definitely a memory management bug somewhere.

SamSchott:
Interestingly this (new) test also segfaults with the current master branch. It's likely is the manifestation of an existing issue, maybe introduced in #201.

SamSchott:
I've found the issue and its protracted. After adding the following debug code to the setter:

```python
        def _objc_setter(objc_self, _cmd, new_value):

            if not isinstance(new_value, self.vartype):
                # If vartype is a primitive, then new_value may be unboxed. If that is the case, box it manually.
                new_value = self.vartype(new_value)

            old_value = get_ivar(objc_self, '_' + attr_name, weak=self.weak)

            print("old_value", old_value.value)
            print("new_value", new_value.value)

            if new_value is old_value:
                return

            if issubclass(self.vartype, objc_id) and new_value and not self.weak:
                # If the new value is a non-null object, retain it.
                print("retaining", new_value.value)
                send_message(new_value, 'retain', restype=objc_id, argtypes=[])

            set_ivar(objc_self, '_' + attr_name, new_value, weak=self.weak)

            if issubclass(self.vartype, objc_id) and old_value and not self.weak:
                # If the old value is a non-null object, release it.
                print("releasing", old_value.value)
                send_message(old_value, 'release', restype=None, argtypes=[])
```

the test `test_class_properties_lifecycle_strong` prints the following to stdout:

```
old_value None
new_value 140575283530480
retaining 140575283530480
releasing 140575283530480
```  

So while setting the property value, it is both retained and released. I suspect this happens because the first access with `get_ivar()` returns `None` and per your documentation of `get_ivar()`:

> For non-object types (everything except :class:`objc_id` and subclasses), the returned data object is backed by the
    ivar's actual memory. This means that the data object is only usable as long as the "owner" object is alive, and
    writes to it will directly change the ivar's value.

Setting the new value therefore directly changes `old_value` and the check `issubclass(self.vartype, objc_id) and old_value` evaluates to `True`, triggering a release.

rubicon/objc/runtime.py
```diff
@@ -909,14 +921,17 @@ def get_ivar(obj, varname):
     ivar = libobjc.class_getInstanceVariable(libobjc.object_getClass(obj), ensure_bytes(varname))
     vartype = ctype_for_encoding(libobjc.ivar_getTypeEncoding(ivar))
 
-    if isinstance(vartype, objc_id):
+    if weak:
+        value = libobjc.objc_loadWeakRetained(obj.value + libobjc.ivar_getOffset(ivar))
+        return libobjc.objc_autoreleaseReturnValue(value)
+    elif isinstance(vartype, objc_id):
         return cast(libobjc.object_getIvar(obj, ivar), vartype)
```
SamSchott, line 928: When is this cast required? Isn't it sufficient to to return a `objc_id` which later gets converted to the correct `ObjCInstance`?

If it is required, do we need add a similar cast before returning the value in the weak case?

dgelessus:
> Hm, I don't think so. Weak properties should not need any handling in dealloc, right? It's the strong properties where we need to explicitly call `release` during dealloc, if I understand things correctly.

Yes, strong properties should also be handled in `dealloc`, but that's less of a problem. Not releasing the values of strong properties "just" causes a memory leak - it doesn't lead to crashes. Weak properties on the other hand *could* cause segfaults if not correctly cleaned up, I think, because a weak pointer gets set to `nil` if the object it points to is deallocated. So if something like this happens:

* Object A (which has a weak property) is allocated
* Object B is allocated
* Object A's weak property is set to point to object B
* Object A is deallocated, but its weak property is not cleaned up
* Object B is deallocated

the runtime will set Object A's weak property to `nil`, but Object A has already been deallocated, so some other memory gets zeroed instead.

But as you've already figured out (thank you for the debugging!) the segfault we're seeing is caused by something else.

> Shouldn't `get_ivar()` return the value stored in the in the ivar instead of the ivar's actual memory in case of a non-null `objc_id`?

Yes, in fact this should always be the case, no matter if the ivar is currently `nil` or not. I'm honestly really confused why you're getting different behavior. If the ivar type is `objc_id`, this branch of `get_ivar` should execute:

```python
    if isinstance(vartype, objc_id):
        return cast(libobjc.object_getIvar(obj, ivar), vartype)
```

and `libobjc.object_getIvar` returns the ivar's value (i. e. the address stored *in* the ivar, not the address *of* the ivar) as an integer, which is then cast to `objc_id`. So the returned object should always be a fresh `objc_id` object that isn't coupled to any existing memory address.

rubicon/objc/runtime.py
dgelessus: You're right that `libobjc.object_getIvar` already returns an `objc_id`. The most common case is that `vartype` is exactly `objc_id`, in which case the cast indeed does nothing. I think it's only needed for the less common case where `vartype` is `Class` (which we have declared as a subclass of `objc_id`) so that the return value is cast to `Class` instead of a plain `objc_id`.

SamSchott:
> Weak properties on the other hand could cause segfaults if not correctly cleaned up

Oh, I had not thought of that. I'll need to check. Incidentally, what is a good way of overriding `dealloc` at runtime? Could one use `method_setImplementation` for this?

> If the ivar type is objc_id, this branch of get_ivar should execute:

I think the problem lies with `isinstance(vartype, objc_id)`. `vartype` is a class and not an instance so the check should probably be replaced with `issubclass`.

SamSchott:
You are right about the cleanup of weak ivars. Testing the cycle which you have outlined prints the following warning:

> __weak variable at 0x7fed284f8e00 holds 0x800007fed267083b instead of 0x7fed26689610. This is probably incorrect use of objc_storeWeak() and objc_loadWeak(). Break on objc_weak_error to debug.

rubicon/objc/runtime.py
SamSchott: Ah, ok! So we should perform a similar cast when returning a value in the weak case as well.

rubicon/objc/api.py
```diff
@@ -403,29 +413,61 @@ def _objc_setter(objc_self, _cmd, new_value):
...
+                # If the old value is a non-null object, release it. There is no need to set the actual ivar to nil.
+                old_value = get_ivar(objc_self, ivar_name, weak=self.weak)
+                send_message(old_value, 'release', restype=None, argtypes=[])
+
+            # Invoke original dealloc.
+            cfunctype = CFUNCTYPE(None, objc_id, SEL)
+            old_dealloc_callable = cast(old_dealloc, cfunctype)
+            old_dealloc_callable(objc_self, SEL("dealloc"))
+
+        replace_method(class_ptr, 'dealloc', _new_delloc, [None, ObjCInstance, SEL])
```
dgelessus, line 469: If there are multiple properties defined on one class, this is going to re-wrap `dealloc` once for each property, right? Although that *works*, IMO it would be better if we define `dealloc` only once and have it clean up all of the object's properties in a simple loop. This would be easier to read and debug, and probably also a bit faster, because it avoids calling back and forth many times between Python and Objective-C.

Though with the current `class_register` mechanism there's no way to implement this - `class_register` is called once for each property, and the call has no information about other properties in the class. So this would probably require adding an extra method to `objc_property` that implements deallocating a single property, and have `ObjCClass._new_from_class_statement` define a `dealloc` method that asks each property to deallocate itself.

rubicon/objc/api.py
```diff
@@ -403,29 +413,61 @@ def _objc_setter(objc_self, _cmd, new_value):
...
+
+            # Clean up ivar.
+            if self.weak:
+                # Clean up weak reference.
+                ivar = libobjc.class_getInstanceVariable(libobjc.object_getClass(objc_self), ivar_name.encode())
+                libobjc.objc_storeWeak(objc_self.value + libobjc.ivar_getOffset(ivar), None)
+            elif issubclass(self.vartype, objc_id):
+                # If the old value is a non-null object, release it. There is no need to set the actual ivar to nil.
+                old_value = get_ivar(objc_self, ivar_name, weak=self.weak)
+                send_message(old_value, 'release', restype=None, argtypes=[])
```
dgelessus, line 462: Could this be simplified somehow by calling `_objc_setter` or `set_ivar`? Especially in the weak case the code is a bit more complex, so it would be better to use our existing functions instead of reimplementing it.

rubicon/objc/api.py
SamSchott: True, it's not particularly elegant. I was already thinking of having a single dealloc method in `ObjCClass._new_from_class_statement` but was torn between having a simpler dealloc and keeping all the registration / reference management functionality contained in the `objc_property` class.

rubicon/objc/api.py
SamSchott: Sure.

SamSchott: I remember now why I did this. The current `set_ivar` implementation raises a TypeError when setting the value to `None`. This can be changed of course, it is common place in ObjC to set a value to nil.

rubicon/objc/runtime.py
```diff
@@ -846,8 +858,8 @@ def send_super(cls, receiver, selector, *args, restype=c_void_p, argtypes=None):
 _keep_alive_imps = []
 
 
-def add_method(cls, selector, method, encoding):
-    """Add a new instance method to the given class.
+def replace_method(cls, selector, method, encoding):
+    """Add a new instance method to the given class or replace an existing instance method.
```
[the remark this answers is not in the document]
SamSchott: Fair point. I can see the value of keeping `add_method` and raising an error when it fails,

rubicon/objc/api.py
```diff
@@ -968,6 +1000,24 @@ def _new_from_class_statement(cls, name, bases, attrs, *, protocols):
...
+        # Add cleanup of ivars / properties to dealloc
+
+        old_dealloc = libobjc.class_getMethodImplementation(ptr, SEL("dealloc"))
+
+        def _new_delloc(objc_self, _cmd):
+
+            # Invoke dealloc callback of each property.
+            for attr_name, obj in attrs.items():
+                if isinstance(obj, objc_property):
+                    obj.dealloc_callback(objc_self, attr_name)
```
dgelessus, line 1012: For consistency with the "register" code above, I would check for the existence of the `dealloc_callback` attribute instead of a specific class. This would also make it easier to add `dealloc_callback`s on other classes in the future (although I can't think of any good use for this right now).

```suggestion
                try:
                    dealloc_callback = obj.dealloc_callback
                except AttributeError:
                    pass
                else:
                    dealloc_callback(objc_self, attr_name)
```

rubicon/objc/api.py
```diff
@@ -968,6 +1000,24 @@ def _new_from_class_statement(cls, name, bases, attrs, *, protocols):
...
+
+            # Invoke original dealloc.
+            cfunctype = CFUNCTYPE(None, objc_id, SEL)
+            old_dealloc_callable = cast(old_dealloc, cfunctype)
+            old_dealloc_callable(objc_self, SEL("dealloc"))
+
+        add_method(ptr, "dealloc", _new_delloc, [None, ObjCInstance, SEL], replace=True)
```
dgelessus, line 1019: What behavior do we want if a class contains both a user-defined `dealloc` method and properties with dealloc callbacks? The implementation here will call the dealloc callbacks before the user-defined `dealloc` method. But in practice I think the opposite order would be more useful - that way the user-defined `dealloc` could call methods on objects stored in properties before they are released by the callbacks (which could cause them to be deallocated right away).

Implementing this would be a bit more difficult. A user-defined `dealloc` is expected to end with a call to the superclass `dealloc`. But if a class has dealloc callbacks, those would need before the superclass `dealloc` (but also after the user-defined `dealloc` code). There's no way for Rubicon to insert the dealloc callbacks into the user's `dealloc` method, so the user would have to manually call the dealloc callbacks before the super call (via an extra function/method provided by Rubicon).

A different solution would be to allow users to define their own dealloc callback that gets called before any dealloc callbacks from properties. Then users could use that callback instead of manually overriding `dealloc`, and let Rubicon generate a `dealloc` method that calls all callbacks in the right order.

SamSchott: I agree that calling the user's `dealloc` before clearing any instance variables can make life at lot easier for the user. But I don't really like either of the proposed solutions, both look like workarounds that force the user to become aware of the memory management which we are otherwise performing behind the scenes.

I can possibly think of two alternative approaches that could provide a better user experience (but would complicate our own implementation):

1. We document that user should not call the `dealloc` of the super class in their own implementation, as in a proper ARC environment. We then call the following methods in order: First, any user-defined dealloc, second, our cleanup code and third the super class's dealloc. This would mean treating the dealloc definition differently from other method definitions, at least internally. It would however be transparent to the user.

2. We call our own cleanup code *after* calling the old `delloc` implementation (which includes the user's code and any calls to the super dealloc). I'm not if this is possible to do reliably.

What do you think, is either of those options worth the effort?

rubicon/objc/api.py
```diff
@@ -368,26 +368,36 @@ class MySubclass(NSObject):
...
         if issubclass(self.vartype, objc_id):
-            attrs.append(objc_property_attribute_t(b'&', b''))  # retain
+            reference = b'W' if self.weak else b'&'
+            attrs.append(objc_property_attribute_t(reference, b''))
         return (objc_property_attribute_t * len(attrs))(*attrs)
 
     def class_register(self, class_ptr, attr_name):
-        add_ivar(class_ptr, '_' + attr_name, self.vartype)
+
+        ivar_name = '_' + attr_name
```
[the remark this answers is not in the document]
SamSchott: It would be an attribute which is not set during `__init__` but only after `class_register` or `protocol_register` is called, since the `objc_property` does not have a "name" independent of the class / protocol that it's registered with. So we'd only save a single call of composing the `ivar_name` in the `delloc_callback`. Not sure if it's worth it...

rubicon/objc/api.py
SamSchott: Incidentally, is there any reason we are not using `hasattr` instead? It looks a bit more elegant and is also used in `ObjCProtocol.__new__()`.

rubicon/objc/api.py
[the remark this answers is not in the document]
dgelessus: Ah, you're right - I overlooked that `attr_name` is a parameter of both methods and so can't be set in `__init__`. Ignore my comment then 🙂

rubicon/objc/api.py
dgelessus: The point of using `try`/`catch AttributeError` is that it accesses the attribute only once and checks that it exists at the same time as getting its value. This matters mainly when you're dealing with multithreading and the attribute might be deleted in between the `hasattr` check and the value being read. That can't happen here though, so in this case there would be no problem with using the shorter `hasattr` version instead of a `try`/`catch`.

rubicon/objc/api.py
dgelessus: Option 1 would be ideal, because it's simple to use and matches what you do in normal ARC Objective-C code. But implementing this behavior now would silently break existing code that overrides `dealloc` and correctly calls the superclass `dealloc` at the end - that would cause the superclass `dealloc` to get called twice, which can lead to other objects being `release`d too often and other similar problems. That's the main reason why I suggested adding a separate user-definable callback for the same purpose.

Rubicon is still before version 1.0, so we *could* still make breaking changes like this, especially if it improves usability in the long term. But if we do that, we should try to throw errors for code relying on the old behavior, to avoid silent double frees in code that was previously correct.

I honestly don't know if option 2 would be safe or not... There *should* be no way for the superclass `dealloc`s to corrupt ivars defined by the subclass, but I wouldn't really rely on it. Especially `NSObject`'s top-level `dealloc` could do things that make the entire object unusable somehow.

Another alternative would be to not touch user-defined `dealloc`s at all, and only generate an automatic `dealloc` if the user hasn't already overridden `dealloc` manually. This would be fully compatible with existing code and should also be simple to implement. The disadvantage is that if the user really needs to add custom code to `dealloc`, they then have to manually do all the cleanup that the dealloc callbacks would have done automatically.

SamSchott: I do prefer Option 1 from those choices, or not touching use-defined deallocs at all together with a clear documentation on manual cleanup. My only issue with Option 1 is, how can we raise an error if user does call the superclass dealloc? Is there an elegant way of doing so? We are not a compiler after all and don't want to inspect the actual code in the user's dealloc.

SamSchott: I can confirm that running our cleanup *after* the superclass dealloc does lead to trouble during the ivar cleanup. In particular, `libobjc.class_getInstanceVariable()` returns `None`. So this is not viable.

I've implemented the first option now: calling the user's dealloc, then our cleanup, then the superclass dealloc. For the time being, there is no special error handling if the user calls the superclass dealloc manually. It will however raise (obscure) errors when we run our own cleanup. Without our own cleanup code (for example when no properties are declared), dealloc is called twice and leads to a segfault. In either case, users will notice that *something* is wrong without knowing what it might be. Not ideal...

SamSchott: The best workaround I can come up with, short of actually inspecting the user's dealloc code when creating the class, is to print a warning when `send_super` is called, before either segfaulting or failing to complete the dealloc. What do you think, is this acceptable?

dgelessus: Sorry for the late reply - had some uni stuff that I needed to finish in the last few days.

The implementation you wrote looks good IMO. You're right that we can't do much to guard against old code that still calls `send_super` at the end of `dealloc`. The only option I can think of is what you've already suggested - add a special case inside `send_super` that warns whenever `send_super(self, "dealloc", ...)` is called. If we do that, we just need to make sure that the warning doesn't appear when Rubicon itself calls `send_super(self, "dealloc", ...)` - probably using an internal keyword to suppress the warning?

That way I think we could even "fix" the segfaults. If `send_super(self, "dealloc", ...)` is called *without* the special internal keyword argument, we can make it show a warning and then return without actually calling the super method. That way, if a user-defined `dealloc` calls `send_super`, the super `dealloc` won't actually be called yet. Then the user-defined `dealloc` returns to Rubicon, which runs the dealloc callbacks then makes the real `send_super` call (with the internal keyword argument, so that this time it actually calls the super method).

This isn't a very nice solution - but if it works, I would rather have some less nice code in Rubicon that shows a helpful warning about the change, rather than breaking existing code so that it causes unexplained segfaults.

tests/test_core.py
```diff
@@ -1318,3 +1359,32 @@ def test_objcinstance_retain_release(self):
...
+        obj.__dict__["_did_dealloc"] = False
+
+        attr0 = NSObject.alloc().init()
+        attr1 = NSObject.alloc().init()
+
+        obj.attr0 = attr0
+        obj.attr1 = attr1
+
+        assert attr0.retainCount() == 2
+        assert attr1.retainCount() == 1
```
dgelessus, line 1383: Only noticed this now - our tests use plain standard `unittest` and not PyTest, so these asserts should be written using `self.assertEqual` and not the `assert` keyword.

tests/test_core.py
```diff
@@ -1318,3 +1359,32 @@ def test_objcinstance_retain_release(self):
...
+
+        # ObjC object will be deallocated, can only access Python attributes afterwards.
+        obj.release()
+
+        assert obj._did_dealloc, "custom dealloc did not run"
+        assert attr0.retainCount() == 1, "strong property value was not released"
+        assert attr1.retainCount() == 1, "weak property value was released"
```
dgelessus, line 1390: As above, this should use `self.assertEqual` and `self.assertTrue`.

dgelessus (commented):
Can confirm that the `send_super(..., "dealloc", ...)` warning works as expected. A couple of small things, then we are really done I think 🙂

tests/test_core.py
SamSchott: Oops. Yes, now that you say that...

rubicon/objc/runtime.py
```diff
@@ -799,6 +815,13 @@ def send_super(cls, receiver, selector, *args, restype=c_void_p, argtypes=None):
             .format(tp=type(cls))
         )
 
+    if not _allow_dealloc and selector.name == b"dealloc":
+        warnings.warn(
+            "You should not call the superclass dealloc manually when overriding dealloc. Rubicon-objc "
+            "will call it for you after releasing objects stored in properties and ivars."
```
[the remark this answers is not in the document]
SamSchott: Good point

SamSchott:
What do you think about the expanded docs on memory management? Is the new section on reference cycles helpful? I was trying to balance being brief and giving all the necessary information + code snippets.

dgelessus:
Yes, the new docs are definitely a useful addition! That way people running into reference cycle problems can find out about weak properties more easily. I don't think the docs are too long or detailed either.

dgelessus (approved):
:shipit:

PR #213, opened 2021-07-30
Autoreleasepool context manager

### LINKED ISSUE #210: High-level API for autorelease pools (`@autoreleasepool`, `NSAutoreleasePool`)
(Mostly copied from https://github.com/beeware/rubicon-objc/pull/209#discussion_r635685910)

Rubicon should provide a proper API for autorelease pools, probably as a context manager (`with objc.autoreleasepool(): ...`).

The traditional way to create and destroy autorelease pools is by creating and releasing [`NSAutoreleasePool`](https://developer.apple.com/documentation/foundation/nsautoreleasepool?language=objc) objects. However, in normal Objective-C code, `NSAutoreleasePool` is more or less superseded by `@autoreleasepool` blocks, and when using ARC, `@autoreleasepool` is the only option - see the warning in [the `NSAutoreleasePool` docs](https://developer.apple.com/documentation/foundation/nsautoreleasepool?language=objc). Internally, `@autoreleasepool` blocks are translated to calls to the Objective-C runtime functions `objc_autoreleasePoolPush` and `objc_autoreleasePoolPop`, which are undocumented and not declared in Apple's headers, but [seem to be widely known](https://stackoverflow.com/q/15014152).

For the moment, it's *probably* fine to use `NSAutoreleasePool` for managing autorelease pools from Rubicon - the class isn't officially deprecated, and we aren't affected by ARC restrictions. But for future-proofing it would be better to define our own high-level API for autorelease pools that uses the appropriate mechanism under the hood.

SamSchott:
This PR adds an `autoreleasepool` context manager to `rubicon.obj.runtime` which mimics Objective-C's `@autoreleasepool` blocks. Closes #210.

## PR Checklist:

### DIFF
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -1,5 +1,6 @@
 import os
 import warnings
+from contextlib import contextmanager
 from ctypes import (
     ArgumentError, CDLL, CFUNCTYPE, POINTER, Structure, Union, addressof, alignment, byref, c_bool, c_char_p,
     c_double, c_float, c_int, c_longdouble, c_size_t, c_uint, c_uint8, c_void_p, cast, memmove, sizeof, util,
@@ -992,7 +993,8 @@ def set_ivar(obj, varname, value, weak=False):
         memmove(obj.value + libobjc.ivar_getOffset(ivar), addressof(value), sizeof(vartype))
 
 
-class autoreleasepool:
+@contextmanager
+def autoreleasepool():
     """
     A context manager that has the same effect as a @autoreleasepool block in Objective-C.
 
@@ -1002,11 +1004,9 @@ class autoreleasepool:
     they may be still be useful when your code temporarily allocates large amounts of memory which you want to
     explicitly free before the end of a cycle.
     """
-    def __init__(self):
-        self.pool = None
+    pool = libobjc.objc_autoreleasePoolPush()
 
-    def __enter__(self):
-        self.pool = libobjc.objc_autoreleasePoolPush()
-
-    def __exit__(self, type, value, traceback):
-        libobjc.objc_autoreleasePoolPop(self.pool)
+    try:
+        yield
+    finally:
+        libobjc.objc_autoreleasePoolPop(pool)
--- a/rubicon/objc/runtime.py
+++ b/rubicon/objc/runtime.py
@@ -20,6 +20,7 @@ __all__ = [
     'SEL',
     'add_ivar',
     'add_method',
+    'autoreleasepool',
     'get_class',
     'get_ivar',
     'libc',
@@ -361,6 +362,14 @@ libobjc.method_setImplementation.argtypes = [Method, IMP]
 # Class objc_allocateClassPair(Class superclass, const char *name, size_t extraBytes)
 libobjc.objc_allocateClassPair.restype = Class
 libobjc.objc_allocateClassPair.argtypes = [Class, c_char_p, c_size_t]
+
+# void *objc_autoreleasePoolPush(void)
+libobjc.objc_autoreleasePoolPush.restype = c_void_p
+libobjc.objc_autoreleasePoolPush.argtypes = []
+
+# void objc_autoreleasePoolPop(void *pool)
+libobjc.objc_autoreleasePoolPop.restype = None
+libobjc.objc_autoreleasePoolPop.argtypes = [c_void_p]
 
 # id objc_autoreleaseReturnValue(id value)
 libobjc.objc_autoreleaseReturnValue.restype = objc_id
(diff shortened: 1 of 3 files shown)


rubicon/objc/runtime.py
```diff
@@ -981,3 +990,23 @@ def set_ivar(obj, varname, value, weak=False):
...
+    A context manager that has the same effect as a @autoreleasepool block in Objective-C.
+
+    Any objects that are autoreleased within the context will receive a release message when exiting the context. When
+    running an event loop, AppKit will create an autorelease pool at the beginning of each cycle of the event loop and
+    drain it at the end. You therefore do not need to use @autoreleasepool blocks when running an event loop. However,
+    they may be still be useful when your code temporarily allocates large amounts of memory which you want to
+    explicitly free before the end of a cycle.
+    """
+    def __init__(self):
+        self.pool = None
```
dgelessus, line 1006: Minor thing - there's no need to initialize `self.pool` here, the assignment in `__enter__` is enough. That also guards better against accidentally using the attribute before it's set - trying to use a nonexistant attribute gives an exception, whereas a `None` attribute could be silently passed to the C side as a `NULL` pointer and cause a segfault.

rubicon/objc/runtime.py
```diff
@@ -981,3 +990,23 @@ def set_ivar(obj, varname, value, weak=False):
...
+
+    def __enter__(self):
+        self.pool = libobjc.objc_autoreleasePoolPush()
```
dgelessus, line 1009: Speaking of error checking, it might be good to raise an exception here if `self.pool` has already been set. This would guard against reusing the same `autoreleasepool` object in more than one `with` block. (This is very unlikely to happen by accident, but better to check anyway, to avoid weird runtime errors.)

tests/test_core.py
```diff
@@ -1008,18 +1005,17 @@ def test_class_properties_lifecycle_strong(self):
...
+            obj_pointer = obj.ptr.value  # store the object pointer for future use
 
-        properties.object = obj
+            properties.object = obj
 
-        del obj
-        del pool
-        gc.collect()
+            del obj
+            gc.collect()
```
dgelessus, line 1018: If you want to match the behavior of the previous code, the `gc.collect()` call needs to be outside of the `with` block, so that it runs *after* the autorelease pool has been popped:

```suggestion

        gc.collect()
```

tests/test_core.py
```diff
@@ -1029,18 +1025,17 @@ def test_class_properties_lifecycle_weak(self):
...
+            properties.object = obj
 
-        self.assertIs(properties.object, obj)
+            self.assertIs(properties.object, obj)
 
-        del obj
-        del pool
-        gc.collect()
+            del obj
+            gc.collect()
```
dgelessus, line 1038:
```suggestion

        gc.collect()
```

tests/test_core.py
```diff
@@ -1264,38 +1259,34 @@ def test_objcinstance_python_attribute_keep_alive(self):
...
 
-        del obj
-        del pool
+            # Check that our Python attributes are still there.
+            self.assertIs(obj.python_object, wr_python_object())
 
-        gc.collect()
+            # Make sure that all Python objects are freed.
+            del obj
+            gc.collect()
```
dgelessus, line 1289:
```suggestion

        gc.collect()
```

dgelessus (commented):
Thanks for implementing this! Looks good overall, as usual 🙂 A few minor comments/suggestions.

tests/test_core.py
SamSchott: Yes, this change was deliberate. The idea is to delete the Python object first, then trigger a garbage collection to make sure that its `__del__` method is called, then clear the autoreleasepool to clean up any autoreleased objects.

The order does not really matter unless one of the Objective-C objects is kept alive by a Python reference *and* was autoreleased previously. The fact that the tests passed before this change means that the garbage collection itself probably is not required at all.

rubicon/objc/runtime.py
SamSchott: Those are both good points. I am considering rewriting the context manager as function decorated with `@contextmanager`. This will automatically ensure matching `__enter__()` and `__exit__()` calls without manual checks, however with some amount of "magic" going on behind the scenes and somewhat enigmatic error messages when the context manager is used improperly.

dgelessus: Yeah, using `@contextlib.contextmanager` might be a good idea if it checks this automatically. The less precise error messages aren't a big problem I think, because this error can only show up if you write really unusual code (you'd have to store the `autoreleasepool` object somewhere rather than using the normal `with autoreleasepool(): ...` usage pattern). If it ends up being a problem for some reason, we can always switch back to a manual context manager later.

tests/test_core.py
dgelessus: Thinking about it again, the logic behind the previous code was probably that `gc.collect()` comes after all of the `del` calls, to try to force Python to deallocate the `del`ed objects right away - especially the autorelease pool, which only got released once its `ObjCInstance` was deallocated on the Python side. Now that the autorelease pool is explicitly released by the context manager, this is no longer a problem, so the `gc.collect()` call is probably safe to go inside the context manager as you say. (I wouldn't remove the `gc.collect()` completely, because there's always the possibility that a simple `del` might not deallocate the object right away for some reason.)

rubicon/objc/runtime.py
SamSchott: Done.

tests/test_core.py
SamSchott: Right, that makes sense. The order did matter previously, because for the autorelease pool to work its Python reference needed to be deallocated, but this is no longer the case.

SamSchott: Right, that makes sense. The order did matter previously, because for the autorelease pool to work its Python reference needed to be deallocated, but this is no longer the case.

dgelessus (approved):

PR #215, opened 2021-08-16
Enforce object type for weak properties

SamSchott:
This PR enforces that the vartype of weak properties must be map to `objc_id` or `py_object`, i.e., that the property has an object type. This is because creating and storing weak reference in Objective-C is only supported for object types.

If a weak property is declared with a primitive C type or anything that maps to a primitive ctype, setting the value will succeed without an error but retrieving the value will return something entirely different.

## PR Checklist:

### DIFF
--- a/rubicon/objc/api.py
+++ b/rubicon/objc/api.py
@@ -394,7 +394,10 @@ class objc_property(object):
         self._ivar_weak = self.weak and not self._is_py_object
 
         if self.weak and not (self._is_py_object or self._is_objc_object):
-            raise TypeError("Weak properties are only supported for Objective-C or Python object types")
+            raise TypeError(
+                "Incompatible type for ivar {!r}: Weak properties are only supported "
+                "for Objective-C or Python object types".format(vartype)
+            )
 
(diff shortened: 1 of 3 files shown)


rubicon/objc/api.py
```diff
@@ -393,6 +393,9 @@ def __init__(self, vartype=objc_id, weak=False):
         # Check here if we need a weak or strong ivar.
         self._ivar_weak = self.weak and not self._is_py_object
 
+        if self.weak and not (self._is_py_object or self._is_objc_object):
+            raise TypeError("Weak properties are only supported for Objective-C or Python object types")
```
dgelessus, line 397: It might be useful to include the incorrect `vartype` in the error message. Most `TypeError`s in Python itself do this, and it makes debugging a little bit easier.

dgelessus (commented):
Makes sense. Nicely spotted, and thanks for fixing it!

rubicon/objc/api.py
SamSchott: Done.

dgelessus (approved):

PR #220, opened 2022-02-15
`send_super` fixes

### LINKED ISSUE #219: Enforce `restype` and `argtypes` for `send_super`
**Is your feature request related to a problem? Please describe.**
`rubicon.objc.runtime.send_super` doesn't check its arguments very much: it's possible to pass arguments without declaring their `argtypes`, in which case the arguments are assumed to be C-style varargs, with default types assigned by `ctypes`.

This is problematic because the assumed default argument types don't always match what is required/intended (e. g. integer sizes). On some architectures like AArch64 (iOS and ARM Macs), fixed and variable arguments are also passed in different incompatible ways (registers vs. stack), so assuming that all arguments are varargs doesn't work in most cases.

**Describe the solution you'd like**
For `send_message`, we have already introduced some stricter checks (see #174), so we should implement the same checks for `send_super` as well. This is a backwards-incompatible change - any existing code that calls `send_super` and doesn't already set `restype`/`argtypes` will have to be updated.

**Describe alternatives you've considered**
Instead of throwing a hard error, we could just show warnings if `restype`/`argtypes` are missing/mismatched. This would allow existing code to keep working on Intel and give callers more time to update their code. ARM would be broken either way though - if we don't throw an error, most affected calls will not have their arguments passed correctly, and will probably crash.

**Additional context**
The confusion of fixed and variable arguments has already caused some crashes in Toga on ARM Macs. This has been fixed now (beeware/toga#1406, beeware/toga#1411), but adding these checks in `send_super` would help detect similar problematic calls.

samschott:
This PR fixes #219.

It also aims to provide a uniform API between `send_message` and `send_super`. For `send_super`, `argtypes` is a keyword argument which defaults to an empty list. This can be convenient if there are no `args` provided to the call in the first place. I've therefore adapted this for `send_message` as well. Alternatively, one might make `argtypes` a mandatory argument for `send_super`. In any case, I would argue that a uniform API would be nice here.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Set up Python
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: "3.X"
     - name: Install dependencies
@@ -39,7 +39,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Set up Python 3.5
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.5
     - name: Install dependencies
@@ -58,7 +58,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Set up Python 3.7
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: 3.7
     - name: Install dependencies
@@ -81,7 +81,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Set up Python
-      uses: actions/setup-python@v1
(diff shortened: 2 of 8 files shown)


freakboy3742 (changes requested):
Patch looks good; 1 question, and 1 addition request.

The question: The updated version on the setup-python action - is that just a case of using the most recent version, or is there something specific being exercised for which we need the upgrade?

The request: Test cases. We've got a test case for incorrect argument lists for `send_message`; we should add a similar test set for `send_super`.

samschott:
> The updated version on the setup-python action - is that just a case of using the most recent version, or is there something specific being exercised for which we need the upgrade?

The smoke test on Python 3.5 was failing because Python 3.5 no longer comes pre-installed on GitHub test runners. setup-python@v2 allows downloading Python versions which are not preinstalled. The alternative would have been to migrate the test to Python 3.6 but I've found that in the past, tests on Python 3.5 have caught some allocation / deallocation related segfaults which appeared to be more flaky in Python > 3.5.

> Test cases. We've got a test case for incorrect argument lists for send_message; we should add a similar test set for send_super.

Fair point, will do.

freakboy3742:
Having access to a broader range of Python versions will definitely be useful. However, Python 3.5 is now officially EOL, so IMHO we don't need to go to excessive lengths to support it. While it may provide utility in surfacing memory allocation errors, using 3.5 won't be a good strategy for the long term, so if we have concerns, we should find some other way (like valgrind) to surface those errors.

samschott:
I agree, there is no strong case to support it. For now, it just seemed painless to keep testing on Python 3.5 as long as we don't want or need to drop support for any code changes.

freakboy3742 (approved):
👍 Looks good - thanks for the fix!

PR #229, opened 2022-10-28
Add dependabot config, plus CI config to add changenotes.

freakboy3742:
Adds a dependabot configuration to check for updates every Sunday, and automatically add a change note to every dependabot update.

Configuration copied from Briefcase; the `BRUTUS_PAT_TOKEN` has been added to the repo config, and the robots user group has been given write access. 

Dependabot has not been enabled, because this CI needs to be in place before dependabot triggers fire.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,6 @@ jobs:
     name: Pre-test checks
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 4
       matrix:
         task:
         - 'flake8'
@@ -61,7 +60,6 @@ jobs:
     name: macOS compatibility test
     needs: smoke
     strategy:
-      max-parallel: 4
       matrix:
         platform: ['macOS-12']
     runs-on: ${{ matrix.platform }}
@@ -88,7 +86,6 @@ jobs:
     needs: smoke
     runs-on: macOS-latest
     strategy:
-      max-parallel: 4
       matrix:
         python-version: ["3.5", "3.6", "3.8", "3.9", "3.10", "3.11"]
         # As soon as 3.12.0a1 is released:
--- a/.github/workflows/dependabot-changenote.yml
+++ b/.github/workflows/dependabot-changenote.yml
@@ -1,0 +1,45 @@
+name: Dependabot Change Note
+on:
+  push:
+    branches:
+      - 'dependabot/**'
+
+permissions:
+  pull-requests: write
+
+jobs:
+  changenote:
+    name: Dependabot Change Note
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    if: github.actor == 'dependabot[bot]'
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v3.1.0
+        with:
+          fetch-depth: 1
+          token: ${{ secrets.BRUTUS_PAT_TOKEN }}
+
+      - name: Configure git
+        run: |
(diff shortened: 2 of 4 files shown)


mhsmith (approved):

.github/workflows/ci.yml
```diff
@@ -10,7 +10,6 @@ jobs:
     name: Pre-test checks
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 4
```
dgelessus, line 13: What's the motivation for removing `max-parallel` from all the matrix jobs?

.github/workflows/dependabot-changenote.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Commit Change Note
+        run: |
+          # Fetch PR number for commit from Github API
+          API_URL="${{ github.api_url }}/repos/${{ github.repository }}/commits/${{ github.event.head_commit.id }}/pulls"
+          PR_NUM=$(curl -s -H "Accept: application/vnd.github+json" "${API_URL}" | jq -r '.[].number')
+
+          # Create change note from first line of dependabot commit
+          NEWS=$(printf "${{ github.event.head_commit.message }}" | head -n1 | sed -e 's/Bump/Updated/')
+          printf "${NEWS}.\n" > "./changes/${PR_NUM}.misc.rst"
```
dgelessus, line 37: This will evaluate `printf` format specifications and escape sequences in the commit message (twice!), which probably isn't intended. The safe version would be:

```suggestion
          NEWS=$(printf "%s" "${{ github.event.head_commit.message }}" | head -n1 | sed -e 's/Bump/Updated/')
          printf "%s.\n" "${NEWS}" > "./changes/${PR_NUM}.misc.rst"
```

It would be nice to get rid of the double `printf` too, but I can't think of an easy way to do that without breaking the line splitting. I'm sure there's a good `sed` or `awk` one-liner for this, but I don't know how what it would be.

.github/workflows/ci.yml
mhsmith: We couldn't remember why it was added in the first place, and it was slowing down the run. We might as well take as many parallel jobs as GitHub is willing to give us.

dgelessus: Ah, it *removes* the limit - that makes sense then. For some reason I assumed it would default to no parallelization at all.

.github/workflows/dependabot-changenote.yml
freakboy3742: True - but it's also calling `head -n1`, and the first line of the message won't (can't?) ever have anything that will be risky from a formatting perspective. You can see commits like [this one](https://github.com/beeware/briefcase/pull/959/commits/13f779c60820a2d1fb27c1c9ae40b6d9aed8d967) for what gets generated.

dgelessus: The `head -n1` is after the first `printf` call, so even formatting characters *anywhere* in the message could cause problems. You're right that it's unlikely to be an issue since the commit message is auto-generated and has a fixed format - but since the code is already using `printf` anyway, is there any reason to *not* do it the safe way?

freakboy3742: The main reason is inertia :-) . This config is the same as the one that is used on Toga and Briefcase, where it's been working well; it's a piece of code that is exceedingly difficult to test in it's entirely; and if something breaks, it's a PITA to fix it. 

However, none of those are especially *good* reasons; I'll put updating this on my list as a TODO.

PR #230, opened 2022-10-28
Enable black formatting, and add pre-commit hooks

freakboy3742:
Adds a pre-commit hook that enforces black, isort, flake8 and a couple of others.

*Disables* the docs target, in favour of using ReadTheDocs PR CI integration. This ensures that docs will actually be buildable on the one location that matters.

## PR Checklist:

- [x] All new features have been tested
- [x] All new features have been documented
- [x] I have read the **CONTRIBUTING.md** file
- [x] I will abide by the code of conduct

PR #231, opened 2022-11-09
Add Python 3.12 support

freakboy3742:
With the release of 3.12.0a1, we can start testing for 3.12 support.

## PR Checklist:


rubicon/objc/ctypes_patch.py
```diff
@@ -70,7 +69,7 @@ class PyVarObject(ctypes.Structure):
 # we can declare it as an opaque byte array, with the length taken from PyDict_Type.
 class PyDictObject(ctypes.Structure):
     _fields_ = [
-        ("PyDictObject_opaque", (ctypes.c_ubyte*PyDict_Type.tp_basicsize)),
```
freakboy3742, line 73: These changes are all Black related; see #230

dgelessus (approved):
LGTM. Though I think we should wait with this until the Black PR is merged and then rebase, to avoid pointless changes/conflicts in the formatting.

mhsmith (changes requested):
See comments on https://github.com/beeware/briefcase/pull/965.

.github/workflows/ci.yml
```diff
@@ -86,13 +86,17 @@ jobs:
...
     name: Python compatibility test
     needs: smoke
-    runs-on: macOS-latest
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.5", "3.6", "3.8", "3.9", "3.10", "3.11"]
-        # As soon as 3.12.0a1 is released:
-        # python-version: ["3.5", "3.6", "3.8", "3.9", "3.10", "3.11", "3.12.0-alpha - 3.12.0"]
+        python-version: ["3.5", "3.6", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
```
mhsmith, line 92: 3.7 is missing here.

Also, there's some inconsistency in the minimum supported version. ci.yml and tox.ini say 3.5, while ctypes_patch.py says 3.4. It may have been tested with 3.4 in the past, but it isn't anymore, so it would be misleading for the message to say that.

I suggest we set all of these to the minimum version supported by CPython itself, which is currently 3.7.

mhsmith (changes requested):

.github/workflows/ci.yml
dgelessus: 3.7 is intentionally omitted here, because it's already tested by the other jobs (`smoke` and `macOS-versions`). Might be good to put this in a comment though.

Yeah, it would make sense to update ctypes_patch to the same minimum Python version listed elsewhere. If someone really wants to use the code on 3.4 for some reason, they can ignore the warning or patch it out locally.

#230 already drops Python 3.5 and adjusts the CI config accordingly, so we could wait until that is merged and then rebase/update this PR. (I don't think we would gain much from also dropping 3.6, and AFAICT it's trivial to keep it in the test matrix for now.)

freakboy3742: I've set 3.6 as the minimum everywhere; maintaining 3.6 compatibility isn't really costing any extra effort at this point, so we might as well keep it until it becomes onerous to support it (e.g., we dropped 3.5 because it prevented us adopting f-strings). I'll wait until #230 is done before pushing for any further reviews.

dgelessus (approved):
:+1:

mhsmith (approved):

PR #236, opened 2022-11-30
Corrected the identification of x86_64 hardware on iOS simulators.

### LINKED ISSUE #235: x86_64 machine architecture not identified correctly on iOS simulator
When running on the iOS simulator on x86_64 hardware, the Rubicon doesn't correctly identify that it is running on x86_64 hardware; as a result, it fails to correctly marshal some return types.

For example, invoking a method that returns an NSRect fails with a segfault.

**To Reproduce**
1. Run a Toga Hello World app on a x86_64 simulator
2. Get a segfault when trying to access `UIScreen.mainScreen.bounds`

**Screenshots**

**Environment:**
 - Operating System: iOS, running on x86_64
 - Python version: Support package 3.11-b1+, 3.10-b5+, 3.9-8+, 3.8-b10+

**Additional context**

Reported via beeware/toga#1696.

freakboy3742:
The Python support packages for 3.11-b1, 3.10-b5, 3.9-8, and 3.8-b10 included changes to `platform.machine()` to provide a way to identify when Python code was running on iOS simulators.

However, Rubicon was using platform.machine() to identify when the simulator was using x86_64 hardware.

As a result, the `__x86_64__` symbol was never set, which altered the way some messages were marshalled. In particular, calls that returned `NSRect` structures would *not* use `obc_msgSend_stret`, causing a segfault when the return value was received.

This PR alters the check to use `os.uname()` instead of `platform`. `os.uname` is a direct operating system call with much more predictable results; platform is interpreted and transformed in a lot of situations.

This also allows us to use the same check for both ARM and x86.

There are existing tests for this behavior; however, the problem only manifests on iOS hardware.

Fixes #235.
Fixes beeware/toga#1696.

## PR Checklist:

### DIFF
--- a/rubicon/objc/types.py
+++ b/rubicon/objc/types.py
@@ -1,5 +1,5 @@
 import collections.abc
-import platform
+import os
 import struct
 from ctypes import (
     POINTER,
@@ -90,16 +90,15 @@ __all__ = [
 
 __LP64__ = 8 * struct.calcsize("P") == 64
 
-# platform.machine() indicates the machine's physical architecture,
-# which means that on a 64-bit Intel machine it is always "x86_64",
-# even if Python is built as 32-bit.
-_any_x86 = platform.machine() in ("i386", "x86_64")
+# os.uname() provides system identification details; the `machine` attribute
+# indicates the machine's physical architecture. On a 64-bit Intel machine it is
+# always "x86_64", even if Python is built as 32-bit.
+_machine = os.uname().machine
+_any_x86 = _machine in ("i386", "x86_64")
 __i386__ = _any_x86 and not __LP64__
 __x86_64__ = _any_x86 and __LP64__
 
-# On iOS, platform.machine() is a device identifier like "iPhone9,4",
-# but the platform.version() string contains the architecture.
-_any_arm = "ARM" in platform.version()
+_any_arm = _machine.startswith("arm")
 __arm64__ = _any_arm and __LP64__
 __arm__ = _any_arm and not __LP64__
 
--- a/rubicon/objc/types.py
+++ b/rubicon/objc/types.py
@@ -1,5 +1,5 @@
 import collections.abc
-import os
+import platform
 import struct
 from ctypes import (
     POINTER,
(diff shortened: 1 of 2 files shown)


rubicon/objc/types.py
```diff
@@ -90,16 +90,15 @@
...
+# always "x86_64", even if Python is built as 32-bit.
+_machine = os.uname().machine
+_any_x86 = _machine in ("i386", "x86_64")
 __i386__ = _any_x86 and not __LP64__
 __x86_64__ = _any_x86 and __LP64__
 
-# On iOS, platform.machine() is a device identifier like "iPhone9,4",
-# but the platform.version() string contains the architecture.
-_any_arm = "ARM" in platform.version()
+_any_arm = _machine.startswith("arm")
```
dgelessus, line 101: Will this work on actual iOS/iPadOS devices?

From what I can tell, the deleted comment is still accurate as of iOS 15 - the `machine` string returned by the `uname` system call is still a device identifier and not the CPU architecture. The comment you added in beeware/Python-Apple-support@2f42105838ab8f6f7e703ddb929d97758a36145e seems to confirm this.

This means the check will incorrectly fail on real hardware, so the code won't detect *any* architecture. When that happens, it seems that Rubicon by chance does all the right things for ARM64, but we obviously shouldn't rely on that.

freakboy3742: Dang... you're right. I took the "it works in testing" to be an indication that `os.uname.machine` was doing the right thing. Updated to use `platform.processor()`.

rubicon/objc/types.py
```diff
@@ -90,16 +90,18 @@
...
+# platform values added in https://github.com/beeware/Python-Apple-support/commit/2f42105838ab8f6f7e703ddb929d97758a36145e
+_processor = platform.processor()
+_any_x86 = _processor in ("i386", "x86_64")
 __i386__ = _any_x86 and not __LP64__
 __x86_64__ = _any_x86 and __LP64__
 
-# On iOS, platform.machine() is a device identifier like "iPhone9,4",
-# but the platform.version() string contains the architecture.
-_any_arm = "ARM" in platform.version()
+_any_arm = _processor.startswith("arm")
```
dgelessus, line 104:
```suggestion
if _processor:
    _any_arm = _processor.startswith("arm")
else:
    # Fallback when running on iOS without the support package,
    # where platform.processor() is an empty string
    # and the "model" field of uname/platform doesn't indicate the processor architecture.
    # In that case, look for the architecture in the kernel version string.
    _any_arm = "ARM" in platform.version()
```

dgelessus (approved):
LGTM.

I've added a suggestion below for keeping the old iOS ARM detection code as a fallback in case `platform.processor()` doesn't work. This would help when running an old version of the support package, or when not using the BeeWare support package at all - if we want to support that. In this case it doesn't add much complexity.

PR #238, opened 2022-12-06
Ensure that `run_forever_cooperatively` starts a full event loop.

### LINKED ISSUE #228: iOS event loop blocks when thread waits on socket.
When on iOS, if a thread of the main application starts a blocking server, the main event loop will partially block. It's not a complete block; if a GUI event occurs, it will be processed. However, socket activity on the main thread won't be processed as socket events occur. 

**To Reproduce**
1. Run the Toga positron-static demo

There won't be an error, but the web content won't be displayed. If you add logging to the server, you'll find that setting the URL of the main web view doesn't make a successful request to the server 

If you point an external web browser (or curl/httpx/requests) at the iPhone's web server, you'll get ReadError timeouts.

If you add a button to the GUI, mashing the button will "unstick" the main loop; after pressing a dozen or so times, the web request will complete and the web view will display the content from the server.

**Expected behavior**

Starting a socket on a thread shouldn't impact the main event loop.

**Environment:**
 - Operating System: iOS (15+?)
 - Python version: 3.8+
 - Software versions:
   - Briefcase: 0.3.11
   - Toga: 0.3.0dev39

**Additional context**

As a workaround: adding a background "heartbeat" task unsticks the main event loop. This is effectively an automated button mash. For example, using Toga:

```
async def heartbeat(*args, **kwargs):
    while True:
        await asyncio.sleep(0.1)
self.add_background_task(heartbeat)
```

freakboy3742:
Queues a recursive call to `libcf.CFRunLoopRun()` as part of the setup for `run_forever_cooperatively`.

The existing implementation of `run_forever_cooperatively` set up the lifecycle (which was a no-op on iOS), expecting the UIApplicationMain to then start the event loop. It does; however, the event loop that is started in that way appears to have a restricted set of modes; as a result, threads created from the thread that start the event loop get starved because they are only scheduled to run when the main event loop has work to do. This is why "mashing a button" will unstick a child thread; as will queuing a callback that does nothing, but is invoked periodically.

However, `libcf.CFRunLoopRun()` can be invoked recursively; if, when we start the asyncio event loop in "coop" mode, we enqueue a recursive `libcf.CFRunLoopRun()` call, that call will be invoked as soon as the UIApplicationMain starts; but the "inner" event loop is not mode-restricted, and as a result, subthreads aren't starved.

This exposed an edge case with socket callbacks: During shutdown, the reader and write object is set to None, but it is possible for stale socket callback events to be triggered (manifesting as an inability to unpack a `None` object into a (callback, args) tuple. A fix for this is also included.

Fixes #228.

## PR Checklist:

### DIFF
--- a/changes/228.bugfix.rst
+++ b/changes/228.bugfix.rst
@@ -1,0 +1,1 @@
+Background threads will no longer lock up on iOS when an asyncio event loop is in use.
--- a/src/rubicon/objc/eventloop.py
+++ b/src/rubicon/objc/eventloop.py
@@ -223,9 +223,9 @@ class CFSocketHandle(events.Handle):
             )
             return
 
-        if callbackType == kCFSocketReadCallBack:
+        if callbackType == kCFSocketReadCallBack and self._reader:
             callback, args = self._reader
-        elif callbackType == kCFSocketWriteCallBack:
+        elif callbackType == kCFSocketWriteCallBack and self._writer:
             callback, args = self._writer
         else:
             callback = None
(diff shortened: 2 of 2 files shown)


dgelessus (approved):
I'm not familiar with the details here, but all looks sensible to me.

PR #246, opened 2023-01-12
Address thread safety issues.

### LINKED ISSUE #249: Attribute intermittently returns the wrong object
This was discovered while testing https://github.com/beeware/toga/pull/1707: I'll leave it to @freakboy3742 to fill in the details.

---
# The Details

This is something that turns up when running the newly added Toga GUI test suite. I believe it could manifest in other scenarios, but it's proving hard to replicate. Even in the Toga GUI test suite, the problem is intermittent - maybe 1 in 20 runs.

This is the sequence of issues that occurs:
1. An ObjC object is created, and a Python proxy is created. (In the test case where I'm seeing it, the object is an NSLayoutConstraint; presumably one of the objects created when setting up a layout.
2. The ObjC object falls out of scope, and is disposed. However, the Python object has *not* yet been garbage collected of remains on the ObjCInstance._cached_objects web dictionary. 
3. The Objective C runtime re-allocates the same memory address to a new object (In the example case, it's reliably being re-used as the return value for `NSView.subviews` (an NSArray).

When the ObjCInstance is constructed for the NSArray object, Rubicon checks the memory address in the instance weakref dictionary, it determines that an instance already exists, and returns that instance. However, that instance has been previously loaded as an NSLayoutConstraint, so the methods cache has been loaded for a different class; and attempting to iterate over the NSArray raises an error that it isn't an iterable object.

It isn't yet clear if the issue is the gap between deletion of the object on ObjC and deletion/garbage collection of the Python object; or an issue with the retain/release strategy. However, a workaround has been to intercept the return value of `ObjCMethod.__call__`, and if it's an ObjC object, use an `object_getClass` to compare the class name associated with the returned pointer with the cached object; if there's a discrepancy, the cache is purged.

### LINKED ISSUE #250: Segmentation fault when exiting multi-threaded program
### Describe the bug

When running multiple threads, Rubicon will occasionally raise a segfault on exit.

This has been observed in the newly added Toga GUI testbed; ~1 in 10 runs will exit with a segfault

### Steps to reproduce

Run the Toga GUI test case on macOS or iOS (currently in the `macos-probe` branch). It may take multiple runs for the problem to manifest. 

### Expected behavior

No crash on exit.

### Screenshots

_No response_

### Environment

- Operating System: macOS 12.5.1, iOS 16.2
- Python version: 3.10
- Software versions:
  - Briefcase: 0.3.12-dev @ 40fccb137
  - Toga: 0.3.0dev39+ (`macos-probe` branch)

However, I have no reason to believe the problem is specific to these versions.

### Logs

The error seen occurs at shutdown of the test suite.

```
Fatal Python error: Segmentation fault

Thread 0x00000001707ab000 (most recent call first):
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/eventloop.py", line 312 in cancel
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/eventloop.py", line 284 in disable_read
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/eventloop.py", line 364 in _remove_reader
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/selector_events.py", line 94 in _close_self_pipe
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/selector_events.py", line 87 in close
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/asyncio/unix_events.py", line 68 in close
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/eventloop.py", line 619 in close
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pytest_asyncio/plugin.py", line 386 in pytest_fixture_post_finalizer
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/fixtures.py", line 1039 in finish
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/fixtures.py", line 684 in <lambda>
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/runner.py", line 523 in teardown_exact
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/runner.py", line 180 in pytest_runtest_teardown
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/runner.py", line 260 in <lambda>
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/runner.py", line 339 in from_call
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/runner.py", line 259 in call_runtest_hook
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/runner.py", line 220 in call_and_report
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/runner.py", line 132 in runtestprotocol
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/runner.py", line 112 in pytest_runtest_protocol
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/main.py", line 349 in pytest_runtestloop
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/main.py", line 324 in _main
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/main.py", line 270 in wrap_session
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/main.py", line 317 in pytest_cmdline_main
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_callers.py", line 39 in _multicall
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_manager.py", line 80 in _hookexec
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/pluggy/_hooks.py", line 265 in __call__
  File "/Users/rkm/beeware/toga/venv3.10/lib/python3.10/site-packages/_pytest/config/__init__.py", line 167 in main
  File "/Users/rkm/beeware/toga/testbed/tests/testbed.py", line 24 in run_tests
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 953 in run
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 1016 in _bootstrap_inner
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 973 in _bootstrap

Current thread 0x0000000100c2c580 (most recent call first):
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/runtime.py", line 836 in send_message
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/api.py", line 212 in __call__
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/api.py", line 290 in __call__
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/eventloop.py", line 737 in start
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/eventloop.py", line 441 in run
  File "/Users/rkm/beeware/rubicon/objc/src/rubicon/objc/eventloop.py", line 489 in run_forever
  File "/Users/rkm/beeware/toga/cocoa/src/toga_cocoa/app.py", line 341 in main_loop
  File "/Users/rkm/beeware/toga/core/src/toga/app.py", line 551 in main_loop
  File "/Users/rkm/beeware/toga/testbed/tests/testbed.py", line 128 in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 86 in _run_code
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 96 in _run_module_code
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 224 in run_module
  File "<string>", line 1 in <module>
```

### Additional context

Discovered while testing beeware/toga#1707.

The Toga GUI testbed uses pytest-asyncio; this appears to be the root of the problem.

pytest-asyncio generates an event loop for each thread. The testbed uses the secondary thread to drive the test suite, and the main thread is used to run the application. The event loop on the secondary thread injects co-routines into the main thread to build fixtures, and to run the tests.

On shutdown, both event loops are closed. Each event loop has a "self pipe" which is used to wake up the event loop when a co-routine is added to a dormant event loop; it is closing this "self pipe" that raises the segfault.

The issue appears to be that in a multi-threaded case, depending on the exact sequence of events either the CFRunLoop or the socket itself passes out of scope before one of the threads is able to close the self-pipe.

### LINKED ISSUE #251: Race condition when instantiating ObjCInstance objects
### Describe the bug

If you have multiple threads creating instance objects to wrap memory addresses, there is a potential race condition associated with instance and class creation.

ObjCInstances have an instance cache; when an object is created, the cache is checked, and if a cache miss occurs, an instance is created, populating the cache. 

However, there is a gap between the cache miss, and the addition of the new object into the cache. If 2 separate threads wrap the same address at the same time, two Python objects will be created, but the one created first will be overwritten in the cache.

### Steps to reproduce

I haven't been able to reproduce this in practice; this is based on a manual inspection of the code, looking for potential crashes 

### Expected behavior

ObjCInstance creation for a specific memory address should be idempotent.

### Screenshots

_No response_

### Environment

- Operating System: macOS, iOS
- Python version: All
- Software versions:
  - Briefcase: 0.3.12dev
  - Toga: 0.3.0dev39+

### Logs

I haven't been able to reproduce this in practice.

### Additional context

This is a theoretical flaw, based on a manual inspection of the code when trying to diagnose #249, #250 & #252

### LINKED ISSUE #252: Race condition when populating the ObjCClass method/attribute cache
### Describe the bug

ObjCClass maintains a cache of ObjC method/property references. This cache is populated on first use. 

If you have multiple threads using the same Objective C class, there is a potential race condition. If the first thread tries to access a method or property, it will detect that the cache is empty, and attempt to populate it. 

`ObjCClass._load_methods()` involves an iteration over all methods in the class, and over all methods in the superclasses, adding methods to the method cache as they are discovered.

If thread 1 attempts to invoke a method, and discovers the method cache is empty, it will begin to load the method cache. However, this puts the method cache into a state where it no longer appears to be unpopulated; so if a second thread attempts to access a any method on the same class, it will discover an apparently populated cache, and use that cache to access the method. If the method hasn't been populated by the first thread, the second thread will get an AttributeError, even though the attribute is actually valid.

### Steps to reproduce

Run the Toga GUI test case on macOS or iOS (currently in the `macos-probe` branch). This problem manifests infrequently - maybe every ~50 executions.

### Expected behavior

Attribute access should be reliable and idempotent.

### Screenshots

_No response_

### Environment

- Operating System: macOS, iOS
- Python version: All
- Software versions:
  - Briefcase: 0.3.12+
  - Toga: 0.3.12+

### Logs

I haven't got an example to hand - but it always manifests as a vanilla `AttributeError` on an attribute that *should* exist.

### Additional context

Discovered while testing beeware/toga#1707.

freakboy3742:
This PR addresses a number of thread safety issues with Rubicon, identified as a result of the Toga GUI testbed app.

1. CFEventLoop was always binding to the main thread's event loop. Event loops are thread specific; in some cases, this could lead to a segfault on shutdown, as the cleanup process would attempt to register and close the self-socket, but either the main event loop had been closed and disposed of, or the self-socket was no longer valid. (#250)

2. The call to get a run loop is declared as following the GET convention; without explicit retention, it is possible for the event loop to be disposed of before CFEventLoop has finished with it. This is unlikely to occur in practice, but can't hurt to protect against.

3. Calls to CFEventLoop.stop() weren't invoking logic on the parent; as a result, the `_stopping` flag wouldn't be set.

4. If a socket was disabled and the event loop was in the process of closing and an event occurred, it was possible for a run loop to be removed twice. 

5. If a memory address was used by the Objective C runtime, wrapped in a ObjCInstance object, then disposed of by Objective C, there is a window between that disposal and the removal of the object from the ObjCInstance instance cache. As a result, if the memory address is re-used, a stale wrapper would be returned. That wrapper could (and usually would) have a different type to the new object using the address, causing errors because the "new" object will report as being of a different type, and respond to different methods. (#249)

6. In a multithreaded setup, when creating a new ObjCInstance, it was possible to create 2 different ObjCInstance wrappers for the same memory address; one of those wrappers would be lost from the instance cache (as the ObjC memory address is used as a key). (#251)

7. In a multithreaded setup, it was possible for the method cache to become corrupted if two separate threads attempt to access a method or property on an object at the same time. One of the attempts would discover an empty cache, and attempt to populate that cache; the second attempt would see an apparently populated cache, and look up a method or property. If the first thread had not yet populated the cache for the requested method/attribute, it would report an AttributeError, even though the attribute *should* exist. (#252).

There are tests for 5-7; however, I can't work out how to write tests that will reliably exercise 1-4 to occur.

Fixes #249 
Fixes #250 
Fixes #251
Fixes #252 

## PR Checklist:

### DIFF
--- a/changes/249.bugfix.rst
+++ b/changes/249.bugfix.rst
@@ -1,0 +1,1 @@
+The ObjCInstance cache no longer returns a stale wrapper objects if a memory address is re-used by the Objective C runtime.
--- a/changes/250.bugfix.rst
+++ b/changes/250.bugfix.rst
@@ -1,0 +1,1 @@
+It is safe to open an asyncio event loop on a secondary thread. Previously this would work, but would intermittently fail with a segfault when then loop was closed.
--- a/changes/251.bugfix.rst
+++ b/changes/251.bugfix.rst
@@ -1,0 +1,1 @@
+A potential race condition that would lead to duplicated creation on ObjCInstance wrapper objects has been resolved.
--- a/changes/252.bugfix.rst
+++ b/changes/252.bugfix.rst
@@ -1,0 +1,1 @@
+A race condition associated with populating the ObjCClass method/property cache has been resolved.
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -780,15 +780,29 @@ class ObjCInstance:
             return super(ObjCInstance, type(self)).__getattribute__(self, "_objc_class")
         except AttributeError:
             # This assumes that objects never change their class after they are
-            # seen by Rubicon. Technically this might not always be true,
-            # because the Objective-C runtime provides a function
-            # object_setClass that can change an object's class after creation,
-            # and some code also manipulates objects' isa pointers directly
-            # (although the latter is no longer officially supported by Apple).
-            # This is only extremely rarely done in practice though, and then
-            # usually only during object creation/initialization, so it's
-            # basically safe to assume that an object's class will never change
-            # after it's been wrapped in an ObjCInstance.
+            # seen by Rubicon. There are two reasons why this may not be true:
+            #
+            # 1. Objective-C runtime provides a function object_setClass that
+            #    can change an object's class after creation, and some code
+            #    manipulates objects' isa pointers directly (although the latter
+            #    is no longer officially supported by Apple). This is not
+            #    commonly done in practice, and even then it is usually only
+            #    done during object creation/initialization, so it's basically
+            #    safe to assume that an object's class will never change after
+            #    it's been wrapped in an ObjCInstance.
+            # 2. If a memory address is freed by the Objective-C runtime, and
+            #    then re-allocated by an object of a different type, but the
+            #    Python ObjCInstance wrapper persists, Python will assume the
+            #    object is still of the old type. If a new ObjCInstance wrapper
+            #    for the same pointer is re-created, a check is performed to
+            #    ensure the type hasn't changed; this problem only affects
+            #    pre-existing Python wrappers. If this occurs, it probably
+            #    indicates an issue with the retain count on the Python side (as
+            #    the Objective-C runtime shouldn't be able to dispose of an
+            #    object if Python still has a handle to it). If this *does*
+            #    happen, it will manifest as objects appearing to be the wrong
+            #    type, and/or objects having the wrong list of attributes
+            #    available. Refs #249.
             super(ObjCInstance, type(self)).__setattr__(
                 self, "_objc_class", ObjCClass(libobjc.object_getClass(self))
             )
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -2,6 +2,7 @@ import collections.abc
 import decimal
 import enum
 import inspect
+import threading
 import weakref
 from ctypes import (
     CFUNCTYPE,
@@ -991,8 +992,9 @@ class ObjCInstance:
         cls = self.objc_class
         while cls is not None:
             # Load the class's methods if we haven't done so yet.
-            if cls.methods_ptr is None:
-                cls._load_methods()
+            with _ObjCClass_load_methods_lock:
+                if cls.methods_ptr is None:
+                    cls._load_methods()
 
             try:
                 method = cls.partial_methods[name]
@@ -1087,6 +1089,11 @@ class ObjCInstance:
             libobjc.objc_setAssociatedObject(self, key, None, 0x301)
 
 
+# A re-entrant thread lock moderating access to ObjCClass._load_methods().
+# This ensures that only one thread populates the method table for each class.
+_ObjCClass_load_methods_lock = threading.RLock()
+
+
 # The inheritance order is important here.
 # type must come after ObjCInstance, so super() refers to ObjCInstance.
 # This allows the ObjCInstance constructor to receive the class pointer
@@ -1125,7 +1132,7 @@ class ObjCClass(ObjCInstance, type):
         name = ensure_bytes(name)
         ptr = get_class(name)
         if ptr.value is None:
-            raise NameError("ObjC Class '%s' couldn't be found." % name)
+            raise NameError(f"ObjC Class {name} couldn't be found.")
 
         return ptr, name
 
@@ -1284,7 +1291,6 @@ class ObjCClass(ObjCInstance, type):
 
         new_attrs = {
             "name": objc_class_name,
-            "methods_ptr_count": c_uint(0),
             "methods_ptr": None,
             # Mapping of name -> method pointer
             "instance_method_ptrs": {},
@@ -1327,8 +1333,9 @@ class ObjCClass(ObjCInstance, type):
         objc_method = None
         while supercls is not None:
             # Load the class's methods if we haven't done so yet.
-            if supercls.methods_ptr is None:
-                supercls._load_methods()
+            with _ObjCClass_load_methods_lock:
+                if supercls.methods_ptr is None:
+                    supercls._load_methods()
 
             try:
                 objc_method = supercls.instance_methods[name]
@@ -1508,11 +1515,11 @@ class ObjCClass(ObjCInstance, type):
 
     def _load_methods(self):
         if self.methods_ptr is not None:
-            raise RuntimeError("_load_methods cannot be called more than once")
+            raise RuntimeError(f"{self}_load_methods cannot be called more than once")
 
-        self.methods_ptr = libobjc.class_copyMethodList(
-            self, byref(self.methods_ptr_count)
-        )
+        methods_ptr_count = c_uint(0)
+
+        methods_ptr = libobjc.class_copyMethodList(self, byref(methods_ptr_count))
 
         if self.superclass is not None:
             if self.superclass.methods_ptr is None:
@@ -1523,8 +1530,8 @@ class ObjCClass(ObjCInstance, type):
                 partial = self.partial_methods[first] = ObjCPartialMethod(first)
                 partial.methods.update(superpartial.methods)
 
-        for i in range(self.methods_ptr_count.value):
-            method = self.methods_ptr[i]
+        for i in range(methods_ptr_count.value):
+            method = methods_ptr[i]
             name = libobjc.method_getName(method).name.decode("utf-8")
             self.instance_method_ptrs[name] = method
 
@@ -1542,6 +1549,9 @@ class ObjCClass(ObjCInstance, type):
             # order is rest without the dummy "" part
             order = rest[:-1]
             partial.methods[frozenset(rest)] = (name, order)
+
+        # Set the list of methods for the class to the computed list.
+        self.methods_ptr = methods_ptr
 
 
 class ObjCMetaClass(ObjCClass):
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -852,13 +852,18 @@ class ObjCInstance:
                 # discrepancy, purge the cache for the memory address, and
                 # re-create the object.
                 #
-                # We only do this on ObjCInstance (not ObjCClass, ObjCMetaClass,
-                # etc) as there's a race condition on startup - retrieving
-                # `.objc_class` causes the creation of ObjCClass objects, which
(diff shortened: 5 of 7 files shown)


dgelessus (approved):
Oof, this looks like a really hairy set of problems. Thank you very much for debugging and fixing everything!

It's unfortunate that the fix makes `ObjCInstance.__new__` even more complex than it already is, but it's definitely needed. I thought about this a bit and can't think of a way to simplify this without also breaking the fix.

It *might* be possible to optimize some of the locking code in some way and perhaps avoid locking entirely in some cases, but it's probably not worth worrying about that unless the locking causes significant performance problems.

There's a somewhat relevant comment in the implementation of `ObjCInstance.objc_class` about objects changing their class after creation. That probably needs updating now - even though this fix isn't about objects actually changing their class.

freakboy3742:
> Oof, this looks like a really hairy set of problems. Thank you very much for debugging and fixing everything!
> 
> It's unfortunate that the fix makes `ObjCInstance.__new__` even more complex than it already is, but it's definitely needed. I thought about this a bit and can't think of a way to simplify this without also breaking the fix.
>
> It _might_ be possible to optimize some of the locking code in some way and perhaps avoid locking entirely in some cases, but it's probably not worth worrying about that unless the locking causes significant performance problems.

I had exactly the same thought. I tried to make the locks as granular as possible while keeping the locking behavior obvious enough that it's possible to reason that deadlocks won't happen, and without needing to create a lock on each individual attribute. A bunch of the locks around attribute population are *possibly* non-required, as they are populated on an attribute-by-attribute basis; and the cache-spoiling behavior will only lead to a single attribute accessor being created then overwritten. I guess that's a "memory usage at time of first access vs guaranteed single creation" tradeoff; time will tell whether the performance overhead is enough of a problem to warrant revisiting this.

> There's a somewhat relevant comment in the implementation of `ObjCInstance.objc_class` about objects changing their class after creation. That probably needs updating now - even though this fix isn't about objects actually changing their class.

I'll make that update before I merge.

PR #272, opened 2023-02-15
Improve error message when a selector argument has the wrong type

### LINKED ISSUE #271: ctypes.ArgumentError message is split into individual characters
```
rubicon-objc       0.4.5
toga-cocoa         0.3.0
toga-core          0.3.0
```

```pycon
Python 3.8.10 (v3.8.10:3d8993a744, May  3 2021, 09:09:08) 
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import toga
>>> toga.Slider(tick_count=3.0)
  File "<stdin>", line 1, in <module>
  File "/Users/msmith/.venv/bw-38/lib/python3.8/site-packages/toga/widgets/slider.py", line 78, in __init__
    self.tick_count = tick_count
  File "/Users/msmith/.venv/bw-38/lib/python3.8/site-packages/toga/widgets/slider.py", line 156, in tick_count
    self._impl.set_tick_count(tick_count)
  File "/Users/msmith/.venv/bw-38/lib/python3.8/site-packages/toga_cocoa/widgets/slider.py", line 51, in set_tick_count
    self.native.numberOfTickMarks = tick_count
  File "/Users/msmith/.venv/bw-38/lib/python3.8/site-packages/rubicon/objc/api.py", line 1133, in __setattr__
    ObjCBoundMethod(method, self)(value)
  File "/Users/msmith/.venv/bw-38/lib/python3.8/site-packages/rubicon/objc/api.py", line 290, in __call__
    return self.method(self.receiver, *args, **kwargs)
  File "/Users/msmith/.venv/bw-38/lib/python3.8/site-packages/rubicon/objc/api.py", line 212, in __call__
    result = send_message(
  File "/Users/msmith/.venv/bw-38/lib/python3.8/site-packages/rubicon/objc/runtime.py", line 836, in send_message
    result = send(receiver, selector, *args, *varargs)
ctypes.ArgumentError: ('a', 'r', 'g', 'u', 'm', 'e', 'n', 't', ' ', '3', ':', ' ', '<', 'c', 'l', 'a', 's', 's', ' ', "'", 'T', 'y', 'p', 'e', 'E', 'r', 'r', 'o', 'r', "'", '>', ':', ' ', 'w', 'r', 'o', 'n', 'g', ' ', 't', 'y', 'p', 'e', ' ', '(', 's', 'e', 'l', 'e', 'c', 't', 'o', 'r', ' ', '=', ' ', 'r', 'u', 'b', 'i', 'c', 'o', 'n', '.', 'o', 'b', 'j', 'c', '.', 'r', 'u', 'n', 't', 'i', 'm', 'e', '.', 'S', 'E', 'L', '(', 'b', "'", 's', 'e', 't', 'N', 'u', 'm', 'b', 'e', 'r', 'O', 'f', 'T', 'i', 'c', 'k', 'M', 'a', 'r', 'k', 's', ':', "'", ')', ',', ' ', 'a', 'r', 'g', 't', 'y', 'p', 'e', 's', ' ', '=', ' ', '[', '<', 'c', 'l', 'a', 's', 's', ' ', "'", 'c', 't', 'y', 'p', 'e', 's', '.', 'c', '_', 'l', 'o', 'n', 'g', "'", '>', ']', ')')
```

This is because of the following code:

https://github.com/beeware/rubicon-objc/blob/a80a7dc4c81a311ada62f9034d68c7c62977f299/src/rubicon/objc/runtime.py#L835-L840

An exception's `args` attribute should be a list, not a single string.

freakboy3742:
When a selector is provided an argument of the wrong argument type, the error message is (a) a raw ctypes error, (b) includes lots of Rubicon and Python internal detail (e.g., class reprs), and (c) is rendered incorrectly, because the exception argument is provided as a string, not a list.

This alters the exception so it is a Python TypeError (matching what would be returned by a "normal" Python method; and cleans up the message so it only contains the useful pieces of data.

Fixes #271.

## PR Checklist:

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -381,6 +381,18 @@ class RubiconTest(unittest.TestCase):
         with self.assertRaises(TypeError):
             obj.mutateIntFieldWithValue_(123, "extra argument")
 
+    def test_method_incorrect_argument_type(self):
+        """Attempting to call a method with the wrong type of argument throws an exception."""
+
+        Example = ObjCClass("Example")
+        obj = Example.alloc().init()
+
+        with self.assertRaisesRegex(
+            TypeError,
+            r"argument 3: TypeError: wrong type \(mutateIntFieldWithValue: argtypes: c_int\)",
+        ):
+            obj.mutateIntFieldWithValue_(1.234)
+
     def test_method_incorrect_argument_count_send(self):
         """Attempting to call a method with send_message with an incorrect
         number of arguments throws an exception."""
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -1,10 +1,12 @@
 import functools
 import gc
 import math
+import sys
 import threading
 import unittest
 import weakref
 from ctypes import (
+    ArgumentError,
     Structure,
     byref,
     c_char,
@@ -388,10 +390,12 @@ class RubiconTest(unittest.TestCase):
         obj = Example.alloc().init()
 
         with self.assertRaisesRegex(
-            TypeError,
-            # This pattern match needs to be flexible because the exact
-            # text returned in the ctype error changes between Python versions.
-            r"argument 3: .*?: wrong type \(mutateIntFieldWithValue: argtypes: c_int\)",
+            ArgumentError,
+            (
+                r"mutateIntFieldWithValue: argument 3: TypeError: wrong type; argtypes: c_int"
+                if sys.version_info >= (3, 10)
+                else r"mutateIntFieldWithValue: argument 3: <class 'TypeError'>: wrong type; argtypes: c_int"
+            ),
         ):
             obj.mutateIntFieldWithValue_(1.234)
 
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -389,7 +389,9 @@ class RubiconTest(unittest.TestCase):
 
         with self.assertRaisesRegex(
             TypeError,
-            r"argument 3: TypeError: wrong type \(mutateIntFieldWithValue: argtypes: c_int\)",
+            # This pattern match needs to be flexible because the exact
+            # text returned in the ctype error changes between Python versions.
+            r"argument 3: .*?: wrong type \(mutateIntFieldWithValue: argtypes: c_int\)",
         ):
             obj.mutateIntFieldWithValue_(1.234)
 
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -391,10 +391,13 @@ class RubiconTest(unittest.TestCase):
 
         with self.assertRaisesRegex(
             ArgumentError,
-            (
-                r"mutateIntFieldWithValue: argument 3: TypeError: wrong type; argtypes: c_int"
+            r"mutateIntFieldWithValue: argument 3: "
+            + (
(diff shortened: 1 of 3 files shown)


tests/test_core.py
```diff
@@ -381,6 +381,20 @@ def test_method_incorrect_argument_count(self):
...
+        """Attempting to call a method with the wrong type of argument throws an exception."""
+
+        Example = ObjCClass("Example")
+        obj = Example.alloc().init()
+
+        with self.assertRaisesRegex(
+            TypeError,
+            # This pattern match needs to be flexible because the exact
+            # text returned in the ctype error changes between Python versions.
+            r"argument 3: .*?: wrong type \(mutateIntFieldWithValue: argtypes: c_int\)",
```
mhsmith, line 394: Can we list the possible texts using the `(A|B)` syntax, rather than allowing just anything here?

mhsmith: Why does it say "argument 3"? I could understand argument 1 being `self`, or the Objective-C equivalent, but what's argument 2?

mhsmith (commented):
A couple of comments below.

Also, I don't understand why CI hasn't run on the most recent commit, and I don't see any way to run it manually.

dgelessus (commented):
Thanks for the fix and the readability improvements!

For better parity with normal Python `TypeError`s, we could put the Objective-C method name at the start of the error, e. g.:

```patch
-argument 3: TypeError: wrong type (mutateIntFieldWithValue: argtypes: c_int)
+mutateIntFieldWithValue, argument 3: TypeError: wrong type (argtypes: c_int)
```

That said, I feel like it's not really useful to change the exception type from `ctypes.ArgumentError` to `TypeError` - especially considering that the default `ctypes` exception already says `TypeError` in the message string, so the full text seen by the user will be `TypeError: argument 3: TypeError: ...`.

tests/test_core.py
dgelessus: Argument 1 is indeed `self`. Argument 2 is `_cmd`, the selector (method name) of the method being called. Argument 3 and on are the actual method arguments.

freakboy3742: I've taken an alternate approach - include both messages as exact text, with a "if sys.version" check; that way, when we (eventually) deprecate Python 3.9, pyupgrade will automatically do the cleanup and remove the dead branch.

dgelessus (approved):

mhsmith:
Looks like there's a third phrasing showing up in the CI for Python 3.12.

Also, for future reference, do either of you know why CI didn't run for the previous commit? You can still see it by going to the "Checks" tab at the top, then selecting the commit "Make the test resilient to Python version differences".

freakboy3742:
> Looks like there's a third phrasing showing up in the CI for Python 3.12.

Fixed.

> Also, for future reference, do either of you know why CI didn't run for the previous commit? You can still see it by going to the "Checks" tab at the top, then selecting the commit "Make the test resilient to Python version differences".

That's an odd one. Can't say I can spot a reason; I'd chalk it up to a screwup on the part of Github Actions.

mhsmith (approved):

PR #285, opened 2023-03-31
flake8 tweaks

rmartin16:
## Changes
- Remove unnecessary `noqa` markers
- Simplify `flake8` configuration
  - Remove `max-complexity`.....20 is a little high and I'm guessing someone was just trying to disable the check
  - Use [default](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-ignore) list of checks to `exclude`
    - E266 - Too many leading '#' for block comment
      - This doesn't seem to be relevant anymore
    - E501 - Line too long
      -  Added `noqa` markers where appropriate....or broke the line
    - W503 - Line break occurred before a binary operator
      - In the default list of excluded checks

## Notes
This one is a little different from the others since I'm making code changes. If you don't want to introduce these changes for the sake of a linter, I definitely understand and can just put `E501` back in the ignore list....although, I think it's better if that check actually runs.

## PR Checklist:

### DIFF
--- a/changes/285.misc.rst
+++ b/changes/285.misc.rst
@@ -1,0 +1,1 @@
+Removed unnecessary ``noqa`` markers for flake8 and simplified the configuration.
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,5 +64,8 @@ exclude=\
     venv*/*,\
     local/*,\
     .tox/*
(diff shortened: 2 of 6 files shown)


freakboy3742 (approved):
👍

PR #294, opened 2023-04-13
Further cleanup of spelling docs

freakboy3742:
The SSL error on macOS was caused because the PyPI name filter was turned on. This made every single one of the 440k names of PyPI packages a valid word in spelling. So... that needs to be turned off.

That removes the need for the SSL instructions in the contribution guide. 

It also reveals a bunch of additional spelling problems. The only ones that *aren't* legitimate are the ones tied to a package name; for those, I've adopted the policy of:
* If it's being used in a "code" context, use the name as a literal
* If it's being used in a "text" context, ensure preferred project naming and capitalisation is used.

I've also updated all the links that were raising redirects. Ideally, redirects could be raised as an error, but that doesn't appear to be an option with Sphinx.

## PR Checklist:

### DIFF
--- a/docs/background/releases.rst
+++ b/docs/background/releases.rst
@@ -9,10 +9,7 @@ Release History
 Bugfixes
 --------
 
-* Classes that undergo a class name change between ``alloc()`` and ``init()``
-  (e.g., ``NSWindow`` becomes ``NSKVONotifying_Window``) no longer trigger
-  instance cache eviction logic. (`#258
-  <https://github.com/beeware/rubicon-objc/issues/258>`__)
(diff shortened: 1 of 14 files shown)


mhsmith:
This PR unwraps release notes which were wrapped when generating them from the `changes` files (e.g. a80a7dc4c81a311ada62f9034d68c7c62977f299). But why were they wrapped at all, when pyproject.toml doesn't contain a towncrier `wrap` setting, and [the default value is false](https://towncrier.readthedocs.io/en/latest/configuration.html#all-options)? Were you wrapping them manually?

mhsmith (approved):

PR #309, opened 2023-05-12
Add support for PEP563 deferred annotations

### LINKED ISSUE #308: Incompatibility with `from __future__ import annotations`
### Describe the bug

Registering new objc classes by subclassing NSObject with method definitions and type hints fails with an error message after importing `from __future__ import annotations`. It appears that something about how type hints are used to determine Objc-C argument and return types is breaking.

<details>
  <summary>Traceback</summary>

```
KeyError                                  Traceback (most recent call last)
File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/rubicon/objc/types.py:389, in encoding_for_ctype(ctype)
    388 try:
--> 389     return _encoding_for_ctype_map[ctype]
    390 except KeyError:

KeyError: 'int'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
Cell In[2], line 4
      1 from rubicon.objc import NSObject, objc_method
----> 4 class Handler(NSObject):
      5     @objc_method
      6     def initWithValue_(self, v: int):
      7         self.value = v

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/rubicon/objc/api.py:1363, in ObjCClass.__new__(cls, name_or_ptr, bases, attrs, protocols)
   1361             return ObjCMetaClass(ptr)
   1362 else:
-> 1363     ptr, name, attrs = cls._new_from_class_statement(
   1364         name_or_ptr, bases, attrs, protocols=protocols
   1365     )
   1367 objc_class_name = name.decode("utf-8")
   1369 new_attrs = {
   1370     "name": objc_class_name,
   1371     "methods_ptr": None,
   (...)
   1387     "cache_lock": threading.RLock(),
   1388 }

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/rubicon/objc/api.py:1279, in ObjCClass._new_from_class_statement(cls, name, bases, attrs, protocols)
   1277             pass
   1278         else:
-> 1279             class_register(ptr, attr_name)
   1281 # Register any user-defined dealloc method. We treat dealloc differently to
   1282 # inject our own cleanup code for properties, ivars, etc.
   1284 user_dealloc = attrs.get("dealloc", None)

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/rubicon/objc/api.py:348, in objc_method.class_register(self, class_ptr, attr_name)
    346 def class_register(self, class_ptr, attr_name):
    347     name = attr_name.replace("_", ":")
--> 348     add_method(class_ptr, name, self, self.encoding)

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/rubicon/objc/runtime.py:1032, in add_method(cls, selector, method, encoding, replace)
   1030     ctypes_patch.make_callback_returnable(signature[0])
   1031 selector = SEL(selector)
-> 1032 types = b"".join(encoding_for_ctype(ctype) for ctype in signature)
   1034 cfunctype = CFUNCTYPE(*signature)
   1035 imp = cfunctype(method)

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/rubicon/objc/runtime.py:1032, in <genexpr>(.0)
   1030     ctypes_patch.make_callback_returnable(signature[0])
   1031 selector = SEL(selector)
-> 1032 types = b"".join(encoding_for_ctype(ctype) for ctype in signature)
   1034 cfunctype = CFUNCTYPE(*signature)
   1035 imp = cfunctype(method)

File /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/rubicon/objc/types.py:392, in encoding_for_ctype(ctype)
    390 except KeyError:
    391     try:
--> 392         return b"^" + encoding_for_ctype(ctype._type_)
    393     except KeyError:
    394         raise ValueError(f"No type encoding known for ctype {ctype}")

AttributeError: 'str' object has no attribute '_type_'
```
</details>

### Steps to reproduce

Run the subclassing example in the tutorial:

```python
from __future__ import annotations

from rubicon.objc import NSObject, objc_method


class Handler(NSObject):
    @objc_method
    def initWithValue_(self, v: int):
        self.value = v
        return self

    @objc_method
    def pokeWithValue_andName_(self, v: int, name) -> float:
        print("My name is", name)
        return v / 2.0

```

### Expected behavior

Not sure if how easy this will be, but ideally we rubicon-objc should properly handle type hints even with https://peps.python.org/pep-0563/ - Postponed Evaluation of Annotations.

### Screenshots

_No response_

### Environment

- Operating System: macOS 13.3.1 
- Python version: 3.11.3
- Software versions:
  - Toga: 0.4.6

### Logs

```

```

### Additional context

_No response_

freakboy3742:
PEP563 (`from __future__ import annotations`) introduced the idea of deferred annotations - type annotations on function declarations are now parsed as strings, with conversion into actual types being deferred until they are actually required.

This PR adds the changes to perform this annotation evaluation.

PEP563 is opt-in; the approach taken in this PR is compatible with both pre- and post PEP563 code. If code *doesn't* include the `__future__` import, the `typing.get_type_hints()` call returns the type annotations as-parsed.

Fixes #308.

## PR Checklist:

### DIFF
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -3,6 +3,7 @@ import decimal
 import enum
 import inspect
 import threading
+import typing
 import weakref
 from ctypes import (
     CFUNCTYPE,
@@ -93,11 +94,19 @@ _keep_alive_objects = {}
 
 def encoding_from_annotation(f, offset=1):
     argspec = inspect.getfullargspec(inspect.unwrap(f))
+    # typing evaluates a return type of None as <class NoneType>;
+    # convert that back to a None. E721 is the "don't compare types,
+    # use isinstance()" flake8 error. Which is good advice... except
+    # that we're not checking an instance - we *are* comparing types.
+    hints = {
+        var: None if hint == type(None) else hint  # noqa: E721
+        for var, hint in typing.get_type_hints(f).items()
+    }
 
-    encoding = [argspec.annotations.get("return", ObjCInstance), ObjCInstance, SEL]
+    encoding = [hints.get("return", ObjCInstance), ObjCInstance, SEL]
 
     for varname in argspec.args[offset:]:
-        encoding.append(argspec.annotations.get(varname, ObjCInstance))
+        encoding.append(hints.get(varname, ObjCInstance))
 
     return encoding
 
@@ -2219,7 +2228,8 @@ class Block:
 
         .. code-block:: python
 
-            @Block def the_block(arg: NSInteger) -> NSUInteger:
+            @Block
+            def the_block(arg: NSInteger) -> NSUInteger:
                 return abs(arg)
 
         For callables without type annotations, the parameter and return types
@@ -2246,6 +2256,7 @@ class Block:
             # so try to extract them from the function's type annotations.
 
             try:
+                hints = typing.get_type_hints(func)
                 signature = inspect.signature(func)
             except (TypeError, ValueError):
                 raise ValueError(
@@ -2253,22 +2264,23 @@ class Block:
                     "please pass return and argument types directly into Block"
                 )
 
-            if signature.return_annotation == inspect.Signature.empty:
+            try:
+                restype = hints["return"]
+            except KeyError:
                 raise ValueError(
                     "Function has no return type annotation - "
                     "please add one, or pass return and argument types directly into Block"
                 )
 
-            restype = signature.return_annotation
             argtypes = []
-            for name, param in signature.parameters.items():
-                if param.annotation == inspect.Parameter.empty:
+            for name in signature.parameters:
+                try:
+                    argtypes.append(hints[name])
+                except KeyError:
                     raise ValueError(
                         f"Function has no argument type annotation for parameter {name!r} - "
                         f"please add one, or pass return and argument types directly into Block"
                     )
-
-                argtypes.append(param.annotation)
 
         signature = tuple(ctype_for_type(tp) for tp in argtypes)
 
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -94,15 +94,7 @@ _keep_alive_objects = {}
 
 def encoding_from_annotation(f, offset=1):
     argspec = inspect.getfullargspec(inspect.unwrap(f))
-    # typing evaluates a return type of None as <class NoneType>;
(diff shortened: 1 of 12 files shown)


src/rubicon/objc/api.py
```diff
@@ -93,11 +94,19 @@
 
 def encoding_from_annotation(f, offset=1):
     argspec = inspect.getfullargspec(inspect.unwrap(f))
+    # typing evaluates a return type of None as <class NoneType>;
+    # convert that back to a None. E721 is the "don't compare types,
+    # use isinstance()" flake8 error. Which is good advice... except
+    # that we're not checking an instance - we *are* comparing types.
```
samschott, line 100: I'm not sure if I entirely understand this comment. Why not deal with `<class NoneType>` as the `type` for None, similar to how `int` is the type of `1`?

We might need to handle this properly when registering the ObjC method later, for example by registering an encoding for `type(None)`, etc:

```
register_preferred_encoding(b"v", type(None))
```

Or is the issue that CFUNCTYPE does deal well with `<class NoneType>` but only `None`?

freakboy3742: Ill admit I was being a bit lazy here, and just reproducing what the old code was doing. 

You're probably right that we could probably work through the rest of the logic to use NoneType; the only complication I can see is that we can currently rely on the falsiness of None, but we won't be able to if we use NoneType. However, I haven't done enough of a dive into the code to work out if that's a big issue or a minor inconvenience.

samschott: Lazy works for me.

samschott (approved):

samschott:
Thanks for the quick fix!

freakboy3742:
Turns out the fix was even easier than I thought - we just need to add an explicit ctypes mapping for NoneType. If there's no explicit mapping, the type is used as its own ctype mapping; this works for user-supplied types, and also with None (which is the ctypes mapping of void). However, when the encoding started returning NoneType, it was falling through, and Nonetype isn't a valid ctype, so it was raising an error.

samschott (approved):
Yes, that's definitely more elegant. Thanks!

PR #317, opened 2023-05-31
Fix scroll issue #316

### LINKED ISSUE #316: Scroll blocks eventloop for ios
### Describe the bug

We noticed that the event loop blocks when you scroll. This is similar to [issue 228](https://github.com/beeware/rubicon-objc/issues/228)

### Steps to reproduce

1. Create a 2-page linked positron-static demo (you can use my fork example [here](https://github.com/alibeyram/toga/tree/main/examples/positron-static) )
2. Scroll down before clicking to the second page and you should see that it does not work.

if you use Safari Develop > Simulator to inspect the WebView you would see a timeout error.

### Expected behavior

Scrolling should not impact the clicking behavior

### Screenshots

_No response_

### Environment

- Operating System: iOS (16.4)
- Python version: 3.10
- Software versions:
  - Briefcase: 0.3.14
  - Toga: 0.3.1
  - rubicon-objc: 0.4.6

### Logs

_No response_

### Additional context

The workaround in [issue 228](https://github.com/beeware/rubicon-objc/issues/228) also solves this problem.

```python
async def heartbeat(*args, **kwargs):
    while True:
        await asyncio.sleep(0.1)
self.add_background_task(heartbeat)
```

landrybr:
`CFRunLoopRun` causes the run loop to run in its default mode, but it seems like it's terminated when switching to another mode for example `UITrackingRunLoopMode` when the user scrolls.  I think the `NSRunLoop` class can properly handle switches between modes.

Switch `CFRunLoop` with `NSRunLoop` in `iOSLifecycle`

Fixes #316 
## PR Checklist:

### DIFF
--- a/changes/317.bugfix.rst
+++ b/changes/317.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed iOS scroll issue `#316 <https://github.com/beeware/rubicon-objc/issues/316>`_
+iOS now uses a full NSRunLoop, rather than a CFRunLoop.
--- a/changes/317.bugfix.rst
+++ b/changes/317.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed iOS scroll issue `#316 <https://github.com/beeware/rubicon-objc/issues/316>`_
--- a/src/rubicon/objc/eventloop.py
+++ b/src/rubicon/objc/eventloop.py
@@ -79,6 +79,7 @@ kCFSocketWriteCallBack = 8
 kCFSocketAutomaticallyReenableReadCallBack = 1
 kCFSocketAutomaticallyReenableWriteCallBack = 8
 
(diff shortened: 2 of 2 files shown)


freakboy3742 (approved):
Nice work! Thanks for the PR. I'm not afraid to admit that your bug report had me quite concerned that I was going to lose a week of my life spelunking for a fix - I'm very glad you were able to find a fix for me (and quite a simple one, at that).

FWIW, the manifestation of this bug was slightly different for me - *scrolling* worked fine, but clicking on the link stalled. Regardless, moving to a full NSRunLoop appears to have corrected the problem.

I also did a quick check to see if the fix from #238 was still required, or if moving to NSRunLoop would have corrected that as well. Turns out that change *is* still needed.

PR #334, opened 2023-07-09
Goodbye 3.7

freakboy3742:
Python 3.7 is [officially EOL](https://devguide.python.org/versions/), so we can remove it from the list of tested and officially supported versions.

## PR Checklist:

### DIFF
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -1515,37 +1515,36 @@ class ObjCClass(ObjCInstance, type):
     def declare_property(self, name):
         """Declare the instance method ``name`` to be a property getter.
 
-        This causes the attribute named ``name`` on instances of this class to
-        be treated as a property rather than a method --- accessing it returns
-        the property's value, without requiring an explicit method call. See
-        :class:`ObjCInstance.__getattr__` for a full description of how
-        attribute access behaves for properties.
+        This causes the attribute named ``name`` on instances of this class to be
+        treated as a property rather than a method --- accessing it returns the
+        property's value, without requiring an explicit method call. See
+        :class:`ObjCInstance.__getattr__` for a full description of how attribute access
+        behaves for properties.
 
-        Most properties do not need to be declared explicitly using this method,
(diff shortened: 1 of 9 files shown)


src/rubicon/objc/api.py
```diff
@@ -1515,37 +1515,36 @@ def _cache_property_mutator(self, name):
...
+        properties apart from normal zero-argument methods.
 
         .. note::
 
-            In the standard Apple SDKs, some properties are introduced as
-            regular methods in one system version, and then declared as
-            properties in a later system version. For example, the
-            ``description`` method/property of :class:`NSObject` was declared as
-            a regular method `up to OS X 10.9
-            <https://github.com/phracker/MacOSX-SDKs/blob/9fc3ed0ad0345950ac25c28695b0427846eea966/MacOSX10.9.sdk/usr/include/objc/NSObject.h#L40>`__,
```
mhsmith, line 1538: > Removed an explicit github line reference, since recent GUI changes seem to have disabled anchor access for robots.

The links in the current version on RTD still work fine in my browser, so if they provide useful information to non-robots, why remove them?

rmartin16: Forgive the interjection....but the same thing happened to me in another repo...the sphinx link linting [fails](https://readthedocs.org/projects/rubicon-objc/builds/21244102/) for GitHub line number anchors as of recently.

I did discover, though, this can be avoided for other types of failing GitHub anchors by prepending `user-content-` to the anchor in the link; for instance, like this [one](https://github.com/beeware/rubicon-objc/tree/main#community) since the "real" anchor is `user-content-community`. But this strategy does not work for line number anchors...

freakboy3742: The reason for removing was entirely because it wasn't possible to link check; I wasn't aware you could selectively exclude parts of the URL from the link check. I've reverted the change, and added a specific exception for line number anchors. I've confirmed that the URL part is still validated; it will just ignore any `LXX` anchor.

mhsmith (approved):

Issue #222, opened 2022-04-07
Improve repr() calls for NSPoint/NSRect types


freakboy3742:
The Python repr() representation of NSPoint, NSRect, and the CoreFoundation types is 100% accurate, but 0% useful. 

```
>>> p = NSPoint(100,200)
>>> repr(p)
'<rubicon.objc.types.NSPoint object at 0x104da00c0>
```

This makes debugging more difficult than it needs to be.

**Describe the solution you'd like**

It would be desirable if they returned a more useful representation:
```
>>> p = NSPoint(100,200)
>>> repr(p)
'<rubicon.objc.types.NSPoint 100,200>'
```

**Describe alternatives you've considered**

None

**Additional context**

Most Rubicon types will defer to `description`; however, NSPoint et al are structs, not ObjC types, so they don't inherit "useful" debug output.

This came up in the context of beeware/toga#1395, and [a proposed monkeypatch](https://github.com/beeware/toga/blob/c9e988bf3994eebf1a8efb5aee3d2598f8a18e06/src/cocoa/toga_cocoa/libs/__init__.py). While the idea is sound, it should be handled at the Rubicon level, rather than in Toga.

PR #340, opened 2023-07-22
str and repr for NSRect, NSRange, NSEdgeInsets, NSSize, CFRange and UIEdgeInsets

conniepocky:
Improved repr and string calls for NSPoint, NSRect, NSRange, NSEdgeInsets, NSSize, CFRange and UIEdgeInsets.

#222 

## PR Checklist:

### DIFF
--- a/src/rubicon/objc/types.py
+++ b/src/rubicon/objc/types.py
@@ -810,7 +810,7 @@ class NSRect(Structure):
 
     def __str__(self):
         return (
-            f"({self.size.width}x{self.size.height} @ {self.origin.x}, {self.origin.y})"
+            f"{self.size.width}x{self.size.height} @ ({self.origin.x},{self.origin.y})"
         )
 
 
--- a/src/rubicon/objc/types.py
+++ b/src/rubicon/objc/types.py
@@ -771,6 +771,12 @@ else:
             ("y", CGFloat),
         ]
 
+        def __repr__(self):
+            return f"<CGPoint({self.x}, {self.y})>"
+
+        def __str__(self):
+            return f"({self.x}, {self.y})"
+
 
 @with_preferred_encoding(_NSSizeEncoding)
 class NSSize(Structure):
@@ -780,10 +786,10 @@ class NSSize(Structure):
     ]
 
     def __repr__(self):
-        return f"<NSSize {self.width}x{self.height}>"
+        return f"<NSSize({self.width}, {self.height})>"
 
     def __str__(self):
-        return f"{self.width}x{self.height}"
+        return f"{self.width} x {self.height}"
 
 
 if _CGSizeEncoding == _NSSizeEncoding:
@@ -797,6 +803,12 @@ else:
             ("height", CGFloat),
         ]
 
+        def __repr__(self):
+            return f"<CGSize({self.width}, {self.height})>"
+
+        def __str__(self):
+            return f"{self.width} x {self.height}"
+
 
 @with_preferred_encoding(_NSRectEncoding)
 class NSRect(Structure):
@@ -806,12 +818,14 @@ class NSRect(Structure):
     ]
 
     def __repr__(self):
-        return f"<NSRect {self.size.width}x{self.size.height} @ ({self.origin.x},{self.origin.y})>"
+        return (
+            f"<NSRect("
+            f"NSPoint({self.origin.x}, {self.origin.y}), "
+            f"NSSize({self.size.width}, {self.size.height}))>"
+        )
 
     def __str__(self):
-        return (
-            f"{self.size.width}x{self.size.height} @ ({self.origin.x},{self.origin.y})"
-        )
+        return f"{self.size} @ {self.origin}"
 
 
 if _CGRectEncoding == _NSRectEncoding:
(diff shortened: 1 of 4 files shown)


src/rubicon/objc/types.py
```diff
@@ -779,6 +779,12 @@ class NSSize(Structure):
         ("height", CGFloat),
     ]
 
+    def __repr__(self):
+        return f"<NSSize({self.width}x{self.height})>"
```
freakboy3742, line 783: The brackets aren't really needed here - they're common when describing a point, but not when describing a size.
```suggestion
        return f"<NSSize {self.width}x{self.height}>"
```

src/rubicon/objc/types.py
```diff
@@ -779,6 +779,12 @@ class NSSize(Structure):
...
+
+    def __str__(self):
+        return f"({self.width}x{self.height})"
```
freakboy3742, line 786:
```suggestion
        return f"{self.width}x{self.height}"
```

src/rubicon/objc/types.py
```diff
@@ -799,6 +805,14 @@ class NSRect(Structure):
         ("size", NSSize),
     ]
 
+    def __repr__(self):
+        return f"<NSRect({self.size.width}x{self.size.height} @ {self.origin.x}, {self.origin.y})>"
```
freakboy3742, line 809: Here we've got a size (which doesn't need brackets), and a point (which does, but only on the "pointy" part):
```suggestion
        return f"<NSRect {self.size.width}x{self.size.height} @ ({self.origin.x},{self.origin.y})>"
```

src/rubicon/objc/types.py
```diff
@@ -799,6 +805,14 @@ class NSRect(Structure):
...
+
+    def __str__(self):
+        return (
+            f"({self.size.width}x{self.size.height} @ {self.origin.x}, {self.origin.y})"
```
freakboy3742, line 813:
```suggestion
            f"{self.size.width}x{self.size.height} @ ({self.origin.x},{self.origin.y})"
```

src/rubicon/objc/types.py
```diff
@@ -846,6 +860,12 @@ class UIEdgeInsets(Structure):
         ("right", CGFloat),
     ]
 
+    def __repr__(self):
+        return f"<UIEdgeInsets({self.top}, {self.bottom}, {self.left}, {self.right})>"
```
freakboy3742, line 864: It might be helpful to identify which values are which in the repo. There's no "natural" ordering like "width x height" or "(x, y)", so being explicit that `top=1 bottom=2...` is probably the best we can do.

src/rubicon/objc/types.py
```diff
@@ -889,6 +915,12 @@ class CFRange(Structure):
         ("length", CFIndex),
     ]
 
+    def __repr__(self):
+        return f"<CFRange({self.location}, {self.length})>"
```
freakboy3742, line 919: As with insets, there's no "natural" order or label here, so labelling them as "location=123, length=456" would be helpful

tests/test_core.py
```diff
@@ -999,6 +1001,48 @@ def test_objcinstance_nspoint_repr(self):
...
+        """Test NSRect repr and str returns correct value."""
+
+        my_rect = NSRect(NSPoint(10, 20), NSSize(5, 15))
+        self.assertEqual(repr(my_rect), "<NSRect(5.0x15.0 @ 10.0, 20.0)>")
+        self.assertEqual(str(my_rect), "(5.0x15.0 @ 10.0, 20.0)")
+
+    def test_objcinstance_nssize_repr(self):
+        """Test NSSize repr and str returns correct value."""
+
+        my_size = NSSize(5, 5)
```
freakboy3742, line 1014: When writing a test that has multiple inputs, it's generally a good idea to use different values - that way, you can catch accidental transposition errors. For example, if the implementation for this method printed the width for the height, and the height for the width, this test wouldn't identify the problem.

freakboy3742: The same applies to the following tests for range, insets etc.

freakboy3742 (changes requested):
This looks really good - I've flagged a couple of cosmetic changes in the specific printed output, and a couple of potential testing improvements.

dgelessus:
For the `__repr__`s, it may be a good idea to make them look like valid constructor calls, so that they can be copy-pasted back into Python code or the REPL. For example, if `NSSize` uses the repr `f"NSSize({self.width}, {self.height})"`, then the value `NSSize(12, 34)` would be shown exactly like that by `repr`.

(This won't be as nice with `NSRect` though, because it has nested structures. A repr like `NSRect(NSPoint(12, 34), NSSize(56, 78))` would be accurate, but less readable.)

freakboy3742:
> For the `__repr__`s, it may be a good idea to make them look like valid constructor calls, so that they can be copy-pasted back into Python code or the REPL. For example, if `NSSize` uses the repr `f"NSSize({self.width}, {self.height})"`, then the value `NSSize(12, 34)` would be shown exactly like that by `repr`.
> 
> (This won't be as nice with `NSRect` though, because it has nested structures. A repr like `NSRect(NSPoint(12, 34), NSSize(56, 78))` would be accurate, but less readable.)

I hadn't thought of that, but it's a good idea. We can keep `__str__` "human readable coordinates" (so -`(10,20)` or `10x20 @ (5,15)`, but having `__repr__` be "reflective" could be useful from time to time.

freakboy3742 (approved):
I've made a couple more minor tweaks, and cleaned up the test cases; but this is awesome. Thanks so much for this contribution - it might not seem like much, but it will make a huge improvement to my (and other Rubicon users) development experience!

conniepocky:
> 

Your welcome :) and thank you for helping me with doing my first contribition 😁

PR #357, opened 2023-08-29
Correct block handling on M1 hardware

### LINKED ISSUE #225: Blocks not fully working on M1 hardware
**Describe the bug**
The Rubicon ObjC Test suite has failures when executed on M1 hardware. The failures are all in the tests of Block functionality; some, but not all tests are currently failing.

The test failures all appear to be related to the return values from block methods.

**To Reproduce**
1. Run the test suite on M1 hardware2. 
2. Test failures occur:
```
==============================================================================
FAIL: tests/test_blocks.py::BlockTests::test_block_bound_method
------------------------------------------------------------------------------
  File "/Users/rkm/beeware/rubicon/objc/tests/test_blocks.py", line 158, in test_block_bound_method
    self.assertEqual(returned_two_args_block(12, 34), 46)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: None != 46

==============================================================================
FAIL: tests/test_blocks.py::BlockTests::test_block_delegate_auto
------------------------------------------------------------------------------
  File "/Users/rkm/beeware/rubicon/objc/tests/test_blocks.py", line 53, in test_block_delegate_auto
    self.assertEqual(result, 9)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 0 != 9

==============================================================================
FAIL: tests/test_blocks.py::BlockTests::test_block_delegate_auto_struct
------------------------------------------------------------------------------
  File "/Users/rkm/beeware/rubicon/objc/tests/test_blocks.py", line 70, in test_block_delegate_auto_struct
    self.assertEqual(result, 85)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 0 != 85

==============================================================================
FAIL: tests/test_blocks.py::BlockTests::test_block_delegate_method_manual_ctypes
------------------------------------------------------------------------------
  File "/Users/rkm/beeware/rubicon/objc/tests/test_blocks.py", line 31, in test_block_delegate_method_manual_ctypes
    self.assertEqual(result, 5)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 0 != 5

==============================================================================
FAIL: tests/test_blocks.py::BlockTests::test_block_delegate_method_manual_pytypes
------------------------------------------------------------------------------
  File "/Users/rkm/beeware/rubicon/objc/tests/test_blocks.py", line 42, in test_block_delegate_method_manual_pytypes
    self.assertEqual(result, 5)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 0 != 5

==============================================================================
FAIL: tests/test_blocks.py::BlockTests::test_block_property_ctypes
------------------------------------------------------------------------------
  File "/Users/rkm/beeware/rubicon/objc/tests/test_blocks.py", line 14, in test_block_property_ctypes
    self.assertEqual(result, 3)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: None != 3

==============================================================================
FAIL: tests/test_blocks.py::BlockTests::test_block_property_pytypes
------------------------------------------------------------------------------
  File "/Users/rkm/beeware/rubicon/objc/tests/test_blocks.py", line 20, in test_block_property_pytypes
    self.assertEqual(result, 3)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: None != 3

==============================================================================
FAIL: tests/test_blocks.py::BlockTests::test_block_round_trip
------------------------------------------------------------------------------
  File "/Users/rkm/beeware/rubicon/objc/tests/test_blocks.py", line 118, in test_block_round_trip
    self.assertEqual(returned_block(8, 9), 17)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: None != 17

------------------------------------------------------------------------------
Ran 266 tests in 2.15s

FAILED (failures=8)
```

**Expected behavior**
Test suite should pass on M1 hardware

**Screenshots**

**Environment:**
 - M1 Mac hardware
 - Operating System: macOS 12.3.1
 - Python version: 3.8, 3.9, 3.10, 3.11

**Additional context**

It is likely that these failures are also occurring on iOS devices. CI testing this is going to be difficult...

freakboy3742:
Fixes #225.

It appears that the fact that Blocks returned from Objective C APIs works at all is almost accidental.

When a block is returned from an ObjC API, an ObjCBlock instance is constructed, using the `ObjCBlockStruct` data provided for the block instance. This structure includes a pointer to an `invoke()` function, which is the function implementing the block.

The current codebase annotates the `restype` and `argtypes` onto the `invoke` method when the ObjCBlock is constructed. However, if you inspect the value of `invoke.restype` and `invoke.argtypes` immediately after they're assigned, you'll see that the assignment has had no effect - they return `c_void_p` and `(c_void_p,)` respectively (the default prototype for the invoke method, provided as the `_cfunc_type_block_invoke` declaration).

It appears that on x86_64 hardware, this doesn't matter. The `invoke` method can be invoked, ctypes doesn't complain, and the expected value is returned. However, on M1, ctypes raises TypeError exceptions; but because we're elbow-deep in block calls, that exception doesn't stop the application - it just results in a `c_void_p` "None" result from invoking the method. 

If instead we store the restype and argtypes, and apply them just before invoking the method (to a local copy of the invoke method), the invocation succeeds.

This reveals 2 additional bugs in the previous handling:
1. The invoke() method has an implied first argument that is the block itself. If an ObjCBlock is constructed with explicit `argtypes` (rather than inferring them from the block descriptor), this first argument isn't included in the argtypes of the invoke method.
2. If a block accepts a structure as an argument, the block descriptor may involve the use of an anonymous structure. When invoked with ctypes, the field pattern will match, but the structure class won't, causing ctypes to raise a TypeError. It is necessary to either cast the user-provided argument into the anonymous structure, or change the ctypes annotation to match the provided class. Since we needed to late-annotate the argtypes anyway, I've opted to use the second approach (including some light type checks).

This PR also exposes the `objc_id` and `objc_block` objects in the main `rubicon.objc` namespace; I've had frequent need to reference these types when using Rubicon as part of Toga, and it seemed odd that they weren't exposed in the core namespace.

This is an odd one for CI, because the old code also passed CI. The failure mode was only seen on M1 hardware, and until [Github adds M1 runners](https://github.com/github/roadmap/issues/528), we won't be able to do any automated validation of this. 

Huge thanks to @jeamland for his assistance nailing this one to the floor.

## PR Checklist:

### DIFF
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -2169,11 +2169,11 @@ class ObjCBlock:
             # first required argument - the block itself.
             block_arg = [objc_id]
 
-        # Setting the restype and argtypes on the invoke function that is in the
-        # ObjCBlockStruct won't stick, because it's not a distinct Python object that
-        # ctypes can use to attach a type hint. Store the ctypes annotations, and apply
-        # them just before invocation. We're going to have to do some light type
-        # conversion in some cases anyway, so this works out well.
+        # If you set restype and argtypes on the invoke function that is in the
+        # ObjCBlockStruct, subsequent gets won't reflect those changes, because it's not
+        # a distinct Python object that ctypes can use to attach a type hint. Store the
+        # ctypes annotations, and apply them just before invocation. We're going to have
+        # to do some light type conversion in some cases anyway, so this works out well.
         self.invoke_restype = ctype_for_type(restype)
         self.invoke_argtypes = block_arg + [
             ctype_for_type(arg_type) for arg_type in argtypes
@@ -2201,13 +2201,21 @@ class ObjCBlock:
         # of type names, so even if the field types of a structure provided as an
         # argument match, ctypes will raise a TypeError.
         #
-        # To avoid this, look for the `__anonymous__` property on the structure
-        # definition. If it exists, check that the provided argument matches the "shape"
-        # of the anonymous structure. If it matches, modify the invoke signature to
-        # match the type of the argument that was actually provided. The first argument
-        # to invoke is the block being invoked, so we can ignore that type hint.
+        # To avoid this, if an argument is a structure, and the argtype for that
+        # argument is a structure, look for the `__anonymous__` property on the argtype
+        # structure definition - this property is added automatically to structures when
+        # a structure type is constructed from a type descriptor that doesn't provide a
+        # name. If it exists, the structure has been anonymously declared; so we check
+        # that the provided argument matches the "shape" of the anonymous structure. If
+        # it matches, modify the invoke signature to match the type of the argument that
+        # was actually provided. The first argument to invoke is the block being
+        # invoked, so we can ignore that type hint.
         for i, argtype in enumerate(self.invoke_argtypes[1:]):
-            if getattr(argtype, "__anonymous__", False):
+            if (
+                isinstance(args[i], Structure)
+                and issubclass(argtype, Structure)
+                and getattr(argtype, "__anonymous__", False)
+            ):
                 anon_fields = [f[1] for f in argtype._fields_]
                 arg_fields = [f[1] for f in args[i]._fields_]
                 if anon_fields != arg_fields:
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -2161,10 +2161,23 @@ class ObjCBlock:
             if not self.has_signature:
                 raise ValueError("Cannot use AUTO types for blocks without signatures")
             restype, *argtypes = ctypes_for_method_encoding(self.signature)
-        self.struct.contents.invoke.restype = ctype_for_type(restype)
-        self.struct.contents.invoke.argtypes = (objc_id,) + tuple(
+            # If the argtypes have been derived from the signature, they will include
+            # the block as the first argument.
+            block_arg = []
+        else:
+            # If the argtypes are explicitly provided, they *won't* include the
+            # first required argument - the block itself.
+            block_arg = [objc_id]
+
+        # Setting the restype and argtypes on the invoke function that is in the
+        # ObjCBlockStruct won't stick, because it's not a distinct Python object that
+        # ctypes can use to attach a type hint. Store the ctypes annotations, and apply
+        # them just before invocation. We're going to have to do some light type
+        # conversion in some cases anyway, so this works out well.
+        self.invoke_restype = ctype_for_type(restype)
+        self.invoke_argtypes = block_arg + [
             ctype_for_type(arg_type) for arg_type in argtypes
-        )
+        ]
 
     def __repr__(self):
         representation = f"<ObjCBlock@{hex(addressof(self.pointer))}"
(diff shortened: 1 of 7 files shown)


mhsmith:
I only have a very sketchy understanding of Objective C, so I won't be reviewing this one.

src/rubicon/objc/types.py
```diff
@@ -254,10 +254,12 @@ def _create_structish_type_for_encoding(encoding, *, base):
...
 
-    # Create the subclass. The _fields_ are not set yet, this is done later.
-    # The structish is already registered here, so that pointers to this structish type in itself are typed correctly.
+    # Create the subclass. The _fields_ are not set yet, this is done later. The
+    # structish is already registered here, so that pointers to this structish type in
+    # itself are typed correctly. Also annotate the structure with an `__anonymous__`
+    # marker so that we can easily identify anonymous structures.
     py_name = "_Anonymous" if name is None else name.decode("utf-8")
-    structish_type = type(py_name, (base,), {})
+    structish_type = type(py_name, (base,), {"__anonymous__": True})
```
samschott, line 262:
```suggestion
    structish_type = type(py_name, (base,), {"__anonymous__": name is None})
```

src/rubicon/objc/api.py
```diff
@@ -2182,7 +2195,35 @@ def __call__(self, *args):
...
+        # of type names, so even if the field types of a structure provided as an
+        # argument match, ctypes will raise a TypeError.
+        #
+        # To avoid this, look for the `__anonymous__` property on the structure
+        # definition. If it exists, check that the provided argument matches the "shape"
+        # of the anonymous structure. If it matches, modify the invoke signature to
+        # match the type of the argument that was actually provided. The first argument
+        # to invoke is the block being invoked, so we can ignore that type hint.
+        for i, argtype in enumerate(self.invoke_argtypes[1:]):
+            if hasattr(argtype, "__anonymous__"):
```
samschott, line 2210:
```suggestion
            if getattr(argtype, "__anonymous__", False):
```

samschott (changes requested):
Wow, this was a lot of detective work! I've left minor comments inline, otherwise LGTM.

src/rubicon/objc/api.py
samschott: If I understand correctly, there is no need manually check the field types if this is a named structure, right?

samschott: Also, maybe add a check that we are indeed dealing with a structure? Using the presence of an `__anonymous__` field for this is very implicit requires knowledge of when this field is set.

tests/test_blocks.py
```diff
@@ -72,7 +57,6 @@ def exampleMethod_(self, block: objc_block):
         result = instance.blockExample()
         self.assertEqual(result, 9)
 
-    @m1_block_failure
     def test_block_delegate_auto_struct(self):
```
samschott, line 60: How about adding a test case for with a named (non-anonymous) structure as argument as well?

src/rubicon/objc/api.py
freakboy3742: You're correct that no type check is required if it is a named structure. If the block explicitly declares its argument as an NSPoint, ctypes will accept an NSPoint. It's only when the block anonymously declares that it accepts "a structure that contains 2 double fields" that we need to check the shape and modify the argtype so that ctypes won't generate a TypeError.

Explicitly checking for a Structure type is definitely worthwhile - ``__anonymous__`` *should* only exist on structures, but it doesn't hurt to be more explicit.

tests/test_blocks.py
freakboy3742: Done. I've added a test of a mismatched structure as well - however, as Python exceptions aren't propagated back into the ObjC runtime (and then back out into Python), the test is a little weird; but I've annotated the weirdness.

samschott (approved):
Thanks!

PR #374, opened 2023-10-02
Add M1 and Py3.12 runners.

freakboy3742:
Py3.12 is final; Py3.13 is available in dev mode; and M1 runners [are now available for FLOSS projects](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/) using the new macOS-14 runner target.

actions/setup-python doesn't provide M1 builds for Py3.8 or 3.9, so they're excluded from the matrix.

## PR Checklist:


rmartin16:
friendly fyi...since i can't remember if we've explicitly discussed this before...but this [file](https://github.com/actions/python-versions/blob/main/versions-manifest.json) is the source of truth for Python versions available to the `setup-python` action.

freakboy3742:
> friendly fyi...since i can't remember if we've explicitly discussed this before...but this [file](https://github.com/actions/python-versions/blob/main/versions-manifest.json) is the source of truth for Python versions available to the `setup-python` action.

Oh sure - it's mentioned in the failure message for the 3.12 builds. I know they haven't released 3.12 yet - but it will be there in a day or two, so I wanted to be ready. Given that they announced M1 builder availability on the same day as 3.12 was released, I figured I'd roll them together, since neither is urgently required.

mhsmith (approved):

PR #387, opened 2023-11-22
Move to a pure PEP621 project configuration.

freakboy3742:
Inspired by beeware/toga#2229, a full PEP621 configuration update.

There's no functional difference; the one technical benefit (aside from keeping up with modern standards) is that because we're using `setuptools_scm`, we don't need to maintain MANIFEST.in any more. The source discovery process is also much simpler - it requiring *no* explicit setuptools configuration, as we're entirely relying on one of the "default" project layout patterns.

I've audited the generated tarballs, and they have identical content, except for:
* the missing setup.cfg and MANIFEST.in files, 
* a missing `not-zip-safe` file in the egg-info. This file was needed in the days of eggs and `easy-install`, but is [now considered deprecated](https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html).
* minor changes to the METADATA file in the wheel:
  * A different format for authors and license
  * Slightly different URL definitions
  * No `platform: any` declaration. This has been subsumed by wheel tagging.

Also removes the Travis CI configuration, since that's long dead.

## PR Checklist:


dgelessus (commented):
This seems like a good move to me :+1:

pyproject.toml
```diff
@@ -1,7 +1,70 @@
 [build-system]
-requires = ["setuptools>=60", "setuptools_scm[toml]>=7.0"]
+requires = [
+    "setuptools==69.0.0",
+    "setuptools_scm==8.0.4",
```
dgelessus, line 4: Why the pins here instead of minimum versions like before? Surely we support more than just these exact versions.

(The [setuptools pyproject.toml docs](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) say that at least setuptools 61.0.0 is required.

freakboy3742: Reproducibility. For example, the `dynamic` argument in pyproject.toml wasn't required on setuptools 68.0, but it *is* required on 69.0. If you use an open ended pin, a project configuration that works on the day of release can break in the future.

With a hard pin, we know the configuration works with this specific set of versions, and that won't change. Those versions might not be compatible with future versions of pip, Python etc - but then, we can't make any guarantees about what might happen in future regardless.

freakboy3742: To clarify further: build-requires specify the environment that will be created specifically for the purpose of wheel creation. It's guaranteed to be an isolated environment (unless you specifically disable isolation, in which case you're on your own), so we don't need to maintain maximum compatibility with the end-user's environment. You're specifying the exact environment that you want to be present to build the wheel; an open ended pin undermines that, because we can't guarantee that the build environment will match a configuration that we know works.

mhsmith (approved):

PR #401, opened 2024-01-20
Updated __init__.py __all__

qqfunc:
Changed the contents of __all__ in __init__.py to strings.


mhsmith:
Thanks very much. To merge this PR we'll need a couple more things:

* A change note, as described [here](https://toga.readthedocs.io/en/stable/how-to/contribute-code.html#add-change-information-for-release-notes).

* A test to verify the fix. In this case, it could be as simple as running `from rubicon.objc import *` and comparing the contents of `globals()` before and after. This would require a separate test module, because `import *` statements aren't allowed inside functions.

qqfunc:
Sorry, I'm a beginner to such projects and GitHub, so I can't understand detailed operations.
What should I do...
I'm sorry that I can't help you.

mhsmith:
Let's start with the change note. You can create it by following the instructions in the link above, starting with "First submit it without a change note - this will fail ..." You can see from the "Some checks were not successful" caption below that this has already happened. After you push the change note, that should change.

qqfunc:
Thanks to your advice, I think I was able to add a change note.

Issue #181, opened 2020-06-14
Automatic renaming of duplicate class?


jsbain:
in 

`_new_from_class_statement`

if a classname on the objc side already exists, the following is thrown:
    RuntimeError('An Objective-C class named {!r} already exists'.format(name))

While this behavior makes sense for code run from a command line, where the lifetime of new objc classes is short, in an environment like pythonista, where we only get one process, and so any created objccclasses can never be deleted, it would be nice to specify a flag to automatically rename the class on the objc side. 

Maybe some sort decorator on the class, which changes the class name (checks for existing name, and renames if needed ).  or an option that can be overridden in `api` after import.  obviously, one could never lookup the custom class by name with this scheme (alhough perhaps there could also be a dict of remapped classnames)

I am not familiar enough yet with how rubicon-objc is creating the class pair, and whether something like this is even possible.  Would it be possible to disassociate the objc class name from the python class name, or does rubicon rely on those always being the same?

jsbain:
By auto renaming, I am specifically thinking of adding a numerical suffix to the base name (keep trying until a unique name is found) -- similar to the med used by pythonista when the debug flag is set in create_objc_class.

dgelessus:
This would definitely be a useful feature for environments where restarting the Python interpreter is inconvenient (like Pythonista or a REPL).

As far as I can tell, it should be possible to implement this like how Pythonista does it in `objc_util`, i. e. automatically adjust the Objective-C name of the class until it doesn't conflict with an existing class anymore. rubicon.objc itself generally doesn't use the class name - it passes it to Objective-C when creating the class, but afterwards it only uses the `Class` pointer to work with the class (calling methods, etc.). So it *shouldn't* matter if the class has a different name than what the user requested.

Of course the same would also be required for the user code - that is, the user should only reference the newly created class via its Python name and cannot look it up by its Objective-C name via `ObjCClass`. If we make the behavior opt-in via a flag when the class is created, this shouldn't be an issue though - by enabling that flag, the user accepts that they are only allowed to use the class via its pointer/Python name and not via its Objective-C name. (Maybe we should even enforce that - if the user sets the "auto-rename" flag, *always* append an integer suffix, so that it's impossible to accidentally reference the class through a hardcoded Objective-C name.)

[...]

PR #419, opened 2024-03-01
Add automatic renaming system to ObjCClass and ObjCProtocol

qqfunc:
Added automatic renaming options to `ObjCClass` and `ObjCProtocol`. (Documentation about them has also been added.) Refs #181

## PR Checklist:


freakboy3742 (changes requested):
Thanks for the PR; however, it's currently failing CI because the documentation has spelling errors in the changes you've made. There's also no tests of the new behaviour.

freakboy3742 (changes requested):
I've already reviewed this and told you there were two problems. You've fixed one of them.

qqfunc:
Sorry, I'm not familiar with GitHub, so I got some procedures wrong.

src/rubicon/objc/api.py
```diff
@@ -1335,7 +1350,11 @@ class object (in any form accepted by :class:`ObjCInstance`).
         The bases list must contain exactly one :class:`ObjCClass` to be
         extended by the new class. An optional ``protocols`` keyword argument is
         also accepted, which must be a sequence of :class:`ObjCProtocol`\\s for
-        the new class to adopt.
+        the new class to adopt. In addition, another optional keyword argument
+        `auto_rename` determines whether to rename the defined class if a class
+        with the same name has already been defined. (For example, ``MyClass``
+        is renamed to ``MyClass_2``.) If its value is ``None``,
+        :attr:`ObjCClass.auto_rename` is preferred.
```
freakboy3742, line 1357:
```suggestion
        the new class to adopt. 
        
        If the name of the class has already registered with the Objective C
        runtime, the ``auto_rename`` option can be used to ensure that the
        Objective C name for the new class will be unique. A numeric suffix will
        be appended to the Objective C name to ensure uniqueness (for example,
        ``MyClass`` will be renamed to ``MyClass_2``, ``MyClass_3`` etc until a
        unique name is found). By default, classes will *not* be renamed, unless
        :attr:`ObjCProtocol.auto_rename` is set at the class level.
```

src/rubicon/objc/api.py
```diff
@@ -1869,7 +1892,11 @@ def __new__(cls, name_or_ptr, bases=None, ns=None):
         Objective-C protocol from Python (see
         :ref:`custom-classes-and-protocols`). The bases list can contain any
         number of :class:`ObjCProtocol` objects to be extended by the new
-        protocol.
+        protocol. Also, an optional keyword argument `auto_rename` determines
+        whether to rename the defined protocol if a protocol with the same name
+        has already been defined. (For example, ``MyProtocol`` is renamed to
+        ``MyProtocol_2``.) If its value is ``None``,
+        :attr:`ObjCProtocol.auto_rename` is preferred.
```
freakboy3742, line 1899:
```suggestion
        protocol. 
        
        If the name of the class has already registered with the Objective C
        runtime, the ``auto_rename`` option can be used to ensure that the
        Objective C name for the new class will be unique. A numeric suffix will
        be appended to the Objective C name to ensure uniqueness (for example,
        ``MyClass`` will be renamed to ``MyClass_2``, ``MyClass_3`` etc until a
        unique name is found). By default, classes will *not* be renamed, unless
        :attr:`ObjCProtocol.auto_rename` is set at the class level.
```

tests/test_core.py
```diff
@@ -1118,6 +1118,40 @@ class MyClass(NSObject):
...
+
+            class TestClass2(NSObject):
+                pass
+
+        ObjCClass.auto_rename = False
+
+        for _ in range(2):
+
+            class TestClass(NSObject, auto_rename=True):
+                pass
```
freakboy3742, line 1136: This test exercises the fact that the code doesn't raise an exception, but doesn't verify anything about the operation of the code. e.g.:

* The ObjC name of the resulting class is unique, and follows the expected pattern
* The renaming scheme skips names allocated manually that would collide (e.g., if I explicitly declare `TestClass_2`, that name won't be allocated)
* The TestClass instances generated by different renames don't compare as equal
* When a class name is reused, the new method implementations are also used
* The test isn't dependent on the specific order in which the test is invoked.

tests/test_core.py
```diff
@@ -1118,6 +1118,40 @@ class MyClass(NSObject):
...
+
+            class TestProtocol(metaclass=ObjCProtocol):
+                pass
+
+        ObjCProtocol.auto_rename = False
+
+        for _ in range(2):
+
+            class TestProtocol2(metaclass=ObjCProtocol, auto_rename=True):
+                pass
```
freakboy3742, line 1153: As above, this verifies there's no error raised, but doesn't verify any behavior.

docs/tutorial/tutorial-2.rst
```diff
@@ -115,6 +115,21 @@ the same Python shell. If you try, you'll get an error:
...
+
+    >>> ObjCClass.auto_rename = True
+
+You can also do this by using the ``auto_rename`` argument.
+
+.. code-block:: pycon
+
+    >>> class Handler(NSObject, auto_rename=True):
+    ...     pass
+
```
freakboy3742, line 132:
```suggestion
To allow a class name to be re-used, you can set the class variable
:attr:`~rubicon.objc.api.ObjCClass.auto_rename` to ``True``. This option enables
automatic renaming of the Objective C class if a naming collision is detected:

.. code-block:: pycon

    >>> ObjCClass.auto_rename = True

This option can also be enabled on a per-class basis by using the
``auto_rename`` argument in the class declaration:

.. code-block:: pycon

    >>> class Handler(NSObject, auto_rename=True):
    ...     pass

If this option is used, the Objective C class name will have a numeric suffix
(e.g., `Handler_2`). The Python class name will be unchanged.
```

qqfunc:
Thanks for the reviews. I'll modify them.

changes/419.feature.rst
```diff
@@ -0,0 +1 @@
+Added automatic renaming options to ObjCClass and ObjCProtocol.
```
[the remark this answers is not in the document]
qqfunc: Is there no need for a description about `ObjCProtocol`?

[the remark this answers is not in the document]
qqfunc: And, should it be `181.feature.rst`?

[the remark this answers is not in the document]
freakboy3742: Yes to both.

tests/test_core.py
```diff
@@ -1107,6 +1107,8 @@ def test_uiedgeinsets_repr(self):
     def test_duplicate_class_registration(self):
         """If you define a class name twice in the same runtime, you get an error."""
 
+        ObjCClass.auto_rename = False
```
freakboy3742, line 1110: I suspect this is only needed because `test_class_auto_rename` is setting the property; if that class-based property has an impact on other tests, then the test that sets the property should clean up and unset it.

tests/test_core.py
```diff
@@ -1118,6 +1120,67 @@ class MyClass(NSObject):
             class MyClass(NSObject):  # noqa: F811
                 pass
 
+    def test_class_auto_rename(self):
+        """Test automatic renaming system of ObjCClass."""
+
+        # Check that the global option works.
+
+        ObjCClass.auto_rename = True
```
freakboy3742, line 1128: As noted above - if this is a persistent effect, then the entire test should be in a try/finally block:

```
try:
    ObjCClass.auto_rename = True

   ... do test here ...
finally:
    ObjCClass.auto_rename = None
```

tests/test_core.py
```diff
@@ -1118,6 +1120,67 @@ class MyClass(NSObject):
...
+        self.assertIsNot(testclass1, TestClass)
+
+        object1 = TestClass.new()
+
+        with self.assertRaises(AttributeError):
+            object1.oldMethod()
+
+        self.assertEqual(object1.testMethod(), "TEST1")
+
+        # Check that the per-class option works.
```
freakboy3742, line 1155: This test is complex enough that it would be worth splitting into two parts - one for the global setting; and one for the class-specific setting. As part of that, we probably want to make the test class name something that is unique enough that we won't accidentally re-use the same name elsewhere, causing confusion (e.g. call the class TestRenamedClass, rather than TestClass)

tests/test_core.py
```diff
@@ -1193,13 +1256,58 @@ def fiddle_(cls, value: int) -> None:
...
+
+        class TestProtocol(metaclass=ObjCProtocol):  # noqa: F811
+            @objc_method
+            def newMethod(self):
+                pass
+
+        self.assertEqual(TestProtocol.name, "TestProtocol_3")
+        self.assertIsNot(testprotocol1, TestProtocol)
+
+        # Check that the per-protocol option works.
```
freakboy3742, line 1293: As above; this should probably be split into 2 tests, with try/finally isolation, and class names that will minimise the chance of accidental overlap with other tests.

tests/test_core.py
```diff
@@ -1193,6 +1300,8 @@ def fiddle_(cls, value: int) -> None:
     def test_no_duplicate_protocols(self):
         """An Objective-C class cannot adopt a protocol more than once."""
 
+        ObjCProtocol.auto_rename = False
+
```
freakboy3742: This change shouldn't be required any more.
```suggestion
```

freakboy3742 (approved):
I've made a couple of minor tweaks to the class naming, but otherwise, this looks great - thanks for the patch!

PR #435, opened 2024-03-30
Fix the documentation (error message and venv path)

qqfunc:
- Fix the error message in the documentation (`tutorial-2.rst`) to the current one.
- Remove the `.sh` extension from the venv activation path in the documentation (`get-started.rst`).

## PR Checklist:

### DIFF
--- a/changes/435.misc.rst
+++ b/changes/435.misc.rst
@@ -1,0 +1,2 @@
+- Fix the error message in the documentation (`tutorial-2.rst`) to the current one.
+- Remove the `.sh` extension from the venv activation path in the documentation (`get-started.rst`).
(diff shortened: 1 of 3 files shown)


freakboy3742 (approved):
Thanks again for the fix.

So it's clear - the purpose of the release note isn't to reproduce the commit message. It's to summarise changes at the time of release so that end users know what has changed. The actual text doesn't really matter for "misc" level entries, as the text isn't used; but for any other entry, it should be exactly one statement, with no bullets or other start-of-line markup, describing the change *as it would appear in the final release notes*. You're describing a feature that users of the new release will observe, at a level that makes sense to end users - so, you'd say "The documentation now includes a description of how to fooble a barwhiz object", rather than describing the commit that implemented that change ("added a description of how to fooble a barwiz object")

qqfunc:
I guess I didn't understand it well enough. I will be careful.

PR #436, opened 2024-03-31
Improve `__repr__` methods of some classes beeware#432

### LINKED ISSUE #432: Duplicate elements in repr of `ObjCInstance`
### What is the problem or limitation you are having?

The repr of `ObjCInstance` is as follows:
```
<rubicon.objc.api.ObjCInstance (id): (class name) at (address): (debugDescription)>
```

For example:
```
<rubicon.objc.api.ObjCInstance 0x104d0f690: NSObject at 0x6000031a8050: <NSObject: 0x6000031a8050>>
```

The class name and address parts are duplicated because `NSObject.debugDescription` also contains them.
I think this may be a little long.

### Describe the solution you'd like

```
<rubicon.objc.api.ObjCInstance (id): (class name) at (address)>
```
```
<rubicon.objc.api.ObjCInstance 0x104d0f690: NSObject at 0x6000031a8050>
```
In this form, `NSObject.debugDescription` is omitted.

### Describe alternatives you've considered

```
<rubicon.objc.api.ObjCInstance (id): (debugDescription)>
```
```
<rubicon.objc.api.ObjCInstance 0x104d0f690: <NSObject: 0x6000031a8050>>
```
This time, `(class name) at (address)` is omitted.

### Additional context

_No response_

qqfunc:
Made some changes to the representation strings of some classes, as follows:
* Made the string use `__name__`:
  * `ObjCMethod`
* Removed the module names (such as `rubicon.objc.api.`) from the strings:
  * `ObjCPartialMethod`
  * `ObjCBoundMethod`
  * `ObjCInstance`
  * `ObjCClass`
  * `ObjCProtocol`
* Removed the included addresses:
  * `ObjCInstance`
  * `ObjCClass`
  * `ObjCProtocol`

Fixes #432.

## PR Checklist:


changes/436.misc.rst
```diff
@@ -0,0 +1 @@
+Improved `__repr__` methods of `ObjCBoundMethod`, `ObjCClass`, `ObjCInstance`, `ObjCMethod`, `ObjCPartialMethod`, and `ObjCProtocol`.
```
freakboy3742, line 1:
```suggestion
The `__repr__` output for  `ObjCBoundMethod`, `ObjCClass`, `ObjCInstance`, `ObjCMethod`, `ObjCPartialMethod`, and `ObjCProtocol` was simplified.
```

freakboy3742 (changes requested):
The implementations all look good; however, as before, this requires tests. Admittedly, the old behavior wasn't tested; but if we're making a conscious change to behavior, we should ensure that behavior will remain correct in the long term.

changes/436.misc.rst
freakboy3742, line 1: This is a significant feature change - so it should be `432.feature.rst`, referencing the issue number that describes the new feature.

tests/test_core.py
```diff
@@ -986,6 +987,32 @@ def test_partial_method_lots_of_args(self):
...
+        bound_method = obj.init
+        method = bound_method.method
+        receiver = bound_method.receiver
+        partial_method = obj.performSelector.method
+
+        # Check ObjCBoundMethod repr
+        objcboundmethod_repr = repr(obj.init)
+        self.assertIn(ObjCBoundMethod.__qualname__, objcboundmethod_repr)
+        self.assertIn(repr(method), objcboundmethod_repr)
+        self.assertIn(str(receiver), objcboundmethod_repr)
```
freakboy3742, line 1003: Tests usually benefit from being explicit, even if there's content that *can* be obtained programmatically. In this case, the name of the bound method is technically provided by `__qualname__`, but it's better to have an *explicit* test of a *literal* string, rather than relying on the implementation of `__qualname__` to do the right thing. 

Adding to that - a `assertRegex` would be a better match for these tests. Using `assertIn` allows any output ordering, as long as it includes the right parts; a regex-based test lets you check the exact output, with an explicit carveout for the characters that will vary based on memory address.

tests/test_core.py
```diff
@@ -986,6 +987,32 @@ def test_partial_method_lots_of_args(self):
         )
         self.assertEqual(buf.value.decode("utf-8"), pystring)
 
+    def test_objcmethod_objcpartialmethod_objcboundmethod_repr(self):
```
freakboy3742, line 990: No need to be *quite* so explicit in the test name:
```suggestion
    def test_objcmethod_repr(self):
```

freakboy3742 (changes requested):
One comment about the way you've structured the new tests; the same comment applies to most of the repr tests you've added.

freakboy3742 (approved):
I've made a couple of changes to the test cases to make them more explicit - while "assertIn" checks are technically correct, explicit string checks make it a lot clearer what is being asserted and why. Highlighting why this matters - by using the literal string as an assertion, a further cleanup of ObjCMethod's repr became obvious - decoding values that were byte strings back into strings, so the `b''` isn't displayed.

Otherwise - thanks for the patch!

qqfunc:
Thank you too!

Issue #461, opened 2024-04-29
Improve the error message of partial methods


qqfunc:
### What is the problem or limitation you are having?

The current error message is as follows:

```python
>>> obj.someMethod(..., invalidArg=...)
```
```
Traceback (most recent call last):
    ...
ValueError: No method was found starting with 'someMethod' and with keywords {'invalidArg'}
Known keywords are:
frozenset({''})
frozenset({'', 'validArg'})
```

This will be as follows after #453 is resolved because `frozenset` will be changed to `tuple`:

```
ValueError: No method was found starting with 'someMethod' and with keywords ('invalidArg')
Known keywords are:
('')
('', 'validArg')
```

### Describe the solution you'd like

I think the converted selector name and available selector names should be shown.

```
ValueError: Invalid selector 'someMethod:invalidArg:'. Available selectors are: 'someMethod:', 'someMethod:validArg:'
```

### Describe alternatives you've considered

It is also possible to show descriptions of Python, not selectors.

```
ValueError: Invalid method 'someMethod(..., invalidArg=...)'. Available methods are: 'someMethod(...)', 'someMethod(..., validArg=...)'
```

Actual values of arguments are represented as `...`.

freakboy3742:
Agreed this would be a good improvement - especially if it were paired with an implementation of #453 and #148. 

Of the two syntax options you've presented, I'd lean towards the "selector" description - any user is trying to invoke an ObjC method, so I don't think we gain anything by trying to avoid ObjC syntax in error messages.

qqfunc:
Thanks for agreeing. I will try it with the issues.

freakboy3742:
Fixed by #462.

Issue #453, opened 2024-04-28
Partial method argument order


qqfunc:
### What is the problem or limitation you are having?

As described in #26, when there is an Objective-C class that has `someMethod:arg1:arg2:` and `someMethod:arg2:arg1:` methods...

```python
>>> obj.someMethod(..., arg1=..., arg2=...)
>>> obj.someMethod(..., arg2=..., arg1=...)
```

These call the same method (`someMethod:arg1:arg2:`). This seems to be because partial methods only store a single order of arguments.

### Describe the solution you'd like

Allow them to be distinguished. (Then non-existent argument orders may no longer be allowed.) The argument names of partial methods are stored using dictionaries whose keys are `frozenset`. And, the values of the dictionaries are like `(selector string, [argument1 name, argument2 name, ...])`. In this case, storing multiple selectors and lists may be a solution.

### Describe alternatives you've considered

Store no argument names. Then, use [`respondsToSelector:`](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418583-respondstoselector) in `ObjCPartialMethod` to check if the method of specified name exists. (I feel there are some other ways...)

freakboy3742:
Is this a problem you're having with an actual API? If so - which one? I'm a little surprised that Apple would construct an API where the argument order is the discriminator between method forms. 

Regardless, I think you've missed another possible option for implementation. ObjCPartial currently uses `frozenset` to track the options on an ObjCPartialMethod. If that were to be replaced with an "order preserving set" - i.e., more of a "tuple, but inserting the same item is prohibited" - then `arg1:arg2:` and `arg2:arg1` should hash differently, and thus yield different lookups. 

In fact, we might even be able to combine this with a fix for #148 - using a literal `tuple` would be a key part of that fix, with the other part being post-processing the kwargs to remove `__` suffixes.

qqfunc:
I think this is only if I implement such methods myself. I understand such a situation won't be a problem in almost cases. Its priority would be low.

By the way, I thought the use of `frozenset` was for performance since the similar `list` is also contained in dictionary values. Is this wrong? Also, does the way of using `tuple` as keys mean that the `list` of the argument names is no longer necessary?

freakboy3742:
[...]

Issue #148, opened 2020-02-16
Add mechanism for invoking methods with duplicated argument names.


freakboy3742:
Objective C arguments are all named, but ordered. As a result, it is entirely legal for argument names to be duplicated. For example, `NSLayoutConstraint` has a method `+constraintWithItem:attribute:relatedBy:toItem:attribute:multiplier:constant:` - the `attribute` argument name is duplicated.

This is a problem for Rubicon, because the obvious mapping to keyword arguments wont work:
```
NSLayoutConstraint.constraintWithItem(
    item,
    attribute=...,
    relatedBy=...
    toItem=...
    attribute=...
    multiplier=...
    constant=...
)
```

These methods *can* be accessed using the "older" underscore-based API mapping:
```
NSLayoutConstraint.constraintWithItem_attribute_relatedBy_toItem_attribute_multiplier_constant_(...)
```

but this will almost always violate line length constraints.

We need a way to the more Pythonic invocation style on methods that have duplicated arguments.

freakboy3742:
Some ideas:
1) Argument suffixing. Look for a double underscore in any argument name, and strip anything after the double underscore.
```
NSLayoutConstraint.constraintWithItem(
    item,
    attribute__from=...,
    relatedBy=...
    toItem=...
    attribute__to=...
    multiplier=...
    constant=...
)
```

Downsides: you won't be able to invoke any argument that actually has a double underscore in it's name. Objective C tends towards camelCase naming, though, so this is probably a low risk. 

2) Explicit aliasing. At some point in the declaration process, provide a way to map keywords provided in Python to keywords used in the call.
```
NSLayoutConstraint.constraintWithItem.set_argument_mapping(
    fromAttribute='attribute',
    toAttribute='attribute', 
)

NSLayoutConstraint.constraintWithItem(
    item,
    fromAttribute=...,
    relatedBy=...
    toItem=...
    toAttribute=...
    multiplier=...
    constant=...
)
```

This allows for an explicit mapping of names - or even a complete re-mapping of names. For example, you could completely change the API to use snake case, rather than camelCase:
```
NSLayoutConstraint.constraintWithItem.set_argument_mapping(
    attr='attribute',
    related_by='relatedBy',
    to_item='toItem',
    to_attr='attribute',     
)

NSLayoutConstraint.constraintWithItem(
    item,
    attr=...,
    related_by=...
    to_item=...
    to_attr=...
    multiplier=...
    constant=...
)
```
[...]

PR #462, opened 2024-04-30
Partial method enhancement

qqfunc:
The partial method system was changed as follows:
1. Add the "__" keyword mechanism #148
2. Change to check the order of keyword arguments strictly #453
3. Improve the error message #461

## PR Checklist:


tests/objc/Example.m
```diff
@@ -241,6 +241,36 @@ +(NSUInteger) overloaded:(NSUInteger)arg1 extraArg:(NSUInteger)arg2
...
+
++(NSUInteger) overloaded:(NSUInteger)arg duplicateArg:(NSUInteger)arg1 duplicateArg:(NSUInteger)arg2
+{
+    return arg1 + arg2;
+}
+
++(void) dummyForException:(NSUInteger)arg
+{
+    // Dummy method for testing exception
+}
```
freakboy3742, line 267: There's no need to have these extra methods just to demonstrate we can't call them correctly. The test will work just as well by calling `Example.overloaded(0, notAnArgument=10)`. It would be arguably *better* to use `overloaded:...` , because there are more options, and some of them are ambiguous.

tests/test_core.py
```diff
@@ -986,6 +986,31 @@ def test_partial_method_lots_of_args(self):
...
         self.assertEqual(buf.value.decode("utf-8"), pystring)
 
+    def test_partial_method_arg_order(self):
+        Example = ObjCClass("Example")
+
+        self.assertEqual(Example.overloaded(0, extraArg1=0, extraArg2=0), 1)
+        self.assertEqual(Example.overloaded(0, extraArg2=0, extraArg1=0), 2)
+
+        with self.assertRaises(ValueError):
+            Example.overloaded(0, orderedArg2=0, orderedArg1=0)
```
freakboy3742, line 996: It would be worth adding a comment clarifying what is being tested here
```suggestion
        # Although the arguments are a unique match, they're not in the right order.
        with self.assertRaises(ValueError):
            Example.overloaded(0, orderedArg2=0, orderedArg1=0)
```

src/rubicon/objc/api.py
```diff
@@ -1035,28 +1041,11 @@ def __getattr__(self, name):
...
-            is preserved (:pep:`468`). In the future, once Rubicon requires
-            Python 3.6 or newer, "interleaved" method calls will respect keyword
-            argument order. This will fix the kind of conflict described above,
-            but will also disallow specifying the keyword arguments out of
-            order.
+        fully supported. If two arguments have the same label (e.g.
+        ``performSelector:withObject:withObject:``), you can use ``__`` in the
+        keywords like ``performSelector(..., withObject__1=...,
+        withObject__2=...)``. The parts after ``__`` can be anything you can use
+        as argument names.
```
freakboy3742, line 1048:
```suggestion
        fully supported. If two arguments have the same name (e.g.
        ``performSelector:withObject:withObject:``), you can use ``__`` in the
        keywords to disambiguate (e.g., ``performSelector(..., withObject__1=...,
        withObject__2=...)``. Any content after and including the ``__`` in an argument
        will be ignored.
```

src/rubicon/objc/api.py
```diff
@@ -246,28 +246,34 @@ def __repr__(self):
...
 
     def __call__(self, receiver, first_arg=_sentinel, **kwargs):
-        if first_arg is ObjCPartialMethod._sentinel:
+        # Ignore parts of argument names after "__".
+        order = [argname.split("__")[0] for argname in kwargs]
+        args = [arg for arg in kwargs.values()]
+
+        if first_arg is self._sentinel:
             if kwargs:
+                # This will be deleted when #26 is introduced.
```
freakboy3742, line 255: No need to flag this; #26 might not happen unless we can find a good implementation, and if that happens, the fact this isn't needed will become obvious.
```suggestion
```

src/rubicon/objc/api.py
```diff
@@ -246,28 +246,34 @@ def __repr__(self):
         return f"{type(self).__qualname__}({self.name_start!r})"
 
     def __call__(self, receiver, first_arg=_sentinel, **kwargs):
-        if first_arg is ObjCPartialMethod._sentinel:
+        # Ignore parts of argument names after "__".
+        order = [argname.split("__")[0] for argname in kwargs]
+        args = [arg for arg in kwargs.values()]
+
+        if first_arg is self._sentinel:
```
freakboy3742, line 253:
```suggestion
        if first_arg is ObjCPartialMethod._sentinel:
```
Although it's technically equivalent, referencing the sentinel on the class highlights that it is a class-level object (I think it will be *marginally* faster, as well)

src/rubicon/objc/api.py
```diff
@@ -246,28 +246,34 @@ def __repr__(self):
...
             if kwargs:
+                # This will be deleted when #26 is introduced.
                 raise TypeError("Missing first (positional) argument")
-
-            args = []
-            rest = frozenset()
+            rest = (*order,)
```
freakboy3742, line 257: This is technically correct, but isn't intuitive to read. I'd suggest creating the original `order` as a tuple comprehension; then this can be just `rest = order`, and the version in the else branch is `rest = ("",) + order`

changes/148.feature.rst
```diff
@@ -0,0 +1 @@
+Objective-C methods with multiple identical argument names can now be called using the "__" keyword and any suffix.
```
freakboy3742, line 1:
```suggestion
Objective-C methods with repeated argument names can now be called by using a "__" suffix in the keyword argument keyword to disambiguate the Python keyword arguments.
```

changes/453.feature.rst
```diff
@@ -0,0 +1 @@
+With "interleaved" syntax, the order of method arguments are now strictly checked.
```
freakboy3742, line 1: This should be a `.removal` change note, as it is a backwards incompatible change (`.removal` will be included the "backwards incompatible" section of the release notes).
```suggestion
The order of keyword arguments used when invoking methods must now match the order they are defined in the Objective-C API. Previously arguments could be in any order.
```

changes/461.feature.rst
```diff
@@ -0,0 +1 @@
+Updated the exception message when calling Objective-C methods with invalid keyword arguments to include selector names.
```
freakboy3742, line 1: The error message has been improved when an Objective-C selector cannot be found matching the provided arguments.

src/rubicon/objc/api.py
```diff
@@ -246,28 +246,34 @@ def __repr__(self):
         return f"{type(self).__qualname__}({self.name_start!r})"
 
     def __call__(self, receiver, first_arg=_sentinel, **kwargs):
-        if first_arg is ObjCPartialMethod._sentinel:
+        # Ignore parts of argument names after "__".
+        order = [argname.split("__")[0] for argname in kwargs]
```
freakboy3742, line 250: As mentioned below - given that `order` is only ever used as a tuple, it makes sense to construct it as one.
```suggestion
        order = tuple(argname.split("__")[0] for argname in kwargs)
```

freakboy3742 (changes requested):
This looks pretty good. I've flagged a couple of testing issues, some minor implementation cleanups, and some updates to the release notes. 

There's also a need for a documentation update to describe the `__1` syntax.

qqfunc:
Thanks for the reviews. I have made some changes as you say, and added a description to the documentation.

docs/tutorial/tutorial-1.rst
```diff
@@ -52,6 +52,22 @@ The second argument (``relativeToURL``) is accessed as a keyword argument. This
...
+    ...     SEL("someMethod:"),
+    ...     withObject__1=data1,
+    ...     withObject__2=data2
+    ... )
+
+In this example, the ``withObject`` arguments are duplicated, so the suffixes
+``__1`` and ``__2`` are added to avoid argument name collisions. Since ``__``
+and any part after it are ignored, it is possible to use any suffix starting
+with it.
+
```
freakboy3742: `performSelector` isn't a great example here, as it can be easily confused with the fact that we're discussing how to access selectors. Going "meta" in the tutorial is a potential source of confusion.

There's also the following paragraph, which contradicts this one; and the need for API documentation in addition to the tutorial.

freakboy3742 (approved):
Thanks for those fixes. I've made a couple of additional tweaks to the documentation and change notes; but otherwise, this looks great! 

Thanks for these fixes - the ambiguous argument problem has been one of those small annoying feature requests that I've wanted to fix for a while, but it's never been *quite* annoying enough to get to the top of my todo list.

qqfunc:
Thank you!

PR #464, opened 2024-05-12
Bump dependencies in pyproject.toml and tox.ini

brutusthebee:
Bumps versions for dependencies in pyproject.toml and tox.ini.


freakboy3742:
Failure is due to pypa/setuptools_scm#1038. This update should be fine, but we need to wait until the next setuptools-scm release before CI will pass.

freakboy3742 (approved):

PR #473, opened 2024-05-27
Remove pytest-tldr.

freakboy3742:
pytest-tldr is a cosmetic plugin that isn't requried for testing; it can be removed.

This reveals a weird Heisenbug in the handling of inherited classes, plus an issue with a test case pattern being overly restrictive for local testing. The Heisenbug is addressed by adding locking to superclass method loading. 

The original test failures on this PR don't have the extra locking; once locking is added, the tests pass.

## PR Checklist:

### DIFF
--- a/changes/473.bugfix.rst
+++ b/changes/473.bugfix.rst
@@ -1,0 +1,1 @@
+Protection was added against a potential race condition when loading methods defined on a superclass.
--- a/changes/473.misc.rst
+++ b/changes/473.misc.rst
@@ -1,0 +1,1 @@
+pytest-tldr was removed as a development requirement.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,6 @@ dev = [
     "pre-commit == 3.5.0 ; python_version < '3.9'",
     "pre-commit == 3.7.0 ; python_version >= '3.9'",
     "pytest == 8.2.1",
-    "pytest-tldr == 0.2.5",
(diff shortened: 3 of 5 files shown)


samschott (approved):
The changes look good to me! I'm just curious about why the race condition appeared only after removing `pytest-tldr`. Or is this just coincidental?

freakboy3742:
> The changes look good to me! I'm just curious about why the race condition appeared only after removing `pytest-tldr`. Or is this just coincidental?

As best as I can make out, pytest-tldr is doing something subtle that is altering garbage collection. However, I'm still not 100% clear on how this is happening.

The test that fails is a 100% consistent fail; but if you run the test by itself, it passes. I tried bisecting the rest of the tests to see if there was a set of tests that could run before the test that would make it fail; the closest I got was an almost random collection of ~15-20 tests - it wasn't "run another test using the same class" or anything like that. And if you put print statements in the `_load_methods()` call to print the state of the method cache to diagnose what is going on, the test passes.

I eventually worked out that it was a timing issue that was causing the method table for the intermediate object in the inheritance chain to be re-constructed; given that every other invocation of `_load_methods()` involves a lock, it seemed a reasonable guess that locking the inner `_load_methods()` might help - and it did.

I agree it's not a very satisfying answer. The tests aren't running multi-threaded, and there shouldn't be any garbage collection going on (AFAICT)... but the extra lock definitely fixes the problem.

PR #517, opened 2024-09-25
Remove workaround for python/cpython#81061.

freakboy3742:
Rubicon contains a ctypes_patch to work around python/cpython#49960, a limitation in returning structures or unions from a C function.

That workaround itself contains a workaround for python/cpython#81061. However, that issue was resolved by python/cpython#13364, which is available in Python 3.10+.

This PR modifies the use of the inner workaround so that it is only employed on older versions of Python. There are existing test cases to validate that the handling for the python/cpython#81061 case works.

The patch is structured so that when pyupgrade runs in ~October 2025 to remove Python 3.9 support, the removal of the conditional branches will leave working code.

## PR Checklist:

### DIFF
--- a/changes/517.bugfix.rst
+++ b/changes/517.bugfix.rst
@@ -1,1 +1,1 @@
-A workaround for python/cpython#81061 was conditionally removed when running under Python 3.9 or earlier.
+A workaround for `python/cpython#81061 <https://github.com/python/cpython/issues/81061>`__ is now conditionally applied only for the Python versions that require it (Python 3.9 and earlier).
--- a/changes/517.bugfix.rst
+++ b/changes/517.bugfix.rst
@@ -1,0 +1,1 @@
+A workaround for python/cpython#81061 was conditionally removed when running under Python 3.9 or earlier.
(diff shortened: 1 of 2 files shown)


changes/517.bugfix.rst
```diff
@@ -0,0 +1 @@
+A workaround for python/cpython#81061 was conditionally removed when running under Python 3.9 or earlier.
```
dgelessus, line 1: I think you mean something like this, right? (also added link to the CPython issue)

```suggestion
A workaround for `python/cpython#81061 <https://github.com/python/cpython/issues/81061>`__ is now conditionally applied only for the Python versions that require it (Python 3.9 and earlier).
```

dgelessus (commented):
Looks good to me, aside from the changelog entry. Thank you!

(I didn't bother removing this earlier, because the remainder of ctypes_patch is still necessary. But it does make sense to remove this inner workaround once rubicon-objc drops support for the affected Python versions.)

dgelessus (approved):

PR #540, opened 2024-11-12
Force cache eviction when instance classes don't match.

### LINKED ISSUE #539: Foundation collection instances can be cached without their helper wrappers
### Describe the bug

When an instance of a Foundation collection class is created (e.g., NSMutableDictionary), Rubicon uses a collection class (e.g., ObjCMutableDictInstance) to ensure that Pythonic helpers such as `__setitem__` are available. 

It is also possible to create an ObjCInstance directly from an objc_id. If this happens, it may be cached as a raw ObjCInstance, and *not* have the helper methods. 

Rubicon caches object instances based on the object pointer, performing cache eviction only when the class name of the underlying object class doesn't match. However, an NSMutableDictionary created directly, and one created from an objc_id will have the same pointer, and the same underlying class name. As a result, it's possible for a recycled collection class instance to *not* have the helper methods as a result of a cache hit.

### Steps to reproduce

I haven't worked out a simple way to reproduce this programmatically.

beeware/toga-chart#191 contains an example that can reproduce the issue with reasonable reliability. Run the example app, and resize the window continuously. I can usually trigger the exception reported by that ticket within a couple of seconds.

### Expected behavior

Cached ObjC collection instances should always be returned with the appropriate wrapper, even if the memory address has previously been used by an object of the same type.

### Screenshots

_No response_

### Environment

- Operating System: macOS
- Python version: Any
- Software versions:
  - Rubicon 0.4.9+

### Logs

```

```

### Additional context

The workaround fix is to modify the conditions for a cache eviction so that we don't *just* look at the class name, but at the class instance as well. 

Even better would be to ensure that a non-wrapped instance isn't ever cached in the first place... but I'm having difficulty working out how the non-wrapped instance is being created.

Even better still would be to fix the issue with memory-based cache hits (#256).

freakboy3742:
If a cache hit is found for a object pointer, check the underlying ObjC class name *and* the class instance. If *either* are a mismatch, report a cache miss. This ensures that collection types always have their Python wrappers applied, even if they are originally created (and cached) as a base ObjCInstance.

I haven't added a test for this because I haven't been able to find a way to reproduce the issue programmatically - it depends upon cache eviction behaviour, and the only place I've seen this problem occur is via beeware/toga-chart#191.

Fixes #539.
Fixes beeware/toga-chart#191.

## PR Checklist:

### DIFF
--- a/src/rubicon/objc/api.py
+++ b/src/rubicon/objc/api.py
@@ -657,7 +657,6 @@ def type_for_objcclass(objcclass):
     This method is mainly intended for internal use by Rubicon, but is exposed
     in the public API for completeness.
     """
-
     if isinstance(objcclass, ObjCClass):
         objcclass = objcclass.ptr
 
@@ -893,15 +892,27 @@ class ObjCInstance:
                 # it's the correct class.
                 #
                 # Refs #249.
+                #
+                # We also need to evict the cache when the instance *class*
+                # doesn't match. This can happen when an ObjCInstance subclass
+                # (like ObjCMutableDictionary) has been used - an ObjCInstance
+                # will respond to the same selectors as ObjCMutableDictionary,
+                # but it won't have the helper methods (like __setitem__) that
+                # make the wrapper useful.
+                #
+                # Refs #539.
+                #
                 if cls == ObjCInstance or isinstance(cls, ObjCInstance):
                     cached_class_name = cached.objc_class.name
-                    current_class_name = libobjc.class_getName(
-                        libobjc.object_getClass(object_ptr)
-                    ).decode("utf-8")
+                    current_class = libobjc.object_getClass(object_ptr)
+                    current_class_name = libobjc.class_getName(current_class).decode(
+                        "utf-8"
+                    )
+                    current_type = type_for_objcclass(current_class)
                     if (
-                        current_class_name != cached_class_name
-                        and not current_class_name.endswith(f"_{cached_class_name}")
-                    ):
+                        type(cached) is not current_type
+                        or current_class_name != cached_class_name
+                    ) and not current_class_name.endswith(f"_{cached_class_name}"):
                         # There has been a cache hit, but the object is a
                         # different class, treat this as a cache miss. We don't
                         # *just* look for an *exact* class name match, because
--- a/changes/539.bugfix.rst
+++ b/changes/539.bugfix.rst
@@ -1,0 +1,1 @@
+An edge case of the instance cache has been fixed. This edge case would cause collection types (such as NSMutableDictionary) to *not* have their Python helper wrappers applied.


mhsmith (approved):

src/rubicon/objc/api.py
```diff
@@ -893,15 +892,27 @@ class or a metaclass, an instance of :class:`ObjCClass` or
                 # it's the correct class.
                 #
                 # Refs #249.
+                #
+                # We also need to evict the cache when the instance *class*
+                # doesn't match. This can happen when an ObjCInstance subclass
+                # (like ObjCMutableDictionary) has been used - an ObjCInstance
+                # will respond to the same selectors as ObjCMutableDictionary,
+                # but it won't have the helper methods (like __setitem__) that
+                # make the wrapper useful.
```
samschott, line 901: I find this explanation bit hard to understand, especially the 1st sentence "when the instance *class* doesn't match". Are we talking about Python or ObjC classes here? I gather from the remaining text and the implementation that you mean Python classes. 

I'll leave a suggested rewrite, but feel free to reject or adapt. I understand that 'easy to understand' is very subjective :)

```
We also evict the cached instance when its Python type does not
match what we want to use to represent the given object in Python.
For example, we want to represent a NSDictionary instance as ObjCDictInstance to add Python dict API methods (like __setitem__). If we find a ObjCInstance instead in the cache, we evict it.
```

samschott (commented):
Can the reverse also happen? We already have the type that we like in the cache (e.g., ObjCDictInstance) and evict it in favour of one that we don't like (e.g., ObjCInstance)? Or is this prevented by all "type preferences" being correctly registered eventually and never being unregistered?

samschott:
Never mind that comment, I'm just catching up on #539. You believe the issue is caused by the memory address being reused by a different object.

src/rubicon/objc/api.py
freakboy3742: Agreed that's a better explanation of this particular implementation; what I'm trying to work out (in response to @mhsmith's queries) is whether there is a better way to frame the original cache hit based on the actual ObjC class data. It feels like there should be... but I've been going mad trying to work out what that formulation would be. All the obvious stuff seems to have weird edge case failures.

src/rubicon/objc/api.py
```diff
@@ -893,15 +892,27 @@ class or a metaclass, an instance of :class:`ObjCClass` or
...
+                        "utf-8"
+                    )
+                    current_type = type_for_objcclass(current_class)
                     if (
-                        current_class_name != cached_class_name
-                        and not current_class_name.endswith(f"_{cached_class_name}")
-                    ):
+                        type(cached) is not current_type
+                        or current_class_name != cached_class_name
+                    ) and not current_class_name.endswith(f"_{cached_class_name}"):
```
samschott, line 915: How about making "class_name" and the Python type part of the cache key? This seems conceptually cleaner than performing checks on the cached value and then raising a KeyError.

freakboy3742: The issue with using the name in the cache key is L915 - there are classes (NSWindow being one) that change their name after being initialised. If the original name is in the cache key, it won't get a cache hit when the "initialised" object is returned.

freakboy3742:
#543 is likely a better solution for this problem.

freakboy3742:
Closing on the basis that #543 is a much more comprehensive (and less hacky) solution to the problem.

PR #543, opened 2024-11-19
Updated memory management

### LINKED ISSUE #48: Improve memory management documentation

### LINKED ISSUE #256: Improve correspondence between ObjectiveC objects and Python wrappers
### What is the problem or limitation you are having?

In the process of building the Toga GUI testbed, we discovered #249. A fix for this problem was included in #246; however, that fix was really a safety catch for a specific set of circumstances. The bigger question: how did that set of circumstances happen in the first place?

At present, it is possible for an object to be disposed of by Objective C, but due to delays in propagating garbage collection, the `ObjCInstance` wrapping that object may not be immediately disposed of. This means the weak reference dictionary can contain stale objects.

#246 protected against this by protecting against the specific case where a memory address was *reused* - but if a Python object points at memory that has been disposed of and *not* reused, you'll generate a segfault. This is obviously undesirable.

### Describe the solution you'd like

Functionally - it should be impossible to generate a segfault from Python code.

An idea posited by @mhsmith in [the discussion about #249](https://github.com/beeware/rubicon-objc/issues/249#issuecomment-1399472658) was that we could maintain stronger reference to the ObjC object on the Python side. This would mean we add add a `retain` calls whenever a Python-side wrapper is created; `release`-ing only when *Python* has finished with the object. The idea here would be that it's better to have larger memory footprint (or even a small leak), than to have a segfault because a Python object has a stale memory reference.

### Describe alternatives you've considered

Do nothing. The status quo is stable(ish?), and while it, may be theoretically possible to get a stale ObjCInstance object, in practice we're not seeing that problem (after #246 landed, anyway).

### Additional context

Other than the general hesitation of fiddling with a complex area of code that *appears* to work at present - the complication to watch out for here is cyclic references between Python objects and Objective C objects that prevents Objective C from ever releasing memory. It's easy to avoid segfaults by never releasing... but eventually the memory will run out.

### LINKED ISSUE #539: Foundation collection instances can be cached without their helper wrappers

samschott:
This PR changes the memory management model of Rubicon. Previously, we would release objects on Python `__del__`  calls only if we created them ourselves and own them from alloc and similar calls.

In this PR, we always ensure that we own objects when we create a Python wrapper, by explicitly calling `retain` if we did not get them from `alloc` etc and always calling `autorelease` when the Python wrapper is garbage collected. This has a few advantages:

1. Users will no longer need to manually retain objects when receiving them from non-alloc methods and to release them before the Python object goes out of scope to avoid memory leaks.
2. Unless users manually release an object, Rubicon guarantees that a Python wrapper always points to an existing Objective-C object -- the one that it was created for.

This change should be backward compatible for most users because existing manual `retain` and `release` calls don't cause any issues if balanced and would have already caused segfaults if there are more releases than retains. 

TODO:

- [x] Update docs.
- [x] Test with toga.
- [x] Figure out a decent transition plan for clients.

* Fixes #256
* Fixes #539 
* Fixes #48
* Fixes beeware/toga-chart#191

## PR Checklist:


samschott:
@mhsmith, care to have first look? Please note the TODOs still listed above.

freakboy3742 (commented):
This turned out to be a lot less invasive than I thought it would be. 

I've done a check of the toga-chart issue that triggered this change; and it seems to resolve that problem. 

I also did a quick check with Toga's testbed suite on macOS; that code has a bunch of manual `retain`s and `autorelease`/`release`s. I *thought* they should all be balanced though - worst case, objects would be over-retained - so I was a little surprised that it the testbed segfaults almost immediately (and the stack trace doesn't give any obvious pointers what is causing the issue).

If I remove *all* the retains and releases, the testbed segfaults; but on inspection, some of the uses are for objects that are created in Python, then handed to ObjC to manage (e.g., Toolbar items created [here](https://github.com/beeware/toga/blob/main/cocoa/src/toga_cocoa/window.py#L108)), or the `copyWithZone` handler [here](https://github.com/beeware/toga/blob/main/cocoa/src/toga_cocoa/widgets/internal/data.py#L9)). But I guess those uses of memory handling make sense - and they're a lot closer aligned to the "spirit" of ObjC memory handling, Plus, in at least the ToolbarItem case, it could be avoided by keeping the toolbar instance in the cache of items.

freakboy3742:
Related - if we land this, I suspect a version bump to 0.5 might be called for. This is *just* backwards incompatible enough that I think it's worth flagging the significance of the change.

samschott:
Thanks for the thorough checks! I've updated the PR description to give a better summary of the change and also discuss why this should be non-breaking for most users.

I'll have a closer look at the segfaults that you encountered,  later. They might be caused by the usage of release instead of autorelease in __del__ which does not give Objective-C a chance to take over ownership.

Maybe there are also ways to prevent user to shoot themselves in the foot, e.g., raise an exception on manual release calls if there is only a single reference left.

mhsmith:
Thanks, this looks great. I'm busy today, but I'll take a look at this as soon as I can.

samschott:
I've had a closer look now at the segfaults and could identify two cases where they happen:

1. The object is released by Rubicon but still needed in ObjC, for example because it is being assigned to a property. Replacing `release` by `autorelease` in `__del__` fixes this.
2. Toga used to have a few "stray" `release` or `autorelease` calls sprinkled throughout the codebase to manually clean up memory, relying on Rubicon internally setting `_needs_release = False` after those calls and disabling its own cleanup logic. This no longer works and results in one too many release calls now.

https://github.com/beeware/toga/pull/2978 contains all the changes that I found to (1) prevent segfaults and (2) remove now unneeded manual memory management.

docs/how-to/memory-management.rst
```diff
@@ -16,16 +16,17 @@ Rubicon Objective-C operates at runtime, it cannot make use of ARC.
...
 -----------------------------------------
 
-You won't have to manage reference counts in Python, Rubicon Objective-C will do
-that work for you. It does so by tracking when you gain ownership of an object.
-This is the case when you create an Objective-C instance using a method whose
-name begins with ``alloc``, ``new``, ``copy``, or ``mutableCopy``. Rubicon
-Objective-C will then insert a ``release`` call when the Python variable that
-corresponds to the Objective-C instance is deallocated.
+In most cases, you won't have to manage reference counts in Python, Rubicon
+Objective-C will do that work for you. It does so by calling ``retain`` on an
```
mhsmith, line 20: The special-casing of ``alloc``, ``new``, ``copy``, or ``mutableCopy`` should still be mentioned, because if it was implemented as described without the special case, objects created with those methods would be leaked.

src/rubicon/objc/api.py
```diff
@@ -815,7 +794,9 @@ def objc_class(self):
     def _associated_attr_key_for_name(name):
         return SEL(f"rubicon.objc.py_attr.{name}")
 
-    def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
+    def __new__(
+        cls, object_ptr, _name=None, _bases=None, _ns=None, returned_from_method=b""
+    ):
```
mhsmith, line 799: The docstring paragraph starting "Rubicon currently does not perform any automatic memory management" should be updated.

changes/256.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Retain Objective-C objects when creating Python wrappers and release them when the
+Python wrapped is garbage collected.
```
mhsmith, line 2: > This change should be backward compatible for most users because existing manual retain and release calls don't cause any issues if balanced

Unfortunately, as you've discovered, existing calls may not be balanced if they depended on the previous `_needs_release` behavior. So we should increase the version number to 0.5.0 to prevent older Toga versions from seeing the new behavior, and make this a "removal" change note with advice on how to adapt to it.

And if we get this right, that advice should simply be "remove ALL instances of retain, release and autorelease from your code, and never worry about it again". If we can do that in Toga, then there's a pretty good chance that it will work for other clients as well.

src/rubicon/objc/api.py
```diff
@@ -815,7 +794,9 @@ def objc_class(self):
     def _associated_attr_key_for_name(name):
         return SEL(f"rubicon.objc.py_attr.{name}")
 
-    def __new__(cls, object_ptr, _name=None, _bases=None, _ns=None):
+    def __new__(
+        cls, object_ptr, _name=None, _bases=None, _ns=None, returned_from_method=b""
```
mhsmith, line 798: `returned_from_method` should have a leading `_` to indicate it's not part of the public API.

src/rubicon/objc/api.py
```diff
@@ -944,39 +873,11 @@ class or a metaclass, an instance of :class:`ObjCClass` or
...
-
     def __del__(self):
-        """Release the corresponding objc instance if we own it, i.e., if it
-        was returned by a method starting with :meth:`alloc`, :meth:`new`,
-        :meth:`copy`, or :meth:`mutableCopy` and it wasn't already explicitly
-        released by calling :meth:`release` or :meth:`autorelease`."""
-        if self._needs_release:
-            send_message(self, "release", restype=objc_id, argtypes=[])
+        # Autorelease our reference on garbage collection of the Python wrapper. We use
+        # release instead of autorelease allow ObjC to take ownership of an object when
```
mhsmith, line 878:
```suggestion
        # Autorelease our reference on garbage collection of the Python wrapper. We use
        # autorelease instead of release to allow ObjC to take ownership of an object when
```

tests/test_core.py
```diff
@@ -1859,94 +1869,80 @@ class TestO:
...
-        self.assertEqual(
-            attr0.retainCount(), 1, "strong property value was not released"
-        )
-        self.assertEqual(attr1.retainCount(), 1, "weak property value was released")
+        # Assert that the obj was deallocated.
+        self.assertIsNone(wr.weak_property)
+
+    # def test_objcinstance_dealloc(self):
+    #     class DeallocTester(NSObject):
+    #         attr0 = objc_property()
```
mhsmith, line 1919: Commented-out test should be cleaned up.

mhsmith (commented):
I've added #539 and #48 to the Fixes list in the top comment.

tests/test_core.py
samschott: Darn, I forgot about this one. I did not want to remove this test, just update it. The  old "method of testing" now causes a segfault at the end. Updated now.

src/rubicon/objc/api.py
samschott: Done. Apologies for that, it seems I forgot some of the planned "cleanup" when switching this PR from draft to "ready for review".

src/rubicon/objc/api.py
samschott: Done.

docs/how-to/memory-management.rst
samschott: I wasn't quite sure whether that level of detail is needed here. Updated now.

changes/256.bugfix.rst
samschott: Indeed, this will require migration advice in the release notes. Will add that.

Regarding the new version number, how does this work? Just tag a commit? You seam to have something fancy going on these days with setuptools_scm.

changes/256.removal.rst
```diff
@@ -0,0 +1,4 @@
+Rubicon no longer skips releasing an Objective-C objects when its Python wrapped is
```
mhsmith, line 1:
```suggestion
Manual calls to ``release`` or ``autorelease`` no longer cause Rubicon  
to skip releasing an Objective-C object when its Python wrapper is
```

src/rubicon/objc/api.py
```diff
@@ -814,10 +814,17 @@ class or a metaclass, an instance of :class:`ObjCClass` or
...
+        Rubicon retains an Objective-C object when it is wrapped in an
+        :class:`ObjCInstance` and autoreleases it when the :class:`ObjCInstance` is
+        garbage collected.
+
+        The only exception to this are objects returned by methods which create an
+        object (starting with "alloc", "new", "copy", or "mutableCopy"). We do not
+        explicitly retain them because we already own objects created by us, but we do
+        autorelease them on garbage collection of the Python wrapper.
+
+        This ensures that the :class:`ObjCInstance` can always be  used from Python
```
mhsmith, line 826:
```suggestion
        This ensures that the :class:`ObjCInstance` can always be used from Python
```

changes/256.bugfix.rst
```diff
@@ -1,2 +1,10 @@
 Retain Objective-C objects when creating Python wrappers and release them when the
-Python wrapped is garbage collected.
+Python wrapped is garbage collected. This means that manual ``retain`` calls and
+subsequent ``release`` or ``autorelease`` calls from Python are no longer needed with
```
mhsmith, line 3: I think this change note counts as a "feature" rather than a "bugfix".

changes/256.bugfix.rst
mhsmith: Yes, nothing is required in the PR itself; we would tag the main branch after it's merged.

mhsmith: Yes, nothing is required in the PR itself; we would tag the main branch after it's merged.

tests/test_core.py
```diff
@@ -1859,76 +1869,64 @@ class TestO:
...
-        self.assertTrue(obj._needs_release)
-
-        # Explicitly release the object.
-        obj.release()
-
-        # Check that we no longer need to release it.
-        self.assertFalse(obj._needs_release)
+    def test_objcinstance_created_retained(self):
+        with autoreleasepool():
+            # Create an object which we don't own.
```
mhsmith, line 1874: What's the purpose of the `autoreleasepool` here? It doesn't ensure the object is released, because it doesn't cover the end of the Python wrapper's lifetime.

I think it would make sense to combine the "retained" tests with the "gc_released" tests. Then we'd have two tests, each one covering the full lifetime of an object:

* test_objcinstance_created, based on `alloc`,
* test_objcinstance_retained, based on `stringWithString`

tests/test_core.py
```diff
@@ -1859,76 +1869,64 @@ class TestO:
...
-    def test_objcinstance_retain_release(self):
-        NSString = ObjCClass("NSString")
+        with autoreleasepool():
+            del obj
+            gc.collect()
 
-        # Create an object which we don't own.
-        string = NSString.stringWithString("test")
+        # Assert that the obj was deallocated.
+        self.assertIsNone(wr.weak_property)
```
mhsmith, line 1899: We should also verify that the object is *not* deallocated between `gc.collect` and the end of the `autoreleasepool`. This distinguishes between `release` and `autorelease`.

tests/test_core.py
```diff
@@ -1940,9 +1938,11 @@ def dealloc(self):
         self.assertEqual(attr1.retainCount(), 1)
 
         # ObjC object will be deallocated, can only access Python attributes afterwards.
```
mhsmith, line 1940: Comment should be reworded or removed.

src/rubicon/objc/api.py
```diff
@@ -944,39 +880,11 @@ class or a metaclass, an instance of :class:`ObjCClass` or
...
     def __del__(self):
-        """Release the corresponding objc instance if we own it, i.e., if it
-        was returned by a method starting with :meth:`alloc`, :meth:`new`,
-        :meth:`copy`, or :meth:`mutableCopy` and it wasn't already explicitly
-        released by calling :meth:`release` or :meth:`autorelease`."""
-        if self._needs_release:
-            send_message(self, "release", restype=objc_id, argtypes=[])
+        # Autorelease our reference on garbage collection of the Python wrapper. We use
+        # autorelease instead of release to allow ObjC to take ownership of an object when
+        # it is passed for example to a strong property.
```
mhsmith, line 886: Passing an ObjCInstance to a property setter wouldn't destroy the Python wrapper.

```suggestion
        # it is returned from a factory method.
```

changes/256.bugfix.rst
samschott: Done.

tests/test_core.py
samschott: Replaced now with a comment to explain why we need the autorelease pool.

tests/test_core.py
samschott: Good point. Done.

tests/test_core.py
samschott: > What's the purpose of the autoreleasepool here? It doesn't ensure the object is released, because it doesn't cover the end of the Python wrapper's lifetime.

The autorelease pool is needed to ensure that not only the ObjCInstance is garbage collected (which triggers an autorelease) but also the wrapped Objective-C object is deallocated, which happens only after the pool is emptied. Without it, we get an
```
AssertionError: 2 != 1 : strong property value was not released`.
```
I'll add a comment about this.

> I think it would make sense to combine the "retained" tests with the "gc_released" tests. Then we'd have two tests, each one covering the full lifetime of an object.

I quite like separate test cases because failures will more directly point the relevant problem. But in this case, if the initial retain does not work, both tests will fail anyways, so the benefit is not that large.

Merged those test cases now.

freakboy3742:
@samschott FYI - I'm on the last day of PyCon AU sprints; I might not get a chance to look at this until tomorrow when I'm back in my office for a return to semi-normal business.

src/rubicon/objc/api.py
```diff
@@ -840,15 +846,22 @@ class or a metaclass, an instance of :class:`ObjCClass` or
...
                 # same object.
-                return cls._cached_objects[object_ptr.value]
+                cached_obj = cls._cached_objects[object_ptr.value]
+
+                # If a cached instance was returned from a call such as `copy` or
+                # `mutableCopy`, we take ownership of an additional refcount. Release
+                # it here to prevent leaking memory, Python already owns a refcount from
+                # when the item was put in the cache.
+                if _returned_from_method.startswith(_OWNERSHIP_METHOD_PREFIXES):
+                    send_message(object_ptr, "release", restype=objc_id, argtypes=[])
```
samschott, line 856: I worry a bit that the mental load of following when we retain and release is a bit much now since the reader needs to think though the `__new__` method being invoked multiple times with the same pointer.

Alternatives such as explicitly tracking a "Python refcount" might be easier to understand but would also add complexity.

Suggestions are welcome.

samschott:
There is yet another hairy issue with the current implementation which will lead to memory leaks: `alloc().init()` chains where init changes the objects memory address, as documented at https://developer.apple.com/documentation/objectivec/nsobject/1418641-init?language=objc:

> In some cases, a custom implementation of the init method might return a substitute object. You must therefore always use the object returned by init, and not the one returned by [alloc](https://developer.apple.com/documentation/objectivec/nsobject/1571958-alloc?language=objc) or [allocWithZone:](https://developer.apple.com/documentation/objectivec/nsobject/1571945-allocwithzone?language=objc), in subsequent code.

This is the case for example with `NSDictionary.alloc().init()` where the init returns a different object. The catch is that the object returned by `init` in this case seems to be owned by the caller as well (not only the object returned by alloc) and we start leaking memory. Example code with the implementation from this PR:

```python
from rubicon.objc import *
from rubicon.objc.runtime import autoreleasepool


with autoreleasepool():
  adict = NSDictionary.alloc()
  idict = adict.initWithObjects(["some_object"], forKeys=["some_key"])

assert idict.retainCount() == 2  # Retain it manually by Rubicon and implicitly.
print(adict)  # segaults because adict was deallocated
```

The above has two problems:

1. It leads to segfaults if the user attempts to use the allocated object after the autorelease pool was drained.
2. It leaks memory because the initialized object is retained twice but Rubicon is only aware of one of the retains.

src/rubicon/objc/api.py
```diff
@@ -91,6 +91,10 @@
 # the Python objects are not destroyed if they are otherwise no Python references left.
 _keep_alive_objects = {}
 
+# Methods that return an object with is implicitly retained by the caller.
```
mhsmith, line 94:
```suggestion
# Methods that return an object which is implicitly retained by the caller.
```

tests/test_core.py
```diff
@@ -1937,7 +1960,9 @@ def dealloc(self):
         self.assertEqual(attr0.retainCount(), 2)
         self.assertEqual(attr1.retainCount(), 1)
 
-        # ObjC object will be deallocated, can only access Python attributes afterwards.
+        # Delete the Python wrapper and ensure that the Objective-C object is
+        # deallocated after ``autorelease`` on garbage collection. This will also
+        # rigger a decrement in the retain count of attr0.
```
mhsmith, line 1965:
```suggestion
        # trigger a decrement in the retain count of attr0.
```

src/rubicon/objc/api.py
```diff
@@ -768,6 +772,8 @@ class ObjCInstance:
     # Refs #251.
     _instance_lock = threading.RLock()
 
+    _python_refcount = 0
+
```
mhsmith, line 776: Looks like this is a leftover from a previous approach.

```suggestion

```

src/rubicon/objc/api.py
mhsmith: I'm fine with this approach: if it's possible to deal with the situation immediately after the `copy`, that's definitely easier to follow than maintaining an extra refcount.

samschott:
Looks like my assessment from https://github.com/beeware/rubicon-objc/pull/543#issuecomment-2501626701 was not quite correct:

The `adict` object does not seem to be deallocated, the `print` statement seems to segfault for other reasons.

mhsmith:
>  `NSDictionary.alloc().init()` where the init returns a different object
>
> The above has two problems:
> 
> 1. It leads to segfaults if the user attempts to use the allocated object after the autorelease pool was drained.
> 2. It leaks memory because the initialized object is retained twice but Rubicon is only aware of one of the retains.

I guess then we'll also have to special-case methods starting with `init`. If `init` returns a different object to the one it was called on, then instead of creating a new ObjCInstance, replace the existing instance's pointer with that one and return it.

> The `adict` object does not seem to be deallocated, the `print` statement seems to segfault for other reasons.

I wouldn't worry too much about that object. It might be deallocated, or it might be an "empty dictionary" placeholder that gets recycled by every call to `NSDictionary.alloc`, but either way, we have no right to access it after calling `init`.

freakboy3742:
> There is yet another hairy issue with the current implementation which will lead to memory leaks: `alloc().init()` chains where init changes the object's memory address, as documented at https://developer.apple.com/documentation/objectivec/nsobject/1418641-init?language=objc:

Interesting... I wonder if *this* was the thing triggering #539...

> > `NSDictionary.alloc().init()` where the init returns a different object
> > The above has two problems:
> > 
> > 1. It leads to segfaults if the user attempts to use the allocated object after the autorelease pool was drained.
> > 2. It leaks memory because the initialized object is retained twice but Rubicon is only aware of one of the retains.
> 
> I guess then we'll also have to special-case methods starting with `init`. If `init` returns a different object to the one it was called on, then instead of creating a new ObjCInstance, replace the existing instance's pointer with the new one and return it.

To make sure we're on the same page, AIUI, your proposal is extra handling in `ObjCBountMethod.__call__` that will be invoked if `self.method.name.startswith("init")`; this logic will:
1. check that `self.receiver` (the object on which the method is invoked) is the same value as the return value of the method; 2. If it isn't the same pointer, `__call__` will modify the ObjCInstance instance itself, and the ObjCInstance cache to reflect the new value, releasing the "old" version from the alloc 
3. If it is the same pointer, it does nothing and just returns the pointer as it normally would.

That sounds broadly reasonable to me; I have 2 questions/concerns:

1. Are there "init" methods that don't start with `init`? I can't think of any examples off the top of my head, but that might be my memory fading with age. 

2. Is there a particular reason you suggest patching the old instance, rather than creating a new instance? It seems like this would be less complex to implement (possibly even the default behavior?), with the only downsides being the cost of creating a second Python object, and the fact that the alloc object and the init object aren't the same - but that mirrors the underlying ObjC behavior, so it strikes me as the sort of thing that can be documented.

src/rubicon/objc/api.py
freakboy3742: I agree. An extra refcount starts to get into re-implementing garbage collection territory, and I'm not sure that's a game that is worth it. I'm comfortable with this as a known edge case, and documenting it with some related usage patterns and advice.

freakboy3742:
> I guess then we'll also have to special-case methods starting with `init`. If `init` returns a different object to the one it was called on, then instead of creating a new ObjCInstance, replace the existing instance's pointer with the new one and return it.

Something that just occurred to me in reviewing beeware/toga#2978 in the context of this change - does this also resolve the NSImage "init fail" issue? The return value from init will be different to the alloc'd object... so the alloc'd object should be released; the only catch is that we *don't* need to create the ObjCInstance because it's a None object.

freakboy3742 (commented):
This is looking pretty good to me. Toga has an existing <0.5 pin, so we're safe to do a major update here without breaking Toga; and beeware/toga#2978 is a pretty clear "Delete all your manual retain/release" PR, except for one `copy` edge case, and the `NSImage` init failure case (which, AFAICT, will be solved if we address the "init may change address" issue.

src/rubicon/objc/api.py
```diff
@@ -852,68 +844,24 @@ class or a metaclass, an instance of :class:`ObjCClass` or
...
-                        # some Cocoa/UIKit classes undergo a class name change
-                        # between `alloc()` and `init()` (e.g., `NSWindow`
-                        # becomes `NSKVONotifying_NSWindow`). Refs #257.
-                        raise KeyError(object_ptr.value)
-
-                return cached
+                cached_obj = cls._cached_objects[object_ptr.value]
+
+                # If a cached instance was returned from a call such as `copy` or
+                # `mutableCopy`, we take ownership of an additional refcount. Release
```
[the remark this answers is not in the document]
samschott: Done.

src/rubicon/objc/api.py
samschott: Ok, then I'll stick with this :)

samschott:
> Something that just occurred to me in reviewing https://github.com/beeware/toga/pull/2978 in the context of this change - does this also resolve the NSImage "init fail" issue?

If we do things right, i.e., the way that ARC would handle it, it would indeed remove the need for special handling on the NSImage "init failed" issue.

> but that mirrors the underlying ObjC behavior, so it strikes me as the sort of thing that can be documented.

Agreed, I think we would represent different objects by different Python wrappers instead of swapping out the pointer under the hood.

samschott:
I've looked through ARC docs and found the relevant places to explain the `init` behavior: https://clang.llvm.org/docs/AutomaticReferenceCounting.html#semantics-of-method-families.

Basically, all methods in the  `alloc`, `copy`, `mutableCopy`, `new` and `init` families return a retained object.

`init` is special because it _consumes_ its `self` parameter. I.e., its takes ownership of it. This is what allows it to either (1) pass on ownership to the caller of `init` when returning or (2) release the object and return a different retained object.

Getting those semantics wrong seems to me the root cause of https://github.com/beeware/toga/pull/2978.

freakboy3742:
> I've looked through ARC docs and found the relevant places to explain the `init` behavior: https://clang.llvm.org/docs/AutomaticReferenceCounting.html#semantics-of-method-families.
> 
> Basically, all methods in the `alloc`, `copy`, `mutableCopy`, `new` and `init` families return a retained object.
> 
> `init` is special because it _consumes_ its `self` parameter. I.e., its takes ownership of it. This is what allows it to either (1) pass on ownership to the caller of `init` when returning or (2) release the object and return a different retained object.

Nice find - that's a helpful reference. Probably worth adding a link to that in a docstring somewhere as well.

If I'm reading that right (and especially the "Rationale" block"):
1. We need to strip leading underscores before the check against `_OWNERSHIP_METHOD_PREFIXES`
2. There's an edge case where we should be looking for the `objc_method_family` attribute to determine if a method is in one of these special families
3. We *are* safe to assume "starts with init, takes an `id`, returns an `id`" is a method that needs to do an extra release

(1) seems like an edge case, but it's also not hard to accomodate. (2) also seems like an edge case, but one that's a little harder to implement, so if that was bumped to a "known limitation" with documentation and open feature ticket, I'd be OK with that.

freakboy3742:
> 2. Is there a particular reason you suggest patching the old instance, rather than creating a new instance? 

Speaking with @mhsmith in person, he pointed out that if we create a new Python object for the `init`, the `alloc` Python object will continue to be a valid Python reference, but one that will be explosive if used. That's completely consistent with the ObjC behavior, but would be at surprising behavior to a Python user. Replacing the pointer reference on the existing object avoids this.

This is all true; my question is how much we care, balanced against the complexity and efficiency of the implementation.

Rubicon is a Python wrapper around ObjC; I don't think it's unreasonable to expect our users to need a basic understanding of ObjC, especially if we document these edge cases. "Don't use the alloc ptr once it's been init'd" is one of those edge cases. And the use case for keeping a pointer to the `alloc` independent of the `init` is niche at best.

The only thing that gives me pause is the performance aspect. If every `.alloc().init()` causes 2 Python objects to be created, that is going to have a direct performance overhead. If it will only be 2 Python objects for objects that change pointer, that's a lot less of a concern. 

If the fix to rewrite the pointer for those cases is <10 lines of code with no real other risks, then we might as well take the win. But if it's more complex than that, or there's any question over whether there are edge cases in getting the caching right, then the simple solution of "just create 2 objects and document the limitation" is something I can live with.

samschott:
> Speaking with @mhsmith in person, he pointed out that if we create a new Python object for the init, the alloc Python object will continue to be a valid Python reference, but one that will be explosive if used.

That's a good point. Sending messages to an uninitialized object will segfault for most selectors (with a few useful exceptions such as `init`). But I don't think this pitfall can be avoided completely. If the user chooses to do a standalone `alloc()` instead of `alloc().init()`, they can still run use an uninitialized object.

Another option could be explicitly track / flag uninitialized objects.

> The only thing that gives me pause is the performance aspect. If every .alloc().init() causes 2 Python objects to be created, that is going to have a direct performance overhead. If it will only be 2 Python objects for objects that change pointer, that's a lot less of a concern.

I am trying out a solution at the moment that would indeed only create 2 Python objects if the pointer changes, keeping with what ObjCInstances are supposed to represent. I am however still running into test failures in Python 3.12 which are a bit concerning.

> If I'm reading that right (and especially the "Rationale" block") [...]

I think you are reading this correctly. Though my first stab is certainly less sophisticated than that.

mhsmith:
> AIUI, your proposal is extra handling in `ObjCBountMethod.__call__` that will be invoked if `self.method.name.startswith("init")`

I'm not familiar enough with Rubicon to suggest the best way to implement this, but yes, this is generally what I had in mind.

> this logic will:
> 
> 1. check that `self.receiver` (the object on which the method is invoked) is the same value as the return value of the method; 2. If it isn't the same pointer, `__call__` will modify the ObjCInstance instance itself, and the ObjCInstance cache to reflect the new value, releasing the "old" version from the alloc

Nothing should be `release`d, because the old pointer was already "consumed" by the `init` method.

> Are there "init" methods that don't start with `init`?

It looks like this is answered by the specification Sam found, along the converse question of "are there methods that do start with `init` but shouldn't have special behavior".

> does this also resolve the NSImage "init fail" issue? The return value from init will be different to the alloc'd object... so the alloc'd object should be released; the only catch is that we _don't_ need to create the ObjCInstance because it's a None object.

Again, we shouldn't do another `release`, because `init` has already done that. But we should set the ObjCInstance's internal pointer to NULL, and make sure we handle that NULL pointer in the future as follows:
* Attempting to call a method on the object or use it in any way should throw an exception.
* When the object's `__del__` method is called, it should *not* be autoreleased.

mhsmith:
> 1. We need to strip leading underscores before the check against `_OWNERSHIP_METHOD_PREFIXES`

It's a little more complicated than that:

> A selector is in a certain selector family if, ignoring any leading underscores, the first component of the selector either consists entirely of the name of the method family or it begins with that name followed by a character other than a lowercase letter. For example, `_perform:with:` and `performWith:` would fall into the `perform` family (if we recognized one), but `performing:with` would not.

Or for a more realistic example, any method beginning with the string `initialize` is *not* in the `init` family.

The spec also gives a few other conditions for deciding whether a method is in a family, but I don't know whether it's possible to check them at runtime, or how likely they are to matter in practice.

samschott:
> The spec also gives a few other conditions for deciding whether a method is in a family, but I don't know whether it's possible to check them at runtime, or how likely they are to matter in practice.

I don't have enough experience with Objective-C to know this either, so I've implemented the name-base checks for now. Name-based rules seem sufficient for _selectors_ families, which is required but not sufficient for the _method_ family.

I've found what I believe is a relatively lightweight solution now that does not change ObjCInstance pointers under the hood and keeps the current semantics of "if we have Python wrapper, we Obj-C object must still exist". Admittedly, this object is much less useful when not yet initialized, but that's a problem that I don't want to solve for in this PR.

Why I like the current current solution:

1. It has upholds clear promises in the API: A `ObjCInstance` instance always refers to the same wrapped Objective-C object and this object is never deallocated while we still have a Python wrapper. This will hopefully make it easier to reason about future issues.
2. It treats all methods that return a retained object (`{"init", "alloc", "new", "copy", "mutableCopy"}`) equally when putting in them cache.
3. It has special handling for `init` consuming its `self` argument.

src/rubicon/objc/api.py
```diff
@@ -1627,17 +1571,22 @@ def _load_methods(self):
 
         methods_ptr = libobjc.class_copyMethodList(self, byref(methods_ptr_count))
 
-        if self.superclass is not None:
-            if self.superclass.methods_ptr is None:
-                with self.superclass.cache_lock:
-                    self.superclass._load_methods()
+        # Traverse superclasses and load methods.
```
samschott, line 1574: I am not entirely sure while a full traversal of superclasses is required now but wasn't previously for tests on Python 3.12 + macOS 15 to pass. But I do believe that a class hierarchy travels makes sense regardless.

src/rubicon/objc/api.py
```diff
@@ -847,73 +860,32 @@ class or a metaclass, an instance of :class:`ObjCClass` or
...
-                        raise KeyError(object_ptr.value)
-
-                return cached
+                cached_obj = cls._cached_objects[object_ptr.value]
+
+                # A `copy` call can return the original object if it is immutable. This
+                # is typically done for optimization. If this object is already in our
+                # cache, we take ownership of an additional reference. Release it here
+                # to prevent leaking memory.
+                # See https://developer.apple.com/documentation/foundation/nscopying.
```
freakboy3742, line 876: Should this comment also mention the "`init` returning the same object" edge case?

src/rubicon/objc/api.py
```diff
@@ -91,6 +91,21 @@
...
+
+def _get_method_family(method_name: str) -> str:
+    # See https://clang.llvm.org/docs/AutomaticReferenceCounting.html#method-families.
+    method_name = method_name.lstrip("_").split(":")[0]
+    leading_lowercases = []
+    for c in method_name:
+        if c.isupper():
+            break
+        leading_lowercases.append(c)
+    return "".join(leading_lowercases)
```
freakboy3742, line 107: The implementation here looks fine; since we can't guarantee we're actually exercising it fully, we should probably add some explicit unit tests that we're catching all the possible naming strategies that it should match.

src/rubicon/objc/api.py
freakboy3742: Agreed this is weird; I can only assume we're hitting some weird cache initialisation order thing (e.g., the test was previously initializing the super class before the class that was causing a problem). However, the solution here makes sense.

tests/test_core.py
```diff
@@ -1409,7 +1415,8 @@ class Ivars(NSObject):
         self.assertEqual(r.size.height, 78)
 
     def test_class_properties(self):
-        """A Python class can have ObjC properties with synthesized getters and setters."""
+        """A Python class can have ObjC properties with synthesized getters and setters
+        of ObjCInstance type ."""
```
freakboy3742, line 1419: Minor typo:
```suggestion
        of ObjCInstance type."""
```

tests/test_core.py
```diff
@@ -1859,76 +1870,127 @@ class TestO:
...
+        """
+        with autoreleasepool():
+            # Return an object which we don't own. Using str(uuid) here instead of a
+            # common string ensure that we get have the only reference.
+            obj = NSString.stringWithString(str(uuid.uuid4()))
 
-        # Check that it is marked for release.
-        self.assertTrue(obj._needs_release)
+        # Check that the object is retained when we create the ObjCInstance.
+        self.assertEqual(obj.retainCount(), 1, "object was not retained")
```
freakboy3742, line 1883: The assertion doesn't seem to match the comment; should this be:
```suggestion
        self.assertEqual(obj.retainCount(), 1, "object was retained")
```
or does the comment need to be clarified?

changes/256.feature.rst
```diff
@@ -0,0 +1,10 @@
+Retain Objective-C objects when creating Python wrappers and release them when the
+Python wrapped is garbage collected. This means that manual ``retain`` calls and
+subsequent ``release`` or ``autorelease`` calls from Python are no longer needed with
+very few exceptions such as:
+
+1. When implementing methods like ``copy`` that are supposed to create an object, if
+   the returned object is not actually newly created.
+2. When dealing with side effects of methods like ``init`` that may release an object
+   which is still referenced from Python. See for example
+   https://github.com/beeware/toga/issues/2468.
```
freakboy3742, line 10: Is this exception still required?

freakboy3742 (changes requested):
This looks really good to me. A couple of comments inline, mostly about the comments :-)

freakboy3742:
> I've found what I believe is a relatively lightweight solution now that does not change ObjCInstance pointers under the hood and keeps the current semantics of "if we have Python wrapper, we Obj-C object must still exist". Admittedly, this object is much less useful when not yet initialized, but that's a problem that I don't want to solve for in this PR.
> 
> Why I like the current current solution:

Agreed - I like the solution you've presented here. It may still have some interesting behavior if you try to use an alloc'd object... but then, so will ObjC, so it shouldn't be surprising.

src/rubicon/objc/api.py
samschott: Done.

changes/256.feature.rst
samschott: Indeed, no longer needed.

tests/test_core.py
samschott: The message is printed as an error message in case of an assertion failure, so it should be the "opposite" of the inline comment.

src/rubicon/objc/api.py
samschott: Good point.

src/rubicon/objc/api.py
samschott: Running the toga testbed suite reveals [a similar issue](https://github.com/beeware/toga/actions/runs/12088971183/job/33713572376?pr=2978) for `TogaSlider.setValue_animated_` where the superclass method cannot be found. This is despite the fix here and could be related to the TogaSlider being a subclass defined in Python.

samschott: Note that the direct method call `TogaSlider.setValue_animated(value, True)` succeeds.

samschott: I've managed to fix that by allowing direct method lookup if there is no cached partial method, similar to what we already do for the old style syntax.

This is very likely a race condition somewhere, but the code around `ObjCClass._load_methods`, `ObjCClass._cache_method` and `ObjCPartialMethod` quite complex and hard to follow.

tests/test_core.py
freakboy3742: 🤦 Of course.

src/rubicon/objc/api.py
samschott: But all those changes to method lookup are making this PR a bit unwieldy. I've reverted some of the unneeded commits and am happy to split this off into an entirely different PR if that makes it easier.

freakboy3742 (approved):
I'm happy with this PR as it stands now; in addition to solving the originally reported problem (beeware/toga-chart#191), it solves the bigger problem of memory management (#256) problem; it addresses the remaining iOS leakage issues (see beeware/toga#2853); and apparently also fixes a method caching issue that has been unreported to date. 

I've left one comment about the argument used when constructing arguments, but that's a minor cleanup suggestion, and one that I'll gladly back down from if you disagrees.

Before we merge, I'd also like @mhsmith's final review in case he can think of any edge cases or other issues I might have missed.

src/rubicon/objc/api.py
```diff
@@ -223,12 +275,7 @@ def __call__(self, receiver, *args, convert_args=True, convert_result=True):
 
         # Convert result to python type if it is an instance or class pointer.
         if self.restype is not None and issubclass(self.restype, objc_id):
-            result = ObjCInstance(result)
-
-        # Mark for release if we acquire ownership of an object. Do not autorelease here because
-        # we might retain a Python reference while the Obj-C reference goes out of scope.
-        if self.name.startswith((b"alloc", b"new", b"copy", b"mutableCopy")):
-            result._needs_release = True
+            result = ObjCInstance(result, _returned_from_method=self.name)
```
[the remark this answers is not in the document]
samschott: Passing boolean makes sense to me. It's the kind of thing you see in review rather than going through iterations of the code yourself :)

src/rubicon/objc/api.py
samschott: Well, something is still very fishy about my solution in this PR. `_load_methods` should have already been recursive, with the main difference that is was hard-stopping the recursion if one of the superclasses had already called `_load_methods`. This makes assumptions of the superclass chain not changing during `init`.

The new solution forces recursion to continue, bad in a horribly hacky way.

samschott: Ok, I have had another look at all those recursive calls and I think that (1) I do understand better now what is going on and (2) have a more elegant solution than this PR to method loading fixes (https://github.com/samschott/rubicon-objc/tree/method-loading).

But this is indeed better reviewed in isolation, so I am happy to proceed with this PR as is and send out an independent PR for stylistic improvements and hopefully clearer code logic.

src/rubicon/objc/api.py
```diff
@@ -214,6 +252,17 @@ def __call__(self, receiver, *args, convert_args=True, convert_result=True):
         else:
             converted_args = args
 
+        # Init methods consume their `self` argument (the receiver), see
+        # https://clang.llvm.org/docs/AutomaticReferenceCounting.html#semantics-of-init.
+        # To avoid segfaults on garbage collection if `init` does not return `self` but
```
mhsmith, line 257: This affects not just garbage collection, but any subsequent use of the object.

```suggestion
        # To avoid segfaults on garbage collection if `init` does not return `self` but
        # To ensure the receiver pointer remains valid if `init` does not return `self` but
```

src/rubicon/objc/api.py
```diff
@@ -897,7 +967,8 @@ def __del__(self):
         # Autorelease our reference on garbage collection of the Python wrapper. We use
         # autorelease instead of release to allow ObjC to take ownership of an object when
         # it is returned from a factory method.
-        send_message(self, "autorelease", restype=objc_id, argtypes=[])
+        if send_message and objc_id:
```
mhsmith, line 970: How could `send_message` or `objc_id` ever be unavailable?

tests/test_core.py
```diff
@@ -1954,6 +1948,37 @@ def copyWithZone_(self):
...
+
+            self.assertIsNotNone(
+                wr.weak_property,
+                "object was deallocated before end of autorelease pool",
+            )
+
+        self.assertIsNone(wr.weak_property, "object was not deallocated")
+
+    def test_objcinstance_init_none(self):
+        """We do segfault if init returns a different object than it received in alloc."""
```
mhsmith, line 1976:
```suggestion
        """We do not segfault if init returns nil."""
```

mhsmith (commented):
I like this solution as well. Just a few more comments:

src/rubicon/objc/api.py
samschott: During teardown of the Python interpreter, for example after calling `exit()`, see warning in the  [\_\_del\_\_](https://docs.python.org/3/reference/datamodel.html#object.__del__) documentation. Maybe checking for `None` is not the best way to handle this since it is (1) not race free and (2) does not handle if those objects are deleted instead of set to None.

Catching `NameError` (object was deleted) and `TypeError` (object was set to None) may be a safer in this regard.

src/rubicon/objc/api.py
samschott: Updated now, thanks!

src/rubicon/objc/api.py
```diff
@@ -91,9 +91,47 @@
...
+
+def get_method_family(method_name: str) -> str:
+    """Returns the method family from the method name. See
+    https://clang.llvm.org/docs/AutomaticReferenceCounting.html#method-families for
+    documentation on method families and corresponding selector names."""
+    method_name = method_name.lstrip("_").split(":")[0]
+    leading_lowercases = []
+    for c in method_name:
+        if c.isupper():
+            break
```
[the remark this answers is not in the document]
samschott: Darn, thanks for catching that! Reading the ARC docs again, it looks like we have no option than to explicitly check for the method name starting with a family string:

> A selector is in a certain selector family if, ignoring any leading underscores, the first component of the selector either consists entirely of the name of the method family or it begins with that name followed by a character other than a lowercase letter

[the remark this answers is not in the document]
samschott: > To cover all the special families, there should also be a test that reference counting works correctly on an object returned from alloc which was never initialized.

Technically, Rubicon treats `alloc` no different than `new`, `copy`, etc. It has only special handling for `init`.

But agreed, it is good to test `alloc` in isolation because it is special on the Objective-C side in not returning an initialized object.

Added a test case now.

src/rubicon/objc/api.py
samschott: FYI, I am actually seeing those errors printed to stderr in Python 3.13 on exit if there is no special handling.

src/rubicon/objc/api.py
```diff
@@ -100,13 +100,14 @@ def get_method_family(method_name: str) -> str:
...
-    for c in method_name:
-        if c.isupper():
-            break
-        leading_lowercases.append(c)
-    return "".join(leading_lowercases)
+    first_component = method_name.lstrip("_").split(":")[0]
+    for family in _RETURNS_RETAINED_FAMILIES:
+        if first_component.startswith(family):
+            remainder = first_component.removeprefix(family)
+            if remainder == "" or remainder[0].isupper():
```
mhsmith, line 107: "A selector is in a certain selector family if, ignoring any leading underscores, the first component of the selector either consists entirely of the name of the method family or it begins with that name followed by a character other than a lowercase letter."

```suggestion
            if remainder == "" or not remainder[0].islower():
```

tests/test_core.py
```diff
@@ -2186,7 +2209,9 @@ def work():
...
     def test_get_method_family(self):
-        self.assertEqual(get_method_family("perform"), "perform")
-        self.assertEqual(get_method_family("performWith:"), "perform")
-        self.assertEqual(get_method_family("_performWith:"), "perform")
-        self.assertEqual(get_method_family("_perform:with:"), "perform")
+        self.assertEqual(get_method_family("mutableCopy"), "mutableCopy")
+        self.assertEqual(get_method_family("mutableCopy:"), "mutableCopy")
+        self.assertEqual(get_method_family("_mutableCopy:"), "mutableCopy")
+        self.assertEqual(get_method_family("_mutableCopy:with:"), "mutableCopy")
+        self.assertEqual(get_method_family("_mutableCopyWith:"), "mutableCopy")
```
mhsmith, line 2216:
```suggestion
        self.assertEqual(get_method_family("_mutableCopyWith:"), "mutableCopy")
        self.assertEqual(get_method_family("_mutableCopy_with:"), "mutableCopy")
```

src/rubicon/objc/api.py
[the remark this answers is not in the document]
mhsmith: For completeness I think there should also be a functional test of `mutableCopy`.

src/rubicon/objc/api.py
```diff
@@ -91,9 +91,47 @@
...
+            break
+        leading_lowercases.append(c)
+    return "".join(leading_lowercases)
+
+
+def method_name_to_tuple(name: str) -> (str, tuple[str, ...]):
+    """
+    Performs the following transformation:
+
+    "methodWithArg0:withArg1:withArg2" -> "methodWithArg0", ("", "withArg1", "withArg2")
```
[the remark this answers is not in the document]
mhsmith: Do I understand correctly that if a selector contains any colons, then it must end with a colon?

src/rubicon/objc/api.py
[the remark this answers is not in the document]
mhsmith: And `new`. Maybe the common code between these tests could be factored out.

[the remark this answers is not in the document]
samschott: I guess that would also mean including a test for `copy` when a new object is returned? Sure, I can add all of those. Factoring out common code would indeed be nice, but may be tricky because we are testing garbage collection and deallocation.

Something to look at tomorrow.

src/rubicon/objc/api.py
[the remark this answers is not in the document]
samschott: I believe so, yes, because only the part before the first colon is the message name, and everything that follows must be a named parameter `argumentName:` or an unnamed parameter `:`. The first argument is always unnamed. See https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjectiveC/Chapters/ocObjectsClasses.html#//apple_ref/doc/uid/TP30001163-CH11-SW6.

src/rubicon/objc/api.py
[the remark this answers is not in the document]
samschott: Done.

tests/test_core.py
```diff
@@ -1872,128 +1894,114 @@ class TestO:
         self.assertIsNone(wr_python_object())
 
     def test_objcinstance_returned_lifecycle(self):
-        """An object is retained when creating an ObjCInstance for it and autoreleased
-        when the ObjCInstance is garbage collected.
+        """An object is retained when creating an ObjCInstance for it without implicit
+        ownership It is autoreleased when the ObjCInstance is garbage collected.
```
mhsmith, line 1898:
```suggestion
        ownership. It is autoreleased when the ObjCInstance is garbage collected.
```

mhsmith (approved):

mhsmith:
Thanks again for doing this!

HalfWhitt:
Never would I have dreamed that one graph resizing incorrectly would lead to all this work! Hats off for sure!

freakboy3742:
Echoing what @mhsmith said - a huge thanks for this. This removes a huge wart on Rubicon's design; and as the Toga update PR highlights, resolves a bunch of weird edge cases and memory issues. 

Following up on [this comment](https://github.com/beeware/rubicon-objc/pull/543#discussion_r1868358885) - should I cut a 0.5.0 release now, or do you want me to hold off until you've had a chance to take a swing at that update as well? There's no particular pressure to get a release out the door, but given this is a big change, it feels like we should make it official sooner rather than later so that the impact isn't forgotten.

PR #558, opened 2025-01-27
Add a compatibility layer for the asyncio changes in Python 3.14.0a4.

### LINKED ISSUE #557: Test suite failing under Python 3.14.0a4
### Describe the bug

Python 3.14.0a4 has just been released; it makes some changes to asyncio that cause the Rubicon test suite to fail:
```
_____________________ ERROR collecting tests/test_async.py _____________________
tests/test_async.py:7: in <module>
.tox/py/lib/python3.14/site-packages/rubicon/objc/eventloop.py:659: in <module>
    class EventLoopPolicy(events.AbstractEventLoopPolicy):
E   AttributeError: module 'asyncio.events' has no attribute 'AbstractEventLoopPolicy'. Did you mean: 'AbstractEventLoop'?
=============================== warnings summary ===============================
.tox/py/lib/python3.14/site-packages/rubicon/objc/eventloop.py:6
.tox/py/lib/python3.14/site-packages/rubicon/objc/eventloop.py:6
  /Users/runner/work/rubicon-objc/rubicon-objc/.tox/py/lib/python3.14/site-packages/rubicon/objc/eventloop.py:6: DeprecationWarning: 'asyncio.DefaultEventLoopPolicy' is deprecated and slated for removal in Python 3.16
    from asyncio import (
```

### Steps to reproduce

Run Rubicon's CI on Python 3.14.0a4

### Expected behavior

Test suite should pass

### Screenshots

_No response_

### Environment

- Operating System: all
- Python version: 3.14.0a4+

### Logs

```

```

### Additional context

_No response_

freakboy3742:
Python 3.14.0a4 [deprecated the use of custom EventLoopPolicy instances](https://github.com/python/cpython/issues/127949), in favor of instantiating custom event loops directly. This PR adapts to that change by:

1. Aliasing the imports of the EventLoopPolicy base classes to their new "internal" names. This solves the immediate testing issue that was causing failures in CI.
2. Adapts to a name change in `iscoroutinefunction` that raises now warnings in Python 3.14.
3. Adding `RubiconEventLoop()` as a public API. On Python 3.14+, this is the CFEventLoop class; on earlier versions, it is a wrapper method that creates and installs an event loop policy, then creates a new event loop.

This approach should be backwards compatible. Any code using the old policy-based approach will continue to work, but will raise a DeprecationWarning advising users to switch to the `RubiconEventLoop()` approach. Code using the `RubiconEventLoop()` approach will work on all supported versions of Python, and *won't* raise a warning (even though technically it is using the deprecated EventLoopPolicy class.

Fixes #557.

## PR Checklist:


mhsmith (approved):

src/rubicon/objc/eventloop.py
```diff
@@ -733,6 +767,25 @@ def set_child_watcher(self, watcher):
...
+        # If they're using RubiconEventLoop(), they've done the necessary adaptation.
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore",
+                message=r"^Custom EventLoopPolicy instances have been deprecated by Python 3.14",
+                category=DeprecationWarning,
+            )
+            policy = EventLoopPolicy()
+        set_event_loop_policy(policy)
+        return new_event_loop()
```
samschott, line 783: I expect rubicon-objc to be typically used in cases where it's ok to set the event loop policy globally. However, it might be a bit less invasive and more compatible with the Python 3.14 approach if return the event loop instance without any side effects. For example:

```Python
policy = EventLoopPolicy()
return policy.get_event_loop()
```

WDYT?

freakboy3742: So... [`get_event_loop()` has been deprecated as well](https://github.com/python/cpython/issues/83710) (and has been for a couple of releases). 

The [Python PR that triggered this change](https://github.com/python/cpython/issues/127949) documents the "suggested" API usage as:
```
loop = MyCustomEventLoop()
loop.run_until_complete(task)
```

src/rubicon/objc/eventloop.py
```diff
@@ -17,14 +18,34 @@
...
+    # replacement API; the feature can be removed.
+    #
+    # Python 3.14 also started the deprecation of event loop policies, to be
+    # finalized in Python 3.16; there was some symbol renaming to assist in
+    # making the deprecation visible. See
+    # https://github.com/python/cpython/issues/127949 for details.
+    from asyncio import (
+        _AbstractEventLoopPolicy as AbstractEventLoopPolicy,
+        _DefaultEventLoopPolicy as DefaultEventLoopPolicy,
+    )
```
samschott, line 39: I assume those imports will stop working with Python 3.16. Is there anything we can do to prepare for this now in addition to the deprecation warning below?

freakboy3742: Sure - since we know the deprecation will be finalised with 3.16, we can make these imports and the actual EventLoopPolicy definition conditional on `version_info < (3, 16)`. That way, when, in 6 years time, we run the `pyupgrade` with the 3.16 minimum version, all the dead code will be automatically removed.

samschott: Thanks! No strong opinions here, I'll leave it to you whether you think thats worth doing :)

src/rubicon/objc/eventloop.py
freakboy3742: > How about the following then?
> 
> ```
> policy = EventLoopPolicy()
> return policy.new_event_loop()
> ```

That's functionally identical to `asyncio.new_event_loop()` - but I guess it doesn't hurt to be explicit about which policy is creating the loop and avoiding a function call.
 
> This might be stupid question, but why special case Python < 3.14 at all? Could we use the "suggested" API usage and instantiate the `CFEventLoop` directly on older Python versions as well?

In < 3.14, there are still *some* functions on the policy - child watchers, mechanisms to get the default loop, and so on. For maximum compatibility, I figured it was best to preserve the old API until it is fully deprecated.

samschott (approved):

PR #596, opened 2025-05-27
Fix a Version Typo in CTypes Patch

johnzhou721:
The >=3.15 is for the *not* part of being compatible, and when you invert it, the "inclusive" = becomes exclusive, so it's only up to 3.14.  Python 3.15 development has not started when blaming line 27.

## PR Checklist:


freakboy3742 (approved):
Good catch; thanks.

PR #598, opened 2025-06-07
Remove an Extra Warning Filter in eventloop.py

johnzhou721:
The version is smaller than 3.14 in the branch of the customly-defined RubiconEventLoop function, so there's no "deprecations in 3.14" to speak of when creating an event loop policy.

## PR Checklist:


changes/598.misc.rst
```diff
@@ -0,0 +1 @@
+Remove an extra warning filter in eventloop.py.
```
johnzhou721, line 1:
```suggestion
Remove an extra warning filter in eventloop.py.

```

freakboy3742 (commented):
You've misunderstood what is going on here. 

Any attempt to instantiate EventLoopPolicy raises a warning in Python 3.13 or earlier, as a pre-warning to users that they need to adapt their code. The implementation of RubiconEventLoop on Python 3.13 and earlier uses the *old* API (which *hasn't* been deprecated on Python3.13 and earlier) to reproduce the behavior that will exist o Python 3.14+. If users have invoked `RubiconEventLoop()`, they've already done the adaptation that will be required to avoid the deprecation in future, so the warning can be silenced.

johnzhou721:
Sorry. "[H]ave been" meant that it's already done, but forgot that CPython has backports.

PR #658, opened 2025-10-15
Move to MkDocs from Sphinx.

### LINKED ISSUE #188: Consider using sphinx-autoapi to generate docs without having to import rubicon.objc
Many parts of our reference documentation are generated using the Sphinx [autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) extension, which includes contents of docstrings from our Python code into the documentation. To extract the docstrings, autodoc imports the modules in question and reads the `__doc__` attributes at runtime.

Unfortunately, this causes some problems with Rubicon, because even just importing `rubicon.objc` requires a working Objective-C runtime (to look up Objective-C runtime functions and core classes like `NSObject`), which isn't available in documentation building environments like Read the Docs. We were able to work around this issue by writing [a mock version of `ctypes`](https://github.com/beeware/rubicon-objc/blob/5cfd466ac439c729cbff055a8b1d3404df322d7c/docs/_mocked_modules/ctypes/__init__.py) and using it during the documentation build in place of the real `ctypes` module. This mock `ctypes` implements just enough APIs to make `rubicon.objc` import successfully and allow Sphinx to extract the docstrings. (See #150 and #154 for details.)

In the PR discussion I said:

> It would be really nice if Sphinx autodoc could extract docstrings by statically parsing the code instead of importing/executing it, then we wouldn't have this issue...

It turns out that something like this exists: [sphinx-autoapi](https://github.com/readthedocs/sphinx-autoapi). I haven't looked into it in detail or tried it yet, but it looks promising. It would be great if we could change our doc build to use this extension instead of regular Sphinx autodoc - then we could get rid of our custom mock `ctypes` version and wouldn't have to worry about runtime import problems anymore.

kattni:
This is ready for review. Note that I used a combination of the cleanup scripts and manual search-and-replace to handle the conversion from rST to Markdown. The amount of content in Rubicon made it likely that tailoring the scripts to Rubicon would take more time than manual updates to handle the rest, so I went with manual.

TODOs:
- [ ] Update the `griffe` dependency in `pyproject.toml` to the eventual commit hash on merge of [this PR](https://github.com/mkdocstrings/griffe/pull/422).
- [x] @freakboy3742 Please add content to `about/roadmap.md`.
- [x] @freakboy3742 Determine what parts of the `__new__` docstring should be in the class docstring, and what should remain a comment under `__new__`. Currently all content is duplicated in both `__new__` and the associated class docstring. Please search for `# TODO: Duplicated in class docstring.` to find the applicable `__new__` comments.

Closes #188.

## PR Checklist:
 
### DIFF
--- a/.github/workflows/dependabot-changenote.yml
+++ b/.github/workflows/dependabot-changenote.yml
@@ -10,3 +10,5 @@ jobs:
     name: Dependabot Change Note
     uses: beeware/.github/.github/workflows/dependabot-changenote.yml@main
     secrets: inherit
+    with:
+      changenote-format: "md"
--- a/.github/workflows/pre-commit-update.yml
+++ b/.github/workflows/pre-commit-update.yml
@@ -10,3 +10,6 @@ jobs:
     name: Update pre-commit
     uses: beeware/.github/.github/workflows/pre-commit-update.yml@main
     secrets: inherit
+    with:
+      pre-commit-source: "--group pre-commit"
+      changenote-format: "md"


src/rubicon/objc/api.py
```diff
@@ -1016,40 +1092,40 @@ def __getattr__(self, name):
...
         returned using its getter method. An attribute is considered a property
         if any of the following are true:
 
         * A property with the name is present on the class (i.e. declared using
-          ``@property`` in the source code)
+          `@property` in the source code)
         * There is both a getter and setter method for the name
         * The name has been declared as a property using
-          :meth:`ObjCClass.declare_property`
+          [`ObjCClass.declare_property`][rubicon.objc.api.ObjCClass.declare_property]
```
freakboy3742: Since this is a `:meth:` reference, it should probably have the `()`.
```suggestion
          [`ObjCClass.declare_property()`][rubicon.objc.api.ObjCClass.declare_property]
```

src/rubicon/objc/api.py
```diff
@@ -1717,40 +1862,252 @@ def __new__(cls, name_or_ptr):
...
-    * :class:`NSDecimalNumber`: Converted to :class:`decimal.Decimal`
-    * :class:`NSDictionary`: Converted to :class:`dict`, with all keys and
-         values converted recursively
-    * :class:`NSArray`: Converted to :class:`list`, with all elements converted
-         recursively
-    * :class:`NSNumber`: Converted to a :class:`bool`, :class:`int` or
-         :class:`float` based on the type of its contents
-
-    Other objects are returned unmodified as an :class:`ObjCInstance`.
+    * [`objc_id`][rubicon.objc.runtime.objc_id]: Wrapped in an
```
freakboy3742: Some odd indentation and bolding here (and in `ns_for_py` - I think it's the newline being interpreted as a new bullet level?

freakboy3742 (changes requested):
I've found a handful of really cosmetic issues, mostly relating to linking and the top level landing pages:

* Anchor links on the homepage and landing pages (using `[How to][how-to]` rather than `[How to](/howto/index.md)
* Sidebar entry for the FAQ says "Why Rubicon"
* Reference landing page doesn't have any links

But otherwise, this looks incredible. In many places, it's a substantial improvement on the historical Sphinx docs IMHO.

samschott:
Apologies for having not have a look at this earlier. I love it, this is quite an improvement in the completeness and the appearance of the docs!

I'm curious about the newly added stubs. They seem to have doc strings that duplicate what's in code. That could make the API docs a bit more annoying to maintain going forward. Would it be possible to only keep one or the other?

PR #731, opened 2026-03-21
add tests to increase coverage

JohananOppongAmoateng:
Add more tests to increase coverage. #729 

## PR Checklist:

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import pytest
 
 from rubicon.objc import (
     CFRange,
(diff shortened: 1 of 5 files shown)


tests/test_types.py
```diff
@@ -120,3 +145,135 @@ def test_function_NSEdgeInsetsMake():
...
+        ValueError, match="has 3 fields, but a sequence of length 2 was given"
+    ):
+        compound_value_for_sequence([1, 2], MyArray)
+
+    with pytest.raises(TypeError, match="Don't know how to convert a sequence"):
+        compound_value_for_sequence([1], c_int)
+
+
+def test_maker_functions():
+    """Test maker functions (lines 888, 893, 903, 908, 913, 938)"""
```
freakboy3742: Two notes here:

1. You don't refer to specific lines of code in coverage, because a one line unrelated change in the same file will make all the line references invalid. Refer to a specific feature that is being tested.
2. In this test, you're testing 7 different maker methods. These should be 7 independent tests (or, at least, one test that has been parameterised with 7 items). 
3. We're not just looking to verify line coverage - we want to verify that the thing being tested works as advertised. An `isinstance` check is *one* thing to verify - but verifying that the attributes of the created object have been set correctly is the other key detail.

PR #746, opened 2026-04-13
Ensure CocoaLifecycle doesn't terminate the app.

### LINKED ISSUE #745: CocoaLifecycle terminates application instances
### Describe the bug

When the CocoaLifcycle exits, it fully terminates the application instance, rather than just stopping the event loop. This means it is not possible to have business logic *after* the event loop shutdown, and Python`atexit` handlers will not be executed.

### How to reproduce the bug

See beeware/toga#4230

### Minimum example code

```shell
import toga
import atexit
import sys

def my_atexit():
    print(f'my atexit is never called')

original_excepthook = sys.excepthook
def my_exception_handler(type, value, traceback):
    print(f'my uncaught exception handler is never called')
    original_excepthook(type, value, traceback)

class App(toga.App):
    def startup(self):
        self.main_window = toga.MainWindow()
        self.button = toga.Button('exit', on_press=self._button_on_press)
        self.main_window.content = self.button
        self.main_window.show()
    def _button_on_press(self, button):
        # raise RuntimeError()
        self.app.exit()

sys.excepthook = my_exception_handler
# raise RuntimeError()
atexit.register(my_atexit)
app = App('App', 'org.beeware.toga.app')
app.main_loop()
print(f'This is never printed.')
```

### Screenshots

_No response_

### Environment details

- Operating system and version: All; tested on macOS 15.6.1
- Python version: All; tested on 3.12.7
- Software versions:
  - Rubicon 0.5.3

### Logs

```

```

### Additional context

It's interesting to note that CocoaLifecycle is inconsistent with iOSLifecycle and CFLifecycle - those two lifecycles both stop, rather than terminating the app.

freakboy3742:
Fixes #745 

Modifies the `stop()` method on CocoaLifecycle to just stop the event loop, rather than fully terminating the app.

Includes tests for the CFLifecycle and CocoaLifecycle, plus a cleanup of the constructor for CFLifecycle so that an explicit event loop isn't required.

I've tested this with the test case provided in the discussion on beeware/toga#4320, and with the `Window` example in the Toga repo (which has complex on-exit logic); both seem to work with this change.

## PR Checklist:

### DIFF
--- a/src/rubicon/objc/eventloop.py
+++ b/src/rubicon/objc/eventloop.py
@@ -818,8 +818,11 @@ else:  # pragma: no-cover-if-lt-py314
 class CFLifecycle:
     """A lifecycle manager for raw CoreFoundation apps."""
 
-    def __init__(self, cfrunloop):
-        self._cfrunloop = cfrunloop
+    def __init__(self, cfrunloop=None):
+        if cfrunloop:
+            self._cfrunloop = cfrunloop
+        else:
+            self._cfrunloop = libcf.CFRunLoopGetMain()
 
     def start(self):
         libcf.CFRunLoopRun()
@@ -838,10 +841,10 @@ class CocoaLifecycle:
         self._application.run()
 
     def stop(self):
-        self._application.terminate(None)
+        self._application.stop(None)
 
 
-class iOSLifecycle:
+class iOSLifecycle:  # pragma: no cover
     """A life cycle manager for iOS (`UIApplication`) apps."""
 
     def start(self):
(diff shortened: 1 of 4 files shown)


src/rubicon/objc/eventloop.py
```diff
@@ -530,10 +530,7 @@ def run_forever(self, lifecycle=None):
                 "To recursively run the event loop, call run()."
             )
 
-        try:
```
freakboy3742, line 533: This redundancy was revealed during testing. The only way to exit `run_forever()` is to call `stop()` on the event loop so having the event loop call `stop()` means you get a double call. It's not a problem - all it means is a redundant call.

mhsmith (approved):

Issue #786, opened 2026-07-02
Crash on iOS due to NSEvent not being available


freakboy3742:
### Describe the bug

The 0.5.5 release included a fix for #784; however, that fix doesn't work on iOS, because NSEvent isn't available

### How to reproduce the bug

Run any app on iOS with Rubicon 0.5.5.

### Minimum example code

```shell

```

### Environment details

- Operating system and version: All (tested on iOS 26.2)
- Python version: All (tested on 3.14)
- Software versions:
  - Rubicon 0.5.5

### Logs

```
  File "/Users/rkm/Library/Developer/CoreSimulator/Devices/14013343-F25E-400D-B647-AB8A0C589067/data/Containers/Bundle/Application/B5AC46ED-65F2-4EE7-B083-F9EC20CF0988/Python WA Demo.app/app_packages/toga_iOS/app.py", line 2, in <module>
    from rubicon.objc.eventloop import RubiconEventLoop, iOSLifecycle
  File "/Users/rkm/Library/Developer/CoreSimulator/Devices/14013343-F25E-400D-B647-AB8A0C589067/data/Containers/Bundle/Application/B5AC46ED-65F2-4EE7-B083-F9EC20CF0988/Python WA Demo.app/app_packages/rubicon/objc/eventloop.py", line 117, in <module>
    NSEvent = ObjCClass("NSEvent")
  File "/Users/rkm/Library/Developer/CoreSimulator/Devices/14013343-F25E-400D-B647-AB8A0C589067/data/Containers/Bundle/Application/B5AC46ED-65F2-4EE7-B083-F9EC20CF0988/Python WA Demo.app/app_packages/rubicon/objc/api.py", line 1418, in __new__
    ptr, name = cls._new_from_name(name_or_ptr)
                ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/Users/rkm/Library/Developer/CoreSimulator/Devices/14013343-F25E-400D-B647-AB8A0C589067/data/Containers/Bundle/Application/B5AC46ED-65F2-4EE7-B083-F9EC20CF0988/Python WA Demo.app/app_packages/rubicon/objc/api.py", line 1275, in _new_from_name
    raise NameError(f"ObjC Class {name} couldn't be found.")
NameError: ObjC Class b'NSEvent' couldn't be found.
```

freakboy3742:
Fixed by #787

PR #806, opened 2026-08-05
Load AppKit before accessing NSEvent in macOS event loop

aronnaxlin:
## Problem

When `rubicon.objc.eventloop` is imported before any framework that has already loaded AppKit (e.g. in headless daemons or CLI apps that don't use toga), `ObjCClass("NSEvent")` raises `NameError: ObjC Class b'NSEvent' couldn't be found` because AppKit isn't registered in the Objective-C runtime yet.

This is the macOS equivalent of #786 (which is iOS-specific). It affects apps that import `rubicon.objc.eventloop` directly for a CoreFoundation-based asyncio event loop without first loading AppKit through another dependency.

### Reproducer

```python
# Fresh interpreter, no other rubicon/AppKit imports first
```

On macOS (any version, tested on Python 3.14 + rubicon-objc 0.5.6) this crashes with:

```
NameError: ObjC Class b'NSEvent' couldn't be found.
```

## Fix

Call `load_library("AppKit")` before accessing `NSEvent`, mirroring how `CoreFoundation` is already loaded for the CF types at the top of the module. `load_library` is already imported in this module.

`NSRunLoop` is available from Foundation (which `rubicon.objc` loads transitively), so only `NSEvent` needs AppKit.

## PR Checklist:
  - Assisted-by: Oh My Pi (OMP) with Kimi K3
- [ ] This PR will generate a changelog entry

### DIFF
--- a/changes/806.bugfix.md
+++ b/changes/806.bugfix.md
@@ -1,0 +1,1 @@
+The macOS event loop now loads AppKit before accessing NSEvent, fixing a NameError when the event loop is imported before any other framework that loads AppKit (e.g. in non-toga headless apps).
--- a/changes/806.bugfix.md
+++ b/changes/806.bugfix.md
@@ -1,1 +1,1 @@
-The macOS event loop now loads AppKit before accessing NSEvent, fixing a NameError when the event loop is imported before any other framework that loads AppKit (e.g. in non-toga headless apps).
+The macOS event loop integration no longer requires an explicit load of the `AppKit` library before being imported.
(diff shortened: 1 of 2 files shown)


changes/806.bugfix.md
```diff
@@ -0,0 +1 @@
+The macOS event loop now loads AppKit before accessing NSEvent, fixing a NameError when the event loop is imported before any other framework that loads AppKit (e.g. in non-toga headless apps).
```
freakboy3742:
```suggestion
The macOS event loop integration no longer requires an explicit load of the `AppKit` library before being imported.
```

freakboy3742 (approved):
Thanks for the PR! I've tweaked the release note, but otherwise this looks great. Apologies for the inconvenience around the actual problem you hit.

PR #816, opened 2026-08-17
test: Add tests for NSString and NSDictionary error paths

mgstabrani:
Adds tests for the error and fallback paths in `ObjCStrInstance` and `ObjCMutableDictInstance`, which were previously uncovered. This PR adds tests only

## PR Checklist:

Assisted-by: Claude Opus 5 (Claude Code)

### DIFF
--- a/tests/test_NSString.py
+++ b/tests/test_NSString.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import operator
 import os
 
 import pytest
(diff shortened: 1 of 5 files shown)


mgstabrani:
#729

tests/test_NSString.py
```diff
@@ -119,6 +121,36 @@ def test_nsstring_compare(py_left, py_right):
     assert (ns_left > ns_right) == (py_left > py_right)
 
 
+@pytest.mark.parametrize("py_left", TEST_STRINGS)
+@pytest.mark.parametrize("py_right", TEST_STRINGS)
```
freakboy3742, line 125: This is a good parameterisation of the tests, but the TEST_STRINGS don't (obviously) enumerate the basic "A > B" and "B < A" case.

freakboy3742 (approved):
Looking at the coverage results, this was 2 lines away from full coverage on the `collections` module, which got me wondering why. It turns out there was a very subtle bug - `__mul__`  should have been catching TypeError, not AttributeError (to catch the case where a string has `__rmul__`, but not `__index__`. I've added a test for that, and collections.py is at 100% coverage!

Thanks for the PR!

PR #820, opened 2026-08-25
test: Add tests for type encoding conversion and registration

mgstabrani:
Add tests for type encoding registration, conversion, and parsing helpers

## PR Checklist:

Assisted-by: Claude Opus 5 (Claude Code)


mgstabrani:
https://github.com/beeware/rubicon-objc/issues/729

freakboy3742:
Thanks for the PR; apologies for the delay in reviewing. I hope take a look at this in the next few days.

mgstabrani:
> Thanks for the PR; apologies for the delay in reviewing. I hope take a look at this in the next few days.

@freakboy3742 Sure, no problem. Take your time whenever you get a chance

tests/test_types.py
```diff
@@ -109,3 +190,378 @@ def test_function_NSEdgeInsetsMake():
...
+    Structs are compared by their raw bytes, because ctypes structs of equal value don't
+    compare equal to each other.
+    """
+    made = make(*args)
+
+    assert type(made) is type(expected)
+    assert bytes(made) == bytes(expected)
+
+
+def test_arm_detected_without_processor(monkeypatch):
```
freakboy3742: This is very much legacy handling for very old iOS support packages. I think the fix here is to remove the legacy handling, rather than adding some fragile monkey patch-based tests.

tests/test_types.py
```diff
@@ -1,19 +1,100 @@
...
+
+    yield
+
+    for registry, contents in zip(registries, saved, strict=True):
+        registry.clear()
+        registry.update(contents)
+
+
+@pytest.fixture
+def unregistered_type():
```
freakboy3742: Class definitions don't need to go into a fixture; they can be top-level objects, because re-creation won't change anything.

freakboy3742 (approved):
Thanks for the PR! I've pushed some cleanups and simplifications, most of which are either the result of things I've flagged inline, or `no cover` blocks that we can add because we can't easily test 32-bit configurations in CI.

With those changes, there's only 2 blocks of code that are untested in `types.py`..

mgstabrani:
> Thanks for the PR! I've pushed some cleanups and simplifications, most of which are either the result of things I've flagged inline, or `no cover` blocks that we can add because we can't easily test 32-bit configurations in CI.
> 
> With those changes, there's only 2 blocks of code that are untested in `types.py`..

Thanks for the review and for pushing the fixes directly.

PR #823, opened 2026-09-03
Add a CI configuration for iOS testing

freakboy3742:
Adds a CI configuration for iOS.

This is mostly adding a Briefcase and CI configuration; but it also involves re-organizing the ObjC rubiconharness so it can support building different versions for macOS, iOS and iOS simulator. It also involves modifying `load_library` to use the same framework lookup process that CPython uses.

Only tests on ARM64, as adding x86-64 support would be a bunch of headaches for a simulator platform that will be EOL in a year.

There's one test with a TODO; I'm treating that as part of #729.

## PR Checklist:

- [ ] This PR was generated or assisted using an AI tool


mhsmith:
> Doesn't test 3.15 yet, as 3.15 iOS support isn't ready in Briefcase ready yet.

It looks like you've added a 3.15 test, which fails. Do you want to remove it and merge this PR, or delay this PR until Briefcase is ready?

freakboy3742:
> > Doesn't test 3.15 yet, as 3.15 iOS support isn't ready in Briefcase ready yet.
> 
> It looks like you've added a 3.15 test, which fails. Do you want to remove it and merge this PR, or delay this PR until Briefcase is ready?

Adding the 3.15 test was deliberate - I thought I was going to get the updated iOS template out yesterday. I've published that template now, so the test should pass (just waiting on CI to complete)

mhsmith (approved):

PR #832, opened 2026-09-17
test: Add coverage for ctypes_patch.py setfunc-only path

mgstabrani:
## Summary

Add test coverage for the `setfunc`-only error path in `ctypes_patch.py`, and mark two unreachable code branches with `# pragma: no cover` to achieve 100% coverage for this module.

## PR Checklist:
Assisted-by: Claude Haiku 4.5


tests/test_ctypes_patch.py
```diff
@@ -186,3 +186,31 @@ def test_get_stginfo_of_type_uninitialized(tp):
...
     with pytest.raises(TypeError, match="has not been initialized"):
         ctypes_patch.get_stginfo_of_type(tp)
+
+
+@before_313
+def test_no_patch_setfunc_only_stgdict():
+    """A type that already has a setfunc, but no getfunc, cannot be patched."""
+
+    class OnlySetfunc(ctypes.Structure):
+        _fields_ = [("x", ctypes.c_int)]
```
freakboy3742: If this structure is re-used, it should be defined in a common block at the top of the file.

tests/test_ctypes_patch.py
```diff
@@ -186,3 +186,31 @@ def test_get_stginfo_of_type_uninitialized(tp):
...
+
+    stg = ctypes_patch.get_stgdict_of_type(OnlySetfunc)
```
freakboy3742: If this line is the only difference between the two tests, then a single test with an `if sys.version_info < (3, 13)` is preferable to two tests.

freakboy3742 (changes requested):
Two suggestions inline.

The one other issue - adding a `no cover` should always be a last resort, and should generally be accompanied with a code comment justifying the `no cover`. 

What is your justification there? I'm not saying you're *wrong* in claiming it's unreachable - I just want to make sure your justification aligns with my understanding of the situation.

tests/test_ctypes_patch.py
mgstabrani: updated

tests/test_ctypes_patch.py
mgstabrani: I've combined the tests into one. So now it is only used once in that function. Keeping it inside the function.

mgstabrani:
> Two suggestions inline.
> 
> The one other issue - adding a `no cover` should always be a last resort, and should generally be accompanied with a code comment justifying the `no cover`.
> 
> What is your justification there? I'm not saying you're _wrong_ in claiming it's unreachable - I just want to make sure your justification aligns with my understanding of the situation.

Thanks for the feedback! For the no cover justification: `ctypes` only ever calls `getfunc` with `size=0` or the field's registered size (which is fixed since `_fields_` is immutable). Because there's no Python API that can trigger a size mismatch, this defensive check is unreachable in tests. I've gone ahead and added this reasoning as a comment in the code to accompany the pragma.

freakboy3742 (approved):
Ok - thanks for that clarification. I've tweaked the comment a little, but otherwise this is great!

mgstabrani:
> Ok - thanks for that clarification. I've tweaked the comment a little, but otherwise this is great!

Sounds perfect, thank you!

mgstabrani:
#729

